Issue #391, opened 2014-10-01
tables 3.1.1 won't install with Cython 0.21


flyingfrog81:
With Cython 0.21 the install script hangs in setup.py at line 126:

``` python
from Cython.Compiler.Main import Version
```

This results in:

``` bash
pip install cython tables
```

arresting at tables installation with:

``` bash
.. ERROR:: You need Cython 0.13 or greater to compile PyTables!
```

Which is also the wrong error message. 
Development version fixes this behaviour but the pypi package is still affected.

Issue #388, opened 2014-09-16
setup.py fails with Cython 0.21


cimatosa:
due to some structural changes in the Cython module the import called in setup.py (master)

l. 126: from Cython.Compiler.Main import Version

fails. looks different in develop branch.

This essentially results in a setup failure with Error "You need to Cython >= 0.13" when trying to install pytables via pip.

Issue #386, opened 2014-09-11
Cython dependency at the installation


biernackip:
With newest Cython (0.21, at the moment of writing) PyTables do not want to install. Downgrading to version 0.15 helped. Cython 0.21 is simply not recognized.

PR #392, opened 2014-10-03
adding warning about cython 0.21 to readme

lbergelson:
I think it would be helpful to let people know that PyTables isn't working with Cython 0.21 right now.  I looked at the README first and couldn't find any information there.  All the issues #391, #388, #386, #387 are marked closed so they don't show up at first glance either.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -90,6 +90,7 @@ make use of PyTables.
 
 Installation
 ------------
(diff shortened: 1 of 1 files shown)


andreabedini:
Rather than a warning in the docs I think we should just bump up the cython required version.

FrancescAlted:
Agreed.  Required Cython 0.22 in rev f73f87f.

PR #414, opened 2015-01-11
Blosc supports all architectures

keszybz:
https://github.com/FrancescAlted/blosc/issues/3 has been
resolved in version 1.5. Since version 1.5 is now bundled
with PyTables, remove the check.

I built a Fedora package with a similar patch, using an external blosc 1.5.2, and tests pass on i686, amd64 and arm32.

### DIFF
--- a/tables/utilsextension.pyx
+++ b/tables/utilsextension.pyx
@@ -276,7 +276,7 @@ blosc_version = register_blosc_()
 def _arch_without_blosc():
     import platform
     arch = platform.machine().lower()
-    for a in ("arm", "sparc", "mips"):
+    for a in ("arm", "sparc", "mips", "aarch64"):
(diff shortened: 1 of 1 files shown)


avalentino:
Hi @keszybz, thanks for submitting this PR.
Currently PyTables allows the user to build against external blosc libs and I would like to keep supporting older version of blosc.
Could you please modify your PR so that the old behavior is maintained in case of blosc < 1.5?

keszybz:
Patch was borked. This version should be better.

Fedora build on 64-bit arm fails too: http://arm.koji.fedoraproject.org//work/tasks/5171/2845171/build.log. Add an exception for this architecture too.

avalentino:
Thanks, @keszybz

PR #427, opened 2015-03-21
Travis container build

andreabedini:
Hi,

this patch changes the current `.travis.yml` build script to enable the use of TravisCI new container infrastructure. It is pretty straightforward as apt packages can be installed without using sudo by listing them in the configuration file.
I also did some cleanup. The environments "with_system_site_packages" seems not be in use anymore on travis (I can't find any documentation for them), and they  seem not to make any difference. Also we don't need to install `python3-numpy` because `numpy` is always already installed in the virtualenvs that TravisCI provides.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
     - liblzo2-dev
 
 install:
-  - pip install -r requirements.txt --use-mirrors
-  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install -U unittest2 --use-mirrors; fi
+  - pip install -r requirements.txt
+  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install -U unittest2; fi
 
 script: make check
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ addons:
   apt_packages:
     - libhdf5-serial-dev
     - liblzo2-dev
-    - python3-numpy
 
 install:
   - pip install -r requirements.txt --use-mirrors
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ sudo: false
 
 addons:
   apt_packages:
+    - libbz2-dev
     - libhdf5-serial-dev
     - liblzo2-dev
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ addons:
     - liblzo2-dev
 
 install:
-  - pip install -U -r requirements.txt
-  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install -U unittest2; fi
+  - pip install -r requirements.txt
+  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install unittest2; fi
 
 script: make check
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,15 @@ python:
   - 3.3
   - 3.4
 
-before_install:
-  - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


andreabedini:
AppVeyor build fail is spurious.

Now I notice that by allowing updates in 63bf92e I am triggering a full rebuild of numpy. Can we install a wheel?

andreabedini:
I'd vote for leaving only py27 and py34, I promise to write down instructions to help out people stuck with obsolete software distributions!

.travis.yml
```diff
@@ -3,16 +3,19 @@ language: python
...
   - 2.7
-  - "2.7_with_system_site_packages"
   - 3.2
-  - "3.2_with_system_site_packages"
   - 3.3
   - 3.4
 
-before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq libhdf5-serial-dev liblzo2-dev libbz2-dev python3-numpy
```
avalentino: I noticed hat you eft libbz2-dev out of the apt_package list.
Is it wnated?

avalentino:
Thanks for taking care of this

> I'd vote for leaving only py27 and py34, I promise to write down instructions to help out people stuck with obsolete software distributions!

I'm +0 for this.
Is there some specific problem with building all version in travis?
I usually only test p34 and p27 on my box so having build for p26, p32 and p33 in travis is useful IMO.
On the other hand I think that p26, p32 and p33 are not so widely used currently.

I leave the decision to you @andreabedini

.travis.yml
andreabedini: not intended at all, thanks for checking

andreabedini:
well, no, there's no specific problem with building all versions. I just find it excessive as I am keen to support only the last versions of py2 and py3. That said, I didn't remove the other versions from travis, so they are still built with this PR.

FrancescAlted:
I am -0 for merging this right now as I agree with @avalentino in that this provides useful information.  If in the future we decide to drop support for 2.6 or 3.3, then we can merge this in.

avalentino:
We could make a partial merge just dropping changes in the test matrix.

FrancescAlted:
@avalentino Are you happy with this: fe50e52 ?  (don't know very well what the ' '_with_system_site_packages' suffix :P

avalentino:
@FrancescAlted  yes, it should be fine .... ant travis seems to agree :)

> (don't know very well what the ' '_with_system_site_packages' suffix :P 

AFAIK with '_with_system_site_packages' travis uses ubuntu packages for numpy, cython and so on, instead of installing them from PyPi

FrancescAlted:
Ah great, then let's use the ubuntu packages as well (sorry Travis for wasting so many cycles).  Closing.

andreabedini:
you can compare the logs of the test running on the 2.7 and the 2.7_with_system_site_packages 
environments:

https://s3.amazonaws.com/archive.travis-ci.org/jobs/58638064/log.txt
https://s3.amazonaws.com/archive.travis-ci.org/jobs/58638065/log.txt

numpy (version 1.9.1) is preinstalled in both environments (not the ubuntu package but 'pip' preinstalled).

with_system_site_packages comes with an older version of python: 2.7.3 rather than 2.7.9

in both environments cython (0.22) and numexpr (2.4.1) are installed on demand.

given also that the documentation about with_system_site_packages enviroments on travis website has been removed (there's no mention of them now), my conclusion is that they are obsolete. Also useless to us because the builds are pretty much identical (as in: same versions of every dependency).

avalentino:
@andreabedini, sorry I totally misunderstood the current meaning of the '_with_system_site_packages'.

FrancescAlted:
Ok.  While testing for all possible scenarios seems a nice thing, I think that using 'native' python versions for ubuntu does not make a lot of sense.  So I have removed them in rev b38e10c.

PR #432, opened 2015-04-14
Remove duplicate `hdf5_version` in Top-level docs

153957:
The global variable `hdf5_version` appeared twice in the docs.
This removes one of the copies.


153957:
https://pytables.github.io/usersguide/libref/top_level.html#tables.hdf5_version

FrancescAlted:
Thanks!

Issue #343, opened 2014-03-18
Unclear purpose of examples/nested-iter.py


153957:
I was looking for a way to combine multiple queries upon a dataset, i.e. perform a query on the result of another query, similar to the .filter() on a QuerySet in Django. In my search I came accross the example '[nested-iter.py](https://github.com/PyTables/PyTables/blob/master/examples/nested-iter.py)'.

It states: `A small example showing the use of nested iterators within PyTables.`

It does not do what I was looking for, but what its real purpose is is also unclear.
It performs 3 different queries and loops over their results. However, two of those queries/loops are inside other for-loops, causing those queries to be performed for each loop of the previous for loop, i.e. for _each_ row where the pressure is less than 9 perform the query to get every row where the energy is less than 10 and print something..

The problem (in my view) is that the p, q and n are in no way related the result from different queries.. the result is NOT only rows where (pressure < 16) & (pressure < 9) & (energy < 10).

Is this really the intent of the example, because it does not seem very useful?

PR #435, opened 2015-04-16
Remove leftover references to 'nested iterators'

153957:
As reported in #343: Nested iterators don't work as originally shown in the example.
The docstring of Table.where() has been correctly modified in 30f89b9fc5.
However, some references to that example were still in other docstrings.


FrancescAlted:
Merged.  Thanks!

Issue #413, opened 2015-01-07
PyTables 3.1.1 installation fails if numpy and numexpr are not preinstalled


dunpealer:
PyTables 3.1.1 (as well as HEAD) attempts to import both [numpy](https://github.com/PyTables/PyTables/blob/106e48f8d6584a9d5e0b1cdc0d080d5161b63039/setup.py#L119) and [numexpr](https://github.com/PyTables/PyTables/blob/106e48f8d6584a9d5e0b1cdc0d080d5161b63039/setup.py#L122), otherwise erroring out.

This behavior is a problem for anyone building deployment requirements in bulk, for example with `pip install -r requirements.txt` inside a virtualenv. The only way to handle it is install all other dependencies first, including numpy and numexpr, then install PyTables as a special extra step.

Instead, PyTables can require its depedencies in a way that would allow them all to be installed together from a single requirements file. For an example of this behavior implemented, see [numexpr's setup.py](https://github.com/pydata/numexpr/blob/master/setup.py).

illume:
Yeah, it's broken with pip.

The way new pip works is different. It used to install them one at a time from the requirements.txt, but now does them in groups. Unfortunately this breaks the dependency checking for tables when using `pip install -r requirements.txt` as numpy and other deps aren't actually installed when it checks.

Removing the custom dependency checker and using the standard one would be better.

FrancescAlted:
I see, and I think this is a bad problem with new pip.  I would be glad to patch this if you can send a pull request.

andreabedini:
Rewriting setup.py has been in the plans for a while. I'm on my iPad and I can't find it now but here should be a issue already open about this.

FrancescAlted:
Hey, I have tried to address concerns raised here in 9f77459.  The general idea is that setuptools should be used if possible (before that had to be enforced by using the 'FORCE_SETUPTOOLS' environment variable), and hence everything should go more smoothly.  Feedback on the commit above is appreciated.

Also, provided that we are distributing the beast with Cythonized module extensions, I am wondering why Cython is necessary for building.  @avalentino any hints?

avalentino:
> Also, provided that we are distributing the beast with Cythonized module extensions, I am wondering why Cython is necessary for building. @avalentino any hints?

Correct, cython is only needed for development or if one wants to build code checked out form the repo.

IMO it is safe to remove cython from (strict) dependencies even if we will loose the possibility to make something like the following:

```
[...]

PR #439, opened 2015-04-21
Correctly bootstrap numpy in setup.py

andreabedini:
Correctly bootstrapping numpy in setup.py is practically impossible but I am trying to follow what is done in pandas. Their approach is mostly based on http://stackoverflow.com/a/21621689/104099.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,10 @@ from distutils.dep_util import newer
 from distutils.util import convert_path
 from distutils.ccompiler import new_compiler
 
+# we need to avoid importing numpy until we can be sure it's installed
+# this approach is based on this SO answer http://stackoverflow.com/a/21621689
+from distutils.command.build_ext import build_ext as _build_ext
+
 cmdclass = {}
 setuptools_kwargs = {}
 
@@ -83,12 +87,15 @@ VERSION = open('VERSION').read().strip()
 with open('requirements.txt') as f:
     requirements = f.read().splitlines()
 
-# Need to get access to numpy, so do it now and crash early if not there
-try:
-    from numpy.distutils.misc_util import get_numpy_include_dirs
-except:
-    exit_with_error("You need NumPy prior to install PyTables!")
+# see http://stackoverflow.com/a/21621689 (this is also what pandas does)
+class build_ext(_build_ext):
+    def build_extensions(self):
+        numpy_incl = pkg_resources.resource_filename('numpy', 'core/include')
 
+        for ext in self.extensions:
+            if hasattr(ext, 'include_dirs') and numpy_incl not in ext.include_dirs:
+                ext.include_dirs.append(numpy_incl)
+        _build_ext.build_extensions(self)
 
 # ----------------------------------------------------------------------
 
@@ -838,5 +845,6 @@ interactively save and retrieve large amounts of data.
     ext_modules=extensions,
     cmdclass=cmdclass,
     data_files=data_files,
+    cmdclass={'build_ext': build_ext},
     **setuptools_kwargs
 )
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ import glob
 
 # Using ``setuptools`` enables lots of goodies
 from setuptools import setup, find_packages
+import pkg_resources
 
 from distutils.core import Extension
 from distutils.dep_util import newer
@@ -38,7 +39,7 @@ class build_ext(_build_ext):
         _build_ext.build_extensions(self)
 
 
-cmdclass={'build_ext': build_ext},
+cmdclass={'build_ext': build_ext}
 setuptools_kwargs = {}
 
 if sys.version_info >= (3,):
--- a/setup.py
+++ b/setup.py
@@ -27,19 +27,35 @@ from distutils.ccompiler import new_compiler
 # We need to avoid importing numpy until we can be sure it's installed
 # This approach is based on this SO answer http://stackoverflow.com/a/21621689
 # This is also what pandas does.
-from distutils.command.build_ext import build_ext as _build_ext
+from setuptools.command.build_ext import build_ext
 
-class build_ext(_build_ext):
-    def build_extensions(self):
+# Fetch the requisites
+with open('requirements.txt') as f:
+    requirements = f.read().splitlines()
+
+numpy_requirement = [ r for r in requirements if 'numpy' in r ]
+
+class BuildExtensions(build_ext):
+    """Subclass setuptools build_ext command
+
+    BuildExtensions does two things
+    1) it makes sure numpy is available
+    2) it injects numpy's core/include directory in the include_dirs parameter of all extensions
+    3) it runs the original build_ext command
+    """
+
+    def run(self):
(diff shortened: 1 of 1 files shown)


andreabedini:
If we put 'numpy' in setup_requirements, pip will try to install it before running the build. As explained on SO there's a small drawback

> Note, that using setup_requires has a subtle downside: numpy will not only be compiled before building extensions, but also when doing python setup.py --help, for example. 

This might be solved by changing setup_requirements depending on the command line paramenters but it might not be worth doing.

andreabedini:
This should definitely solve #413, as you can run `pip install` without pre-installing any dependency.

andreabedini:
The tests on Travis are only partially useful because in `.travis.yml` we do `pip install -r requirements.txt` prior to install pytables. If you want to test this yourself you should create a clean virtual environment and install pytables from source with `pip install .`

andreabedini:
@FrancescAlted still useful for 3.2?

setup.py
```diff
@@ -594,7 +601,7 @@ def get_cython_extfiles(extnames):
 
 setuptools_kwargs['extras_require'] = {}
 
-setuptools_kwargs['setup_requires'] = requirements
+setuptools_kwargs['setup_requires'] = [ r for r in requirements if 'numpy' in r ]
```
andreabedini: This means numpy whatever version. Is there a better way to express this?

andreabedini:
Inspired by http://pytest.org/latest/goodpractises.html#integration-with-setuptools-test-commands I think I solve the problem of `setup.py` unconditionally installing numpy which happens when using `setup_requires` now numpy is only installed when running `build_ext`.

FrancescAlted:
Wow, looks pretty cool (and somewhat complex, but that is the fate of installation procedures in Python) And yes, we are totally on-time for next release.  Merging.

FrancescAlted:
Uh, it does not seem to work for me:

```
$ pip install .
<snip>
  File "/tmp/pip-5Vz4O1-build/setup.py", line 50, in run

    numpy_incl = pkg_resources.resource_filename('numpy', 'core/include')

  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 953, in resource_filename

    return get_provider(package_or_requirement).get_resource_filename(

  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 233, in get_provider

    __import__(moduleOrReq)

ImportError: No module named numpy

----------------------------------------
Cleaning up...
Command /usr/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-5Vz4O1-build/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-8JYr8X-record/install-record.txt --single-version-externally-managed --compile failed with error code 1 in /tmp/pip-5Vz4O1-build
Storing debug log for failure in /home/faltet/.pip/pip.log
```

FrancescAlted:
Ups, false alarm.  After more retries, it seems to work as intended :)

andreabedini:
Thanks @FrancescAlted, do you have any clue on what caused the problem?

FrancescAlted:
No idea, but I can reproduce the correct behavior, so hopefully nothing to worry about.

andreabedini:
Reviewed files:
- setup.py @ r4

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/pytables/pytables/439)

karlacio:
Hi,

I'm pretty new to PyTables and I hope my problem belongs to here.

I'm trying to install (with pip) PyTables on my mac.
I already have numpy installed but at some point during installation setup.py download numpy sources and try to compile it.
From setup.py output:
    Requirement already up-to-date: numpy>=1.7.1 in /Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages (from tables)

but after copying some files:
    running build_ext
    Searching for numpy>=1.7.1
    Reading http://pypi.python.org/simple/numpy/
    Best match: numpy 1.9.2
    Downloading https://pypi.python.org/packages/source/n/numpy/numpy-1.9.2.zip#md5=e80c19d2fb25af576460bb7dac

Is setup.py doing something wrong or it's just some other problem?

Thanks,
Carlo.

andreabedini:
Hi Carlo, thanks for trying out PyTables and thanks for reporting this. This is interesting, in this pull request I have indeed introduced a mechanism that install numpy on `build_ext` but it shouldn't install it again if it's already available. Moreover I just tested in a virtualenv with python 2.7.9 and pip 6.1.1 and I couldn't reproduce this behaviour (I also tried pip 1.5.6). I have installed numpy first with `pip install numpy` and tables next `pip install tables`. I have few questions: 1) are you installing pytables using pip or running `setup.py` ? 2) does the installation eventually succeed?

karlacio:
Hi Andrea,

some software versions I'm using:
OSX 10.7.5
pip 6.1.1
python 2.7.2

and the replies:
1) On first attempt I tried to install PyTables with pip and I got the problem I described on my previous post. Then I dowloaded PyTables 3.2.0 sources and tried to compile it with the command:

    python setup.py build_ext --inplace --hdf5 /cdvadded/hdf5

in both cases it looks like that, after:
  *running build_ext
   Searching for numpy>=1.7.1*
it assumes that numpy isn't installed.

2) in both cases the installation failed.

I have output of both installations attempts but I think they are too big to be included here.
Can I put them somewhere?

Carlo.

andreabedini:
You can upload the log to gist.guthub.com and link it here.

andreabedini:
gist.github.com (sorry I'm on my phone)

karlacio:
Here they are:

- [pip log] (https://gist.github.com/karlacio/334b22fc3508e394e8f8)
- [setup.py log] (https://gist.github.com/karlacio/56a60e5ca5823659f528)

hope I have linked them correctly.

karlacio:
Update:

I should have something of wrong on my machine.
I tried to install just tables inside a virtualenv and it worked.

Sorry for the trouble.

andreabedini:
> `SystemError: Cannot compile 'Python.h'. Perhaps you need to install python-dev|python-devel.`
definitely indicates some other problem. A virtualenv does help.

karlacio:
Looking around to solve my problem (on just one machine) I saw that also someone else with other packages had problems with setup trying to re-install numpy while it's already installed.

Even from my setup logs it looks like that PyTables setup tries to re-compile numpy and then fails.

Just a curiosity: why it tries to compile/install numpy even if it's there and of a compatible version?
Thanks

andreabedini:
@karlacio the `setup.py` shipped in 3.2.0 includes [this](https://github.com/PyTables/PyTables/blob/develop/setup.py#L55) line that installs numpy if it's not already there. It is not supposed to installed it again. I later realised that, with recent versions of `pip`, that line is not even needed and I plan to remove it soon.

If it causes problems to just, please install numpy manually and delete line 55 of `setup.py`.

PS: perhaps we should stop chatting on this PR and move this discussion to https://gitter.im/PyTables/PyTables.

karlacio:
Sure, thanks again!

PR #444, opened 2015-05-02
pkg-config support

### LINKED ISSUE #442: Build failure on recent ubuntu/debian
```
$ python setup.py build
* Using Python 2.7.9 (default, Apr  2 2015, 15:33:21) 
/usr/bin/ld: impossibile trovare -lhdf5
collect2: error: ld returned 1 exit status
.. ERROR:: Could not find a local HDF5 installation.
   You may need to explicitly state where your local HDF5 headers and
   library can be found by setting the ``HDF5_DIR`` environment
   variable or by using the ``--hdf5`` command-line option.
```

The problem is related to the new layout of the debian hdf5 package that installs libraries and headers in a hdf5/serial sub-directory:

```
/usr/include/hdf5/serial/hdf5.h
/usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so
```

I'm working on a solution based on pkg-config and I'm planning to setup  PR in a couple of days.
Alternative fixes are welcome.

avalentino:
This PR adds to the `setup.py` script the ability to use the pkg-config package to find dependencies.
By default pkg-config is always used (if available), unless explicitly disabled.

This PR should also fix #442.

### DIFF
--- a/doc/source/usersguide/installation.rst
+++ b/doc/source/usersguide/installation.rst
@@ -48,7 +48,7 @@ Prerequisites
 
 First, make sure that you have
 
-* Python_ >= 2.6 including Python 3.x
+* Python_ >= 2.6 including Python 3.x (Python >= 2.7 is highly recommended)
 * HDF5_ >= 1.8.4 (>=1.8.7 is strongly recommended)
 * NumPy_ >= 1.7.1
 * Numexpr_ >= 2.4
@@ -153,6 +153,25 @@ worry too much ;)
 
           $ C_INCLUDE_PATH=/usr/lib/openmpi/include pip install --upgrade tables
 
+    Since PyTables 3.2 can also query the *pkg-config* database to find the
+    required packages. If available, pkg-config is used by default unless 
+    another flag like `--hdf5`, `--bzip2`, etc. has been specified.
+    To force the use of *pkg-config* the `--use-pkgconfig` flag can be used::
+
+      $ python setup.py build --hdf5=/custom/hdf5/path --use-pkgconfig
+
+    or use the :envvar:`USE-PKGCONFIG` environment variable::
+
+      $ env USE_PKGCONFIG=TRUE python setup.py build --hdf5=/custom/hdf5/path
+
+    To suppress the use of *pkg-config*::
+
+      $ python setup.py build --use-pkgconfig=FALSE
+
+    or use the :envvar:`USE-PKGCONFIG` environment variable::
+
+      $ env USE_PKGCONFIG=FALSE python setup.py build
+
 **Windows**
 
     You can get ready-to-use Windows binaries and other development files for
@@ -214,6 +233,11 @@ you can proceed with the PyTables package itself.
    any extra command line arguments as discussed above::
 
       $ python setup.py build
+
+   If the HDF5 installation is in a custom path, e.g. $HOME/hdf5-1.8.15pre7,
+   one of the following commands can be used::
+
+      $ python setup.py build --hdf5=$HOME/hdf5-1.8.15pre7
 
 #. To run the test suite, execute any of these commands.
 
@@ -319,6 +343,15 @@ you can proceed with the PyTables package itself.
 
       $ python setup.py install
 
+   Again if one needs to point to libraries installed in custom paths, then
+   specific setup.py options can be used::
+
+      $ python setup.py install --hdf5=/hdf5/custom/path
+
+   or::
+
+      $ env HDF5_DIR=/hdf5/custom/path python setup.py install
+
    Of course, you will need super-user privileges if you want to install
    PyTables on a system-protected area. You can select, though, a different
    place to install the package using the :option:`--prefix` flag::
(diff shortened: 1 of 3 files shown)


FrancescAlted:
Nice! Feel free to merge it.

Enviat des del meu iPhone

El 02/05/2015, a les 11:37, Antonio Valentino <notifications@github.com> va escriure:

doc/source/usersguide/installation.rst
```diff
@@ -153,6 +153,25 @@ worry too much ;)
 
           $ C_INCLUDE_PATH=/usr/lib/openmpi/include pip install --upgrade tables
 
+    Since PyTables 3.2 can also query the *pkg-config* database to find the
```
FrancescAlted: Maybe this rephrasing would be better? "Starting from PyTables 3.2, setup.py can also query..."

setup.py
```diff
@@ -379,6 +432,7 @@ def get_hdf5_version(headername):
 # is not a good idea.
 CFLAGS = os.environ.get('CFLAGS', '').split()
 LIBS = os.environ.get('LIBS', '').split()
+USE_PKGCONFIG = None
```
FrancescAlted: It would not be better to unconditionally use pkg-config if it is found in the system?

setup.py
```diff
@@ -410,6 +470,22 @@ def get_hdf5_version(headername):
...
+    elif arg.find('--use-pkgconfig') == 0:
+        if arg == '--use-pkgconfig':
+            USE_PKGCONFIG = True
+        else:
+            USE_PKGCONFIG = arg.split('=')[1]
+
+
+if USE_PKGCONFIG is None:
+    USE_PKGCONFIG = os.environ.get('USE_PKGCONFIG', 'TRUE')
+if isinstance(USE_PKGCONFIG, str):
```
FrancescAlted: Why not making pkg-config the default (if found) unless explicitly disabled?

avalentino: My fear is that pkg-config can interfere with other options specified by the user (--hdf5, etc.).
At a first look it should not but the code is quite complex and I'm not sure I see all paths.
This is mainly a conservative choice.
I can make some additional test and enable pkg-config by default always.

doc/source/usersguide/installation.rst
avalentino: Done, thanks

setup.py
```diff
@@ -379,6 +432,7 @@ def get_hdf5_version(headername):
 # is not a good idea.
 CFLAGS = os.environ.get('CFLAGS', '').split()
 LIBS = os.environ.get('LIBS', '').split()
+USE_PKGCONFIG = os.environ.get('USE_PKGCONFIG', 'TRUE')
 
 # ...then the command line.
 # Handle --hdf5=[PATH] --lzo=[PATH] --bzip2=[PATH] --blosc=[PATH]
```
andreabedini: this comment should be updated too (or removed entirely)

setup.py
```diff
@@ -410,6 +464,20 @@ def get_hdf5_version(headername):
         # Don't delete this argument. It maybe useful for distutils
         # when adding more flags later on
         # sys.argv.remove(arg)
+    elif arg.find('--use-pkgconfig') == 0:
```
andreabedini: this whole
```
args = sys.argv[:]
for arg in args:
     if arg.find('--hdf5=') == 0:
```
is quite ugly. Can't we just use `argparse` or even `getopt` from the standard library?

setup.py
```diff
@@ -147,6 +155,7 @@ def add_from_flags(envname, flag_key, dirs):
     add_from_path("CPATH", default_header_dirs)
     add_from_path("C_INCLUDE_PATH", default_header_dirs)
     add_from_flags("CPPFLAGS", "-I", default_header_dirs)
+    add_from_flags("CFLAGS", "-I", default_header_dirs)
```
andreabedini: the logic of `setup.py` is completely obscure to me. Who reads `CFLAGS` from the environment? line 158 or 433 below?

setup.py
```diff
@@ -410,6 +464,20 @@ def get_hdf5_version(headername):
...
         # sys.argv.remove(arg)
+    elif arg.find('--use-pkgconfig') == 0:
+        if arg == '--use-pkgconfig':
+            USE_PKGCONFIG = True
+        else:
+            USE_PKGCONFIG = arg.split('=')[1]
+
+
+if isinstance(USE_PKGCONFIG, str):
+    if USE_PKGCONFIG.upper() in ('TRUE', 'YES', '1', 'ON', 'OK'):
```
andreabedini: do we really need all these variants? what about 'SI', 'OUI', 'PLEASE', 'ENABLED' ? :)

setup.py
FrancescAlted: Yes, I am equally confused.

FrancescAlted:
Will merge and will try to refine this a bit.

setup.py
FrancescAlted: I think this comment is still valid.

setup.py
FrancescAlted: Addressed in d1b1d15

setup.py
andreabedini: Ok it has to be updated to include the new option `--use-pkgconfig`

Issue #443, opened 2015-05-01
Build failure with HDF5 1.8.15pre7


avalentino:
```
$ env HDF5_DIR=$HOME/hdf5-1.8.15/ python3 setup.py build
* Using Python 3.4.3 (default, Mar 26 2015, 22:03:40) 
* Found HDF5 headers at ``/home/antonio/hdf5-1.8.15/include``, library at ``/home/antonio/hdf5-1.8.15/lib``.
.. WARNING:: Could not find the HDF5 runtime.
   The HDF5 shared library was *not* found in the default library
   paths. In case of runtime problems, please remember to install it.
* Found LZO 2 headers at ``/usr/include``, the library is located in the standard system search dirs.
* Skipping detection of LZO 1 since LZO 2 has already been found.
* Found bzip2 headers at ``/usr/include``, the library is located in the standard system search dirs.
* Could not find blosc headers and library; using internal sources.
Compiling tables/utilsextension.pyx because it changed.
Cythonizing tables/utilsextension.pyx
warning: tables/utilsextension.pyx:804:12: Non-trivial type declarators in shared declaration (e.g. mix of pointers and values). Each pointer declaration should be on its own line.
warning: tables/utilsextension.pyx:804:22: Non-trivial type declarators in shared declaration (e.g. mix of pointers and values). Each pointer declaration should be on its own line.
Compiling tables/hdf5extension.pyx because it changed.
Cythonizing tables/hdf5extension.pyx
Compiling tables/tableextension.pyx because it changed.
Cythonizing tables/tableextension.pyx
Compiling tables/linkextension.pyx because it changed.
Cythonizing tables/linkextension.pyx
Compiling tables/_comp_lzo.pyx because it changed.
Cythonizing tables/_comp_lzo.pyx
Compiling tables/_comp_bzip2.pyx because it changed.
Cythonizing tables/_comp_bzip2.pyx
Compiling tables/lrucacheextension.pyx because it changed.
Cythonizing tables/lrucacheextension.pyx
Compiling tables/indexesextension.pyx because it changed.
Cythonizing tables/indexesextension.pyx
Setting compiler flag: -march=native
running build
running build_py
creating build
creating build/lib.linux-x86_64-3.4

[...]

[...]

PR #445, opened 2015-05-02
Fix compatibility with hdf5-1.8.15-pre7

avalentino:
The implementation of H5Epush function has been modified in hdf5-1.8.15.
Now it raises an compilation warning that is turned into an error in Ubuntu 15.04 (and probably in Debian) due to compilation flags used to build the interpreter.
In the specific `-Werror=format-security`.
This PR addresses issue #443 by modifying the `blosc_filter.c` code.

Another option is to add `-Wno-format-security` to the C flags of utilsextension, hdf5extension and tableextension extensions:

```
extra_compile_args=CFLAGS + ['-Wno-format-security']
```

Note: this PR has still not been submitted upstream because I currently have a build problem with c-blosc and can't test the patch properly against the upstream master.

### DIFF
--- a/c-blosc/hdf5/blosc_filter.c
+++ b/c-blosc/hdf5/blosc_filter.c
@@ -20,7 +20,7 @@
 
 #if H5Epush_vers == 2
 /* 1.8.x */
-#define PUSH_ERR(func, minor, str) H5Epush(H5E_DEFAULT, __FILE__, func, __LINE__, H5E_ERR_CLS, H5E_PLINE, minor, str)
+#define PUSH_ERR(func, minor, str...) H5Epush(H5E_DEFAULT, __FILE__, func, __LINE__, H5E_ERR_CLS, H5E_PLINE, minor, str)
 #else
 /* 1.6.x */
 #define PUSH_ERR(func, minor, str) H5Epush(__FILE__, func, __LINE__, H5E_PLINE, minor, str)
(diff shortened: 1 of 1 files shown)


andreabedini:
Thanks for looking into this @avalentino. I had a look at the release notes for 1.8.15-pre7 but I didn't find anything mentioning H5Epush, weird. Do you have any detail on what changed?
Very happy you decided to do the right thing and fix the problem rather than silence the warning! :trollface: You marked this as _in progress_, what remains to be done?

avalentino:
> Thanks for looking into this @avalentino. I had a look at the release notes for 1.8.15-pre7 but I didn't find anything mentioning H5Epush, weird. Do you have any detail on what changed?

I compared the source code of HDF5 1.8.14 and 1.8.15 and it seems that they simplified a bit code logic in H5Epush2. Unfortunately this changes a little bit the behaviour of some compiler with that code.

avalentino:
> You marked this as in progress, what remains to be done?

The patch is done, just needs a review and merge.

Regarding the *in progress* label it is not me to set it (also n other issues).
It seems it is automatically set when issues are opened/closed (not totally clear the logic).
For some reason the change is marked as an action on my part but it is not.
I should probably file an issue against github ...

andreabedini:
ok, looks good to me, merging

andreabedini:
was ok to merge right @FrancescAlted ?

FrancescAlted:
Well, technically this should have been fixed in c-blosc instead and produce a new c-blosc release before being merged back into pytables.  I think Antonio already tried this in c-blosc, but master being currently broken (https://github.com/Blosc/c-blosc/issues/115), he decided to push a PR here.  We could either let things as they are now and move the PR into c-blosc, or remove the PR, and release a new c-blosc to be merged back.

On the other hand, I have been thinking on which version of c-blosc we should include in PyTables.  The thing is that, starting from c-blosc 1.5.0, it has lost the global lock (this is a feature, not a bug :), so it is the calling code who must be responsible for setting up the lock.  Unfortunately PyTables always release the GIL before calling HDF5 primitives (and in particular, the compression code), so I decided to  integrate in PyTables the c-blosc latest pre 1.5.0 (1.4.1).  However, as PyTables does have support for external c-blosc libraries, later or sooner there will be users using external c-blosc > 1.5.0, and they will have problems.  That requires a bit more of thinking, so I will open a ticket on this one.

FrancescAlted:
Added a new ticket on the c-blosc version to use here: https://github.com/PyTables/PyTables/issues/448

FrancescAlted:
Added fix in c-blosc: https://github.com/Blosc/c-blosc/commit/d3ad508a17c889f78c7e782538be36ff0413dcc8

andreabedini:
Well, I guessed @avalentino's patch could be pushed to c-blosc upstream. I think which blosc to use requires more discussion and should not interfere with 3.2 release. Discuss on #448.

Issue #462, opened 2015-05-27
AttributeSet: the ``/`` character is not allowed in object names: '/a/b'


mythsmith:
It is not possible to set an attribute name containing a slash character:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/dist-packages/tables/attributeset.py", line 537, in __setitem__
    self.__setattr__(name, value)
  File "/usr/lib/python2.7/dist-packages/tables/attributeset.py", line 436, in __setattr__
    check_name_validity(name)
  File "/usr/lib/python2.7/dist-packages/tables/path.py", line 93, in check_name_validity
    "in object names: %r" % name)
ValueError: the ``/`` character is not allowed in object names: '/a/b'

But, if you create a node having such an attribute with another tool (hdfview), you will be able to read it:
>>> f.root.a._v_attrs
/a._v_attrs (AttributeSet), 5 attributes:
   [/a/b/c := array([''], 
      dtype='|S1'),
    CLASS := 'GROUP',
    TITLE := '',
    VERSION := '1.0',
    a := 1]

I see no need to block "/" characters in attribute names.

avalentino:
Hi Daniele, nice patch.
Can you please add a small unit test and submit a PR?

andreabedini:
as long as HDF5 accepts slashes in attributes names, we should do too.

mythsmith:
I'm sorry I'm a beginner on github. Just have patience, I'll shortly come up with pull request (and test cases).

andreabedini:
@mythsmith take you time Daniele, no rush :)

andreabedini:
closed by #464

PR #463, opened 2015-05-27
Allow slashes in attribute names

mythsmith:
Added allow_slash=False parameter to tables.path.check_name_validity.
Pass allow_slash=True when checking attribute names in AttributeSet.__setattr__.
Added doctest for tables/path.py
Added attribute names test in test_attributes.py.
issue #462


andreabedini:
This looks good @mythsmith. Perhaps we should consider accepting slashes by default if they are accepted in HDF5. That shouldn't rise compatibility problems has it's a super set of what was previously accepted. What do you think @avalentino?

mythsmith:
I made a little bit of confusion with temporary files, adding them by mistake and later removing them.
Will all this random stuff be added to the develop branch if the pr is accepted? Or just the final delta?

avalentino:
> This looks good @mythsmith. Perhaps we should consider accepting slashes by default if they are accepted in HDF5. That shouldn't rise compatibility problems has it's a super set of what was previously accepted. What do you think @avalentino?

I didn't checked if there is some particular issue with slashes in names in HDF5, I guess there is not.

In PyTables there is the question of natural naming, but I don't remember if it also apples to attributes, we should check.

Apart from the above considerations I don't think that relaxing check on naming rules can cause issues.

mythsmith:
NaturalNaming applies also to attributes, whenever possible. This is true for any invalid character for python names (.-:,;$#%^\... etc...), which will just print a warning if used to create a new attribute. The only exception is the slash "/" (and not "\"), which raises an exception. 
This is just because the same function used to check attributes names (check_name_validity) happened to be used also to check node names, which cannot contain "/" as it is reserved for hierarchy walking purposes. 
As attributes does not contain any hierarchy, "/" is not needed for any special purpose.

Consider also that attribute names containing slashes are already red and modified correctly in current pytables, but they cannot be created (see #462).

andreabedini:
good work @mythsmith. I'm definitely convinced we should accept names including slashes when creating attributes. Now, I am not sure about the new parameter `allow_slash`. I would says `check_name_validity` signature should be rather 

`check_name_validity(name, is_attribute_name = True)`

which is self documenting. Otherwise we could split `check_name_validity` in two (given we have two different rules for node names and attribute names). @mythsmith what do you think? are you ok with renaming the parameter?

Usually we don't care too much about cluttering the git history but since you ask :) you can _squash_ your commits together into a single commits. This changes the history of your branch (since it replaces some commits rather than appending them) and you will need to force your git push to this branch. You can google all this stuff.

mythsmith:
I splitted in two functions: check_attribute_name and check_name_validity. The second simply calls the first, then adds a couple of checks.

I ended up doing a new pull request with cleaned history: #464

PR #467, opened 2015-06-12
Python3 Unicode string support in Atom.from_dtype

tp199911:
In Python3 strings show up as dtype e.g. `'<U50'`. The current code only looks for a Python2-style `'S'`-kind, resulting in
```python
    ValueError: unknown type: 'str1600'
```
upon calling e.g. `create_array` with a string array as `obj=` argument, since this case currently gets handed over to `from_type`.

### DIFF
--- a/doc/source/MIGRATING_TO_3.x.rst
+++ b/doc/source/MIGRATING_TO_3.x.rst
@@ -35,7 +35,7 @@ In Python 3, all strings are natively in Unicode. This introduces some
 difficulties, as the native HDF5 string format is not Unicode-compatible. 
 To minimize explicit conversion troubles when writing, especially :ref:`when 
 creating data sets from existing Python objects <create-signatures>`, string 
-objects are implicitly casted to non-Unicode for HDF5 storage. To make you
+objects are implicitly cast to non-Unicode for HDF5 storage. To make you
 aware of this, a warning is raised when this happens.
 
 This is certainly no true Unicode compatibility, but mainly for convenience 
--- a/doc/source/MIGRATING_TO_3.x.rst
+++ b/doc/source/MIGRATING_TO_3.x.rst
@@ -4,6 +4,7 @@ Migrating from PyTables 2.x to 3.x
 
 :Author: Antonio Valentino
 :Author: Anthony Scopatz
+:Author: Thomas Provoost
 
 This document describes the major changes in PyTables in going from the
 2.x to 3.x series and what you need to know when migrating downstream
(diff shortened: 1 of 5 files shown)


andreabedini:
Hi @tp199911, is this PR superseded by #468? If so, please close this one.

tp199911:
Hello @andreabedini, thanks for your replies. In fact, I merged the other patch into this one, so this PR now contains both changes. So I close the other one?

andreabedini:
Hi @tp199911, this looks ok and I'm going to merge it. As a last thing, could you add a test case (in `test_types.py`) ? The test should fail without your patch. Thanks!

tp199911:
@andreabedini, it is done! Let me know if you need further work on this.

tp199911:
#### I should note that this fix has still a few issues though:
* it **does not** implement support for Unicode, it only allows use of Python3 strings, conditional on the fact that they can be casted as old-school (read: ascii) strings. If not this will generate the error
`UnicodeEncodeError: 'ascii' codec can't encode characters in position xxx: ordinal not in range(128)`
* array creation via create_array will still fail, due to the check in line 204 of `array.py`
* this might create abundant (disk) memory use, as a `U` dtype has an itemsize 4x that of the equivalent `S`. Due to the first point this extra memory usage is useless.

#### I propose the following:
A structured approach when trying to create unicode atoms or arrays:
* In the case of atoms, allow it as in the current patch, but issue a warning, that non-ascii strings will fail to save. Also, correct the itemsize (*does anyone know where this magic multiplier of 4 comes from, and if I can find it in the numpy code?* I'd rather not just put `4` in the code..)
* In the case of a passed `obj` argument in `create_?array`, catch the `UnicodeEncodeError` and throw a custom exception?

Please give your input on this.

tp199911:
@andreabedini, I have implemented my above proposal, for a bit more robust/sane string handling in python3. Let me know if you require further work before merging!

andreabedini:
any of the @PyTables/maintainers can review this?

scopatz:
Yeah, I'm on it.

scopatz:
Hi @tp199911 - I think that this looks good to me.  The only thing that it lacks is any mention that this data type is supported in the documentation.  In particular, your assumptions and limitations should also be mentioned. Sorry about the shameful amount if time it took for us to get to this. I'll happily merge this in after it is documented.

tp199911:
@scopatz, I added a note + example to the Atom docstring, as well as a section to the "Migrating to 3" text. Let me know if this suffices.

scopatz:
Docs look good, thanks @tp199911. But now a test is failing https://travis-ci.org/PyTables/PyTables/jobs/75235035#L756

tp199911:
Doctest doesn't do warnings... Fixed!

scopatz:
Funny, who knew?!

scopatz:
Thanks a ton @tp199911

PR #476, opened 2015-07-23
Allow absence of condition for append_where

### LINKED ISSUE #376: Easily append one table to another -- Feature Request
Right now, to append one whole table to another you would use something like 

    table1.append_where(dstTable=table2, condition='c1 == c1')

In other words, I have to provide some trivial condition and hope there are no NaN values that mess it up. Although, in that case I was told something like '~(col1 != col1)' would work. 

It would be nice if you could append a table using a new method, or changing an existing method so that you don't need to specify a condition.

153957:
Without a condition the entire table is appended to the destination
table.
Includes test for appending without a condition.
Fixes #376

### DIFF
--- a/tables/table.py
+++ b/tables/table.py
@@ -1538,7 +1538,7 @@ class Table(tableextension.Table, Leaf):
         return self.read_coordinates(coords, field)
 
 
-    def append_where(self, dstTable, condition, condvars=None,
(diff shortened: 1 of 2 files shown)


andreabedini:
This looks great @153957, thank you!

Issue #482, opened 2015-08-11
installation problem: some tests fail 


ziggi0703:
during the installation of PyTables I had 4 tests which have failed. Here ist the full output of tables.test():

```python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import tables
>>> tables.test()
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
PyTables version:    3.2.0.dev0
HDF5 version:        1.8.11
NumPy version:       1.9.2
Numexpr version:     2.4.3 (not using Intel's VML/MKL)
Zlib version:        1.2.8 (in Python interpreter)
BZIP2 version:       1.0.6 (6-Sept-2010)
Blosc version:       1.4.4 (2015-05-05)
Blosc compressors:   blosclz (1.0.2.1), lz4 (1.2.0), lz4hc (1.2.0), snappy (1.1.1), zlib (1.2.8)
Cython version:      0.23
Python version:      2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2]
Platform:            Linux-3.13.0-61-generic-x86_64-with-Ubuntu-14.04-trusty
Byte-ordering:       little
Detected cores:      8
Default encoding:    ascii
Default FS encoding: UTF-8
Default locale:      (en_US, UTF-8)
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
Performing only a light (yet comprehensive) subset of the test suite.
If you want a more complete test, try passing the --heavy flag to this script
(or set the 'heavy' parameter in case you are using tables.test() call).
The whole suite will take more than 4 hours to complete on a relatively
modern CPU and around 512 MB of main memory.
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
tables/filters.py:292: FiltersWarning: compression library ``lzo`` is not available; using ``zlib`` instead
  % (complib, default_complib), FiltersWarning)
[...]

PR #483, opened 2015-08-11
remove_rows can remove all rows

eumiro:
Removed the condition blocking the removal of all rows from a table. The new test passed.


FrancescAlted:
Looks good to me.  The test suite is crashing due to #482 but your new test is fine.  Merging.

FrancescAlted:
A possible improvement for the release notes would be to know when the HDF5 library fixed the issue, but probably quite long ago and we have a test, so not really worth the effort.  Thanks!

PR #484, opened 2015-08-12
Use @property decorators

andreabedini:
This patch replaces all code like

```python

def _getnrows(self):
    ...

nrows = property(_getnrows, None, None, "The number of rows in the array.")
```
with the much nicer
```python
@property
def nrows(self):
    "The number of rows in the array."
    ...
```

In few cases it seems we have been exposing both kinds of access (property and function). E.g. `_g_getnode` and `_v_node`. For the moment I kept both but only one should be kept in the future (and perhaps we should be more clear what methods are internal).


scopatz:
Looks great, thanks @andreabedini!

FrancescAlted:
Yes, this code probable dates from the pre-decorators age, so this PR was much needed, thanks.  Regarding the question about public and private functions, normally the private are prefixed with a `_g_` and public with a `_f_`.  The `_v_` is used for public variables.  Also, all public functions have been (or should have been) documented in docstrings.

PR #487, opened 2015-08-14
Fix docstrings for remove_row(s)

153957:
Move docstring about remove_rows back to it (was mistakenly moved when adding remove_row in 0c3b923).
Clarify the effect of omitting the 'stop' for remove_rows.
Clarify that step is not supported (only None or 1).
Fix a typo.
Add example to remove_row to show equivalent remove_rows arguments.

### DIFF
--- a/tables/table.py
+++ b/tables/table.py
@@ -2539,10 +2539,10 @@ class Table(tableextension.Table, Leaf):
     def remove_rows(self, start=None, stop=None, step=None):
         """Remove a range of rows in the table.
 
-        If only start is supplied, only this row is to be deleted.  If a range
-        is supplied, i.e. both the start and stop parameters are passed, all
-        the rows in the range are removed. A step parameter is not supported,
-        and it is not foreseen to be implemented anytime soon.
+        If only start is supplied, that row and all following will be deleted.
+        If a range is supplied, i.e. both the start and stop parameters are
+        passed, all the rows in the range are removed. A step parameter is not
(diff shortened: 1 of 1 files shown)


tables/table.py
```diff
@@ -2539,6 +2539,11 @@ def _add_rows_to_index(self, colname, start, nrows, lastrow, update):
     def remove_rows(self, start=None, stop=None, step=None):
         """Remove a range of rows in the table.
 
+        If only start is supplied, only this row is to be deleted.  If a range
```
avalentino: IMO the above line is not correct and it is in conflict with what stated below.

153957: Ah, correct! I will fix that line.

153957:
The step keyword is not used and the docstring claims that you should not hold your breath for support. Should the keyword be removed from the function until it is actually implemented?

avalentino:
Thanks @153957.

> The step keyword is not used and the docstring claims that you should not hold your breath for support. Should the keyword be removed from the function until it is actually implemented?

Looking at the code it seems that support for the `step` parameter is implemented now.
I didn't found unit tests for it anyway.

153957:
I tried using the step argument with a value other than `1` or `None` and it resulted in Exceptions.

avalentino:
Thanks @153957.
Should be fixed in 35f12684dd81564f8

153957:
Cool! thanks.

Issue #477, opened 2015-07-28
AssertionError in Row.__init_loop when using Table wrapper with indexing and column type or precision mismatch


kaukrise:
Iterating over rows fails in 3.2.0 in classes wrapping Table, but only under certain conditions, as far as I can tell:

* The columns have different types or numeric precision, e.g. Int32Col and Int16Col
* One column is indexed

Here is a minimal example that fails with an AssertionError:

```python
import tables as t

class MinTable(t.Table):
    
    def __init__(self):
        
        f = t.open_file('bla', 'a', driver="H5FD_CORE",driver_core_backing_store=0)
        
        description = {
            'c1': t.Int32Col(),
            'c2': t.Int16Col()
        }
        
        t.Table.__init__(self, f.get_node('/'), 'test', description=description)
        
        self._enable_index()
        
        row = self.row
        row['c1'] = 0
        row.append()
        self.flush()
        
    def _enable_index(self):
        self.cols.c1.create_index()

m = MinTable()
# next line fails
[x.fetch_all_fields() for x in m.where('c1 == 0')]
```

The error produced is:

```
/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tables/table.pyc in where(self, condition, condvars, start, stop, step)
   1499         """
   1500
-> 1501         return self._where(condition, condvars, start, stop, step)
   1502
   1503     def _where(self, condition, condvars, start=None, stop=None, step=None):

/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tables/table.pyc in _where(self, condition, condvars, start, stop, step)
   1538         if profile:
   1539             show_stats("Exiting table._where", tref)
-> 1540         return row._iter(start, stop, step, chunkmap=chunkmap)
   1541
   1542     def read_where(self, condition, condvars=None, field=None,

tables/tableextension.pyx in tables.tableextension.Row._iter (tables/tableextension.c:9112)()

tables/tableextension.pyx in tables.tableextension.Row._init_loop (tables/tableextension.c:10137)()

AssertionError:
```

Changing Int32Col to Int16Col or vice versa produces no error. Also no problem occurs when switching off indexing. This does not happen in the normal Table class, as far as I can tell. The following works fine:

```python
import tables as t

f = t.open_file('bla2', 'a', driver="H5FD_CORE",driver_core_backing_store=0)
[...]

Issue #409, opened 2014-12-08
table.where query does not seem to be able to find rows...


guziy:
Hi:

I am creating a big table but I cannot find the rows which are supposed to be there using ```table.where```. But ```table.get_where_list()``` does find the rows in question (see the screenshot, **Note that there is no output as a result of the loop in table.where**). 

![screen shot 2014-12-08 at 17 58 36](https://cloud.githubusercontent.com/assets/900941/5349065/df8c3dca-7f03-11e4-8c41-1b14faf5a40b.png)

I was trying to reproduce it with smaller dummy files but no luck... Even when I select a table using ptrepack, the problem goes away... 

Not indexing columns also seems to fix the problem, except in this case querying is very slow....

alexfields:
Possibly related to #319, pydata/pandas#8265, and pydata/pandas#9676

FrancescAlted:
This has been hopefully fixed.  See #319 for details.

Issue #319, opened 2014-01-13
BUG: possible indexing/selection bug


jreback:
```
In [10]: tables.print_versions()
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
PyTables version:  3.0.0
HDF5 version:      1.8.4-patch1
NumPy version:     1.7.1
Numexpr version:   2.1 (not using Intel's VML/MKL)
Zlib version:      1.2.3.4 (in Python interpreter)
LZO version:       2.03 (Apr 30 2008)
Blosc version:     1.2.3 (2013-05-17)
Cython version:    0.17.2
Python version:    2.7.3 (default, Jun 21 2012, 07:50:29) 
[GCC 4.4.5]
Platform:          linux2-x86_64
Byte-ordering:     little
Detected cores:    12
Default encoding:  ascii
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=

```

see here: https://github.com/pydata/pandas/issues/5913

Narrowed it down to this:

create a table with a larger ``expectedrows`` that actually storing
create an index on a column
select via a pretty small start/stop range (e.g. in the below example if you use a chunksize of 1M, then it doesn't show up, but 500k makes it fail).

If I don't pass ``expectedrows``, then this works as expected!

Code to reproduce:
```
import numpy as np
import tables

def pr(result):
    print np.unique(result['o'])

n = 7000000
arr = np.zeros((n,),dtype=[('index', 'i8'), ('o', 'i8'), ('value', 'f8')])
arr['index'] = np.arange(n)
arr['o'] = np.random.randint(-20000,-15000,size=n)
arr['value'] = value = np.random.randn(n)

handle = tables.openFile('test.h5','w',filters=tables.Filters(complevel=9,complib='blosc'))
node   = handle.createGroup(handle.root, 'foo')
table  = handle.createTable(node, 'table', dict(
    index   = tables.Int64Col(),
    o   = tables.Int64Col(),
    value  = tables.FloatCol(shape=())),
                            expectedrows=10000000)

table.cols.index.createIndex()
table.append(arr)
handle.close()

v1 = np.unique(arr['o'])[0]
v2 = np.unique(arr['o'])[1]
[...]

PR #489, opened 2015-08-20
Make _calc_nrowsinbuf in tables/leaf.py also check for classes inheriting Table

kaukrise:
A previous fix (for issues #319 and #409) for the size of nrowsinbuf calculated in _calc_nrowsinbuf was already checking specifically for the Table class. However, classes inheriting from Table encounter the same problem. This commit also fixes nrowsinbuf size for classes inheriting Table (issue #477).

### DIFF
--- a/tables/leaf.py
+++ b/tables/leaf.py
@@ -370,8 +370,7 @@ class Leaf(Node):
         buffersize = params['IO_BUFFER_SIZE']
         if rowsize != 0:
             nrowsinbuf = buffersize // rowsize
-            if (self.__class__.__name__ == "Table" or
-                "Table" in [base.__name__ for base in self.__class__.__bases__]):
+            if self.__class__.__name__ == "Table":
                 # The number of rows in buffer needs to be an exact multiple of
                 # chunkshape[0] for queries using indexed columns.
                 # Fixes #319 and probably #409 too.
@@ -383,8 +382,7 @@ class Leaf(Node):
         # to make sure nrowsinbuf is greater than or
         # equal to the chunksize.
         # See gh-206 and gh-238
-        if (self.chunkshape is not None and (self.__class__.__name__ == "Table" or
-            "Table" in [base.__name__ for base in self.__class__.__bases__])):
+        if self.chunkshape is not None and self.__class__.__name__ == "Table":
             if nrowsinbuf < self.chunkshape[0]:
                 nrowsinbuf = self.chunkshape[0]
 
--- a/tables/leaf.py
+++ b/tables/leaf.py
@@ -370,21 +370,8 @@ class Leaf(Node):
         buffersize = params['IO_BUFFER_SIZE']
         if rowsize != 0:
             nrowsinbuf = buffersize // rowsize
-            if self.__class__.__name__ == "Table":
-                # The number of rows in buffer needs to be an exact multiple of
-                # chunkshape[0] for queries using indexed columns.
-                # Fixes #319 and probably #409 too.
-                nrowsinbuf -= nrowsinbuf % self.chunkshape[0]
         else:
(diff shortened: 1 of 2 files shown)


avalentino:
Probably a better solution would be using 

```
isinstance(self, Table)
```

or, even better, to write a specialized method in the Table class.

kaukrise:
That sounds good and was my first intuition, too. But seeing that the existing code used a string comparison for checking the class, I assumed that approach was used for a good reason and mimicked it in my fix.

kaukrise:
Could the motivation for using a string-based comparison in the first place have been to avoid a circular dependency? Adding the import
```python
from tables.table import Table
```
to [leaf.py](https://github.com/PyTables/PyTables/blob/fb9b917142996e7dce427af6b46e3e64b36d8107/tables/leaf.py) and doing
```python
import tables
```
results in an ImportError:
```python
In [1]: import tables
ImportError                               Traceback (most recent call last)
<ipython-input-1-389ecae14f10> in <module>()
----> 1 import tables

.../PyTables/tables/__init__.pyc in <module>()
    117 # Import the user classes from the proper modules
    118 from tables.exceptions import *
--> 119 from tables.file import File, open_file, copy_file
    120 from tables.node import Node
    121 from tables.group import Group

.../kaukrise/PyTables/tables/file.py in <module>()
     45 from tables.filters import Filters
     46 from tables.node import Node, NotLoggedMixin
---> 47 from tables.group import Group, RootGroup
     48 from tables.group import TransactionGroupG, TransactionG, MarkG
     49 from tables.leaf import Leaf

.../PyTables/tables/group.py in <module>()
     27 from tables.path import check_name_validity, join_path, isvisiblename
     28 from tables.node import Node, NotLoggedMixin
---> 29 from tables.leaf import Leaf
     30 from tables.unimplemented import UnImplemented, Unknown
     31

.../PyTables/tables/leaf.py in <module>()
     25 from tables.exceptions import PerformanceWarning
     26 from tables import utilsextension
---> 27 from tables.table import Table
     28
     29 def csformula(expected_mb):

.../PyTables/tables/table.py in <module>()
     32 from tables.flavor import flavor_of, array_as_internal, internal_to_flavor
     33 from tables.utils import is_idx, lazyattr, SizeType, NailedDict as CacheDict
---> 34 from tables.leaf import Leaf
     35 from tables.description import (
     36     IsDescription, Description, Col, descr_from_dtype)

ImportError: cannot import name Leaf
```

I don't want to mess too much with the PyTables structure, so I think this prevents me from using `isinstance(self, Table)` in leaf.py.

Perhaps you could elaborate a bit more on the option of adding a dedicated method in the Table class?

avalentino:
The idea is to write a  specialized `_calc_nrowsinbuf` method for the `Table` class in `table.py` that addresses the specific case of computing the number of rows in the buffer for tables.
The generic version of `_calc_nrowsinbuf`, in `leaf.py`, will no longer need to use string comparison to detect Table instances and make special computation for that case.

andreabedini:
:+1: for what @avalentino says

kaukrise:
@avalentino That is a much neater idea, I'll work on it when I have some time on my hands. Thanks for the support, PyTables is a fantastic project, wouldn't know how to do my research efficiently without it!

kaukrise:
So, here is a pretty naive way of implementing the solution suggested by @avalentino. It duplicates a lot of code from the original Leaf method, which is definitely sub-optimal, but my attempts to reduce the amount of duplication (by adding another private method for the actual specifics of the calculation) wasn't very effective and just seemed to trade one evil for another. Looking forward to your comments and suggestions!

avalentino:
thanks @kaukrise

Issue #478, opened 2015-08-06
File.get_node() can return node from another file


153957:
When using `File.get_node` with a Node instance from another file as where argument it returns the Node from the other file.

For example a file `this.h5` which contains a group called `some` with a tables called `thing`, and an empty file called `other.h5`.
```python
data = tables.open_file('this.h5', 'w')
data2 = tables.open_file('other.h5', 'w')
bla = data.get_node('/some')
bla2 = data2.get_node(bla)
print bla2._v_file
# ... this.h5 ...
bla2 = data2.get_node(bla, 'thing')
# NoSuchNodeError
```

So if you only pass a where argument it might not return a node from the expected file, if you also pass a name argument it will look in the correct file.

I was expecting is that it would use the `_v_pathname` of the node to look that up in the file upon which `get_node` is called.
i.e. `file.get_node(node) == file.get_node(node._v_pathname)`

andreabedini:
Thanks for the report @153957

andreabedini:
> I was expecting is that it would use the _v_pathname of the node to look that up in the file upon which get_node is called.
> i.e. file.get_node(node) == file.get_node(node._v_pathname)

no, it's more complex because there's a cache for opened nodes. Basically `File.get_node(where)` returns just `where` when it's a node object.

153957:
Ok, so if I want to look for a node with the same path as a node from another file I will have to use its `_v_pathname` as string instead of the other node object itself.
With your commit `get_node` is now consistent whether you do or do not use the name keyword, i.e. the type (`Node` or `str`) of the `where` argument determines which file is actually being used, in case of `Node` the file the node belongs to, in case of `str` the file the `get_node` is being called on.
I think that the docstring for get_node needs to be clarified a little bit about the effect of using a `Node` as where argument.

andreabedini:
> With your commit get_node is now consistent whether you do or do not use the name keyword, i.e. the type (Node or str) of the where argument determines which file is actually being used, in case of Node the file the node belongs to, in case of str the file the get_node is being called on.

I thought it's natural behaviour (compatible with tests). `get_node` is also used when copying nodes from one file to another so it can be called with `where` in a different file.

> I think that the docstring for get_node needs to be clarified a little bit about the effect of using a Node as where argument.

You are right, I'll reopen and add the documentation label.

153957:
See #491. Is that ok?

andreabedini:
I merged #491, fantastic work.

PR #491, opened 2015-08-27
Improve docstring for get_node

153957:
Clarify what happens when where is a Node (from a different file). See #478


andreabedini:
Looks great @153957 thanks for your effort!

PR #495, opened 2015-08-31
Rework setup.py

andreabedini:
- Use setuptools_scm to manage package version
- Move version info from the cython extension to the python side
- Use setuptools built in cython support rather than reimplement our own


andreabedini:
This fails miserably because we are using a non standard scheme for tags (`v.3.2.1` rather than `v3.2.1` or `3.2.1`). Maybe we should just do a mass retag :)

andreabedini:
Obsoleted by 7994c6bc361335c7b754660f18f47e893ce4de01.

Issue #43, opened 2011-06-05
indexing should permit uint as well as int


joshmoore:
I have a column of type uint32 in one table containing values for row-numbers in another table. Currently, I am forced to cast this to int32 before I can use it as an index. I think that makes no sense, logically, so I set the type of this issue to defect rather than enhancement.

```python
In [1]: import tables

In [3]: t = tables.openFile('tutorial1.h5')

In [8]: t.root.detector.readout[np.array([2, 3, 4], dtype=np.int32)]
Out[8]: 
array([(512, 2, 256.0, 2, 8, 34359738368, 'Particle:      2', 4.0),
       (768, 3, 6561.0, 3, 7, 51539607552, 'Particle:      3', 9.0),
       (1024, 4, 65536.0, 4, 6, 68719476736, 'Particle:      4', 16.0)], 
      dtype=[('ADCcount', '<u2'), ('TDCcount', '|u1'), ('energy', '<f8'), ('grid_i', '<i4'), ('grid_j', '<i4'), ('idnumber', '<i8'), ('name', '|S16'), ('pressure', '<f4')])

In [9]: t.root.detector.readout[np.array([2, 3, 4], dtype=np.uint32)]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)

/storage4/home/gerrit/checkouts/pytables-2.2/examples/<ipython console> in <module>()

/storage4/home/gerrit/.local/lib/python2.6/site-packages/tables/table.pyc in __getitem__(self, key)
   1709         # Try with a boolean or point selection

   1710         elif type(key) in (list, tuple) or isinstance(key, numpy.ndarray):
-> 1711             coords = self._pointSelection(key)
   1712             return self._readCoordinates(coords, None)
   1713         else:

/storage4/home/gerrit/.local/lib/python2.6/site-packages/tables/leaf.pyc in _pointSelection(self, key)
    587                 coords = numpy.asarray(key, dtype="i8")
    588         else:
--> 589             raise TypeError("Only integer coordinates allowed.")
    590         # We absolutely need a contiguous array

    591         if not coords.flags.contiguous:

TypeError: Only integer coordinates allowed.
```

It's probably a very easy fix, I might have a look at it this weekend.

153957:
This issues has been fixed at some point. Using an array as index works with dtype=uint32, and even dtype=uint64.

However, using a single uint64 value does not work, see #475.

andreabedini:
Should be fixed by #496.

PR #496, opened 2015-09-02
Cast selection indices to integer

### LINKED ISSUE #475: Failure to retrieve a row using a numpy.uint64 as index since PyTables 3.2
Since Pytables 3.2 part of our analysis scripts fail.
It seems we try to use a `numpy.uint64` to retrieve a row from a table, this is apparently no longer supported, it used to work in v3.1. Using a normal `int`, `numpy.uint32` or `numpy.int64` does still work.
Is this intentional? A `uint64` should also be castable to a Python `int`..

With PyTables 3.1:
```python
import tables
data = tables.open_file('muonlab.h5', 'r')
data.root.station_501.events[11]
# (11L, 1433268013, 423254769L, 1433268013423254769L, [316, 417, 6, 5], [2824, 6445, 0, 0], 1.083899974822998, 2.317500114440918, 0.0, 0.0, 995.0, 1180.0, -999.0, -999.0, 1187.5)
data.root.station_501.events[numpy.uint64(11)]
# (11L, 1433268013, 423254769L, 1433268013423254769L, [316, 417, 6, 5], [2824, 6445, 0, 0], 1.083899974822998, 2.317500114440918, 0.0, 0.0, 995.0, 1180.0, -999.0, -999.0, 1187.5)
```

With PyTables 3.2:
```python
import tables
data = tables.open_file('muonlab.h5', 'r')
data.root.station_501.events[11]
# (11L, 1433268013, 423254769L, 1433268013423254769L, [316, 417, 6, 5], [2824, 6445, 0, 0], 1.083899974822998, 2.317500114440918, 0.0, 0.0, 995.0, 1180.0, -999.0, -999.0, 1187.5)
data.root.station_501.events[numpy.uint64(11)]
# ---------------------------------------------------------------------------
# TypeError                                 Traceback (most recent call last)
# <ipython-input-2-64290e3841be> in <module>()
# ----> 1 data.root.station_501.events[numpy.uint64(11)]
# 
# tables/table.pyc in __getitem__(self, key)
#    2112                 key += self.nrows
# -> 2113             (start, stop, step) = self._process_range(key, key + 1, 1)
#    2114             return self.read(start, stop, step)[0]
# 
# tables/leaf.pyc in _process_range(self, start, stop, step, dim, warn_negstep)
#     415 
# --> 416         return slice(start, stop, step).indices(long(nrows))
#     417 
# 
# TypeError: slice indices must be integers or None or have an __index__ method
```

Possibly related #447 

(original issue HiSPARC/sapphire#76)

andreabedini:
This is a bit weird, it was prompted by #475 and #43.

The problem (or part of it) lies in that if

    key  = numpy.uint64(10)

`is_idx(key)` will return `True` but then

    self._process_range(key, key + 1, 1)

fails because `key + 1 == 11.0` (try it yourself).

The uptake is that indices need to be casted in some way. I decided to use
`operator.index` because that's what Python uses internally when something is
used as an index.

Interestingly enough #43 got fixed accidentally: `Array._interpret_indexing`
fails because `_process_range` fails, just like above. But then
`Array.__getitem__` goes on trying different strategies and finally
`_fancy_selection` does the cast.

This PR is for the upcoming 3.2.2

Closes #475, #43

### DIFF
--- a/tables/array.py
+++ b/tables/array.py
@@ -12,6 +12,7 @@
 
 """Here is defined the Array class."""
 
+import operator
 import sys
 
 import numpy
@@ -394,6 +395,8 @@ class Array(hdf5extension.Array, Leaf):
                 raise IndexError("Too many indices for object '%s'" %
                                  self._v_pathname)
             elif is_idx(key):
+                key = operator.index(key)
+
                 # Protection for index out of range
                 if key >= self.shape[dim]:
                     raise IndexError("Index out of range")
--- a/tables/index.py
+++ b/tables/index.py
@@ -13,13 +13,15 @@
 """Here is defined the Index class."""
 
 from __future__ import print_function
-import sys
-from time import time, clock
+import math
+import operator
 import os
 import os.path
+import sys
 import tempfile
-import math
 import warnings
+
+from time import time, clock
 
 import numpy
 
@@ -1787,6 +1789,8 @@ class Index(NotLoggedMixin, indexesextension.Index, Group):
         """
 
         if is_idx(key):
+            key = operator.index(key)
+
             if key < 0:
                 # To support negative values
                 key += self.nelements
--- a/tables/table.py
+++ b/tables/table.py
@@ -12,12 +12,14 @@
 
 """Here is defined the Table class."""
 
-import sys
 import math
-import warnings
+import operator
 import os.path
-from time import time
+import sys
+import warnings
+
 from functools import reduce as _reduce
+from time import time
 
 import numpy
 import numexpr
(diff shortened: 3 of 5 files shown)


andreabedini:
I don't know why travis is not testing this ...

andreabedini:
I am afraid this might hurt users on platforms where a Python int is 32 bit.
@153957 does this fix your specific problem?

153957:
Are there platforms where Python ints are still 32 bit? I thought that Since Python 2.2-2.4 the Python int and long were unified (https://www.python.org/dev/peps/pep-0237/) into integers that automatically grow in memory size is required (depending on available memory).

andreabedini:
@153957 I did some tests on ubuntu trusty 32bit and this seems to work fine. 64bit indices on 32bit platforms might be already broken so we can just go ahead and wait for the users to speak (or cry). What do you think @scopatz @avalentino ?

avalentino:
It is OK for me

andreabedini:
Thanks @avalentino, I'll merge this then.

Issue #404, opened 2014-11-28
Table does not get updated with Row.update() if the loop is broken


joonro:
I'm not sure if this is correct behavior, but it seems a table does not get updated with `Row.update()` when the loop is broken. The following is a example code with `break`:

```python
import tables

class table_Updating(tables.IsDescription):
    value = tables.UInt8Col()

hdf5_a = tables.openFile('TableUpdating.hdf5', 'a')

if "Updating" in hdf5_a.root._v_children:
    hdf5_a.remove_node("/", "Updating")

table = hdf5_a.create_table("/", "Updating", table_Updating)

for i in xrange(10):
    table.row['value'] = 255
    table.row.append()

table.flush()
hdf5_a.close()
```

Then if I do the following:

```python
hdf5_a = tables.openFile('TableUpdating.hdf5', 'a')

table = hdf5_a.get_node("/", name="Updating")

for row in table:
    row['value'] = 2
    row.update()
    table.flush()

    break

print(table[:])

hdf5_a.close()
```

The first value does not get updated for some reason:

```python
[(255,), (255,), (255,), (255,), (255,), (255,), (255,), (255,), (255,), (255,)]
```

When I comment out `break`, all the rows are updated:

```python
[(2,) (2,) (2,) (2,) (2,) (2,) (2,) (2,) (2,) (2,)]
```

I tried a couple of different things, and it seems the loop has to run til the end for the update to work - I found `continue` is fine, but even if I let the loop run several times, with `break`, the table is not updated. It is not updated when the loop is broken by an exception either.

I'm using PyTables 3.1.1 with Anaconda in Windows 8.1 64bit.

FrancescAlted:
Yes, this is a know issue.  I don't remember well the details, but the problem is that when the loop is broken that way, PyTables cannot detect the event, and the internal buffers are not flushed.  But I agree that this should be fixed in one way or another, or at least, warn about this in the manual.

joonro:
Thanks for the reply. Hopefully this gets fixed soon, as this is a bug that is hard to detect for users.

itziakos:
For completeness, the related issue is #8 where a workaround is provided (flushing the row buffers before breaking).

andreabedini:
Closing as duplicated of #8

PR #500, opened 2015-09-09
+ caveat on Row.update (doesn't work with break)

joonro:
Added a caveat that Row.update does not work if the loop is broken with break statement. Related to Issue #404


andreabedini:
Thanks @joonro much appreciated

joonro:
You're welcome!

PR #503, opened 2015-09-17
get rid of -native

### LINKED ISSUE #458: Illegal instruction error
I got `Illegal instruction` error when executing the code below in docker with OS X and ubuntu. However, it works fine using CoreOS with docker and OS X and ubuntu directly. The error seems only happens in version 3.2.0. I solved the issue by downgrading pytables to 3.1.1.
Any idea for this?

``` python
import tables
tables.open_file('a.h5', 'w')
```

jennolsen84:
VirtualBox does not support SSE4 or POPCNT yet, so this will prevent a build on a new machine from inserting newish instructions.  Fixes #458

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -738,7 +738,7 @@ if 'BLOSC' not in optional_libs:
         finally:
             os.remove(fd.name)
 
-    try_flags = ["-march=native", "-msse2"]
+    try_flags = ["-msse2"]
     for ff in try_flags:
         if compiler_has_flags(compiler, [ff]):
             print("Setting compiler flag: " + ff)


andreabedini:
@jennolsen84 thanks for looking into #458. If I understand correctly this PR implies we won't be using SSE4 when available and therefore it's not really a good option. As I have suggested in #458, pip's option `--no-binary` should solve the problem. pip wheels are not really portable, I am not sure there's a general solution to this.

andreabedini:
There seems to be consensus for removing both flags -march and -msse2. So we can remove the whole block 

```
try_flags = ["-march=native", "-msse2"]
for ff in try_flags:
    ...
```

@jennolsen84 could you take care of doing this in your branch?

FrancescAlted:
Well, if you remove the '-msse2' flag then Blosc performance will decrease by a significant amount (2x ~ 4x) .  The solution should be using Blosc >= 1.7 but https://github.com/Blosc/hdf5-blosc/issues/7 must be fixed first.

andreabedini:
@FrancescAlted We won't release without blosc 1.7 and fixing Blosc/hdf5-blosc#7 first. Don't worry.

FrancescAlted:
Ok.  But now I am thinking that even if Blosc >= 1.7 is included in PyTables sources, you will need at least the "-march=native" or "-msse2" so as to use the SSE2 extensions.  And that means that the binaries won't run on systems with no SSE2 instructions (mainly VMs), not because of Blosc, but because the compiler can generate SSE2 code in other places of the code (Cython extensions) as well.  Tricky...

andreabedini:
perhaps users can always recompile with `CFLAGS=-march=native`?

FrancescAlted:
Ah, yes, recompiling with `CFLAGS=-march=native` should be the solution, right (another thing is how binaries are distributed, but this is not our direct concern).

jennolsen84:
OK, so remove both SSE2 and native flags?

As a first pass, we could add a runtime check by checking cpu flags somewhere and warn the user of non-optimal performance when it is run on a CPU with SSE2/SSE4/AVX etc.

Does that sound good?

Another option would be to build multiple binaries (with different optimizations), and then update a list of function pointers on the fly (perhaps on __init__.py ? ).  Then you'd always have the best implementation.  Granted this is a lot more work, and would take much longer.  CPUs are pretty good at optimizing the function tables (at the hardware level), so performance should be very close to optimal

FrancescAlted:
PR #541 could fix this.  @jennolsen84 can you have a try at it and tell us how it goes?

PR #508, opened 2015-10-02
Add lib64 to posix search path. Fixes #507.

### LINKED ISSUE #507: Add lib64 to setup search path
I have installed hdf5 library in a custom path and I realized that setup script is not able to find the library regardless of setting `HDF5_DIR` or any other environment variable. Looking at the script I added [lib64 folder to search path](https://github.com/PyTables/PyTables/blob/develop/setup.py#L310) and everything worked fine.

Please add `lib64` directory to default search path while looking for `libhdf5` shared object.

mehdisadeghi:
Add lib64 to posix search path. Fixes #507.


FrancescAlted:
Seems like a good idea.  Merging.

Issue #518, opened 2015-12-26
Support reference arrays


dashesy:
MATLAB starting version 7.3 generates `hdf5` file format. My guess is that in order to record UTF8 strings it uses an array of references, but it seems even some double formatted values are used as references. 
I get an error for this data type:

    UserWarning: leaf ``/ANN/my_arr`` is of an unsupported type; it will become an ``UnImplemented`` node

In the original file generated by MATLAB I also get error that as a result of VL strings in attribute (issue #48) but that will not affect the reading. Lack of support for reference arrays however means I have to use `h5py` but handling many object references is very very slow if done one value at a time. I also looked at [hdf5storage](https://github.com/frejanordsiek/hdf5storage) but it is very slow handling the reference arrays.

Here is a dump of the file:

```text
HDF5 "Desktop\\test.h5" {
GROUP "/" {
   GROUP "#refs#" {
      ATTRIBUTE "H5PATH" {
         DATATYPE  H5T_STRING {
            STRSIZE 7;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "/#refs#"
         }
      }
      DATASET "a" {
         DATATYPE  H5T_STD_U64LE
         DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
         DATA {
         (0): 0, 0
         }
         ATTRIBUTE "MATLAB_class" {
            DATATYPE  H5T_STRING {
               STRSIZE 15;
               STRPAD H5T_STR_NULLTERM;
               CSET H5T_CSET_ASCII;
               CTYPE H5T_C_S1;
            }
            DATASPACE  SCALAR
            DATA {
            (0): "canonical empty"
            }
         }
         ATTRIBUTE "MATLAB_empty" {
            DATATYPE  H5T_STD_U8LE
            DATASPACE  SCALAR
            DATA {
            (0): 1
            }
         }
      }
      DATASET "b" {
         DATATYPE  H5T_STD_U64LE
         DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
         DATA {
         (0): 0, 0
         }
         ATTRIBUTE "H5PATH" {
            DATATYPE  H5T_STRING {
               STRSIZE 9;
               STRPAD H5T_STR_NULLTERM;
               CSET H5T_CSET_ASCII;
               CTYPE H5T_C_S1;
            }
[...]

PR #519, opened 2015-12-31
Read reference types by dereferencing them as numpy arrays of object.

dashesy:
MATLAB v7.3 generates hdf5 file but uses references to save cell arrays.
cell arrays are like Python lists, they can store simple objects like strings or arrays, or other cell arrays.
The intention is to be able to *only* read newer .mat files, that can be then saved as PyTable format.

Other planned PRs:
1. Read variable length strings that also are in in .mat files
2. Determine MATLAB file format in the beginning and use MATLAB_class to make sense of strings

I appreciate any guidance in general direction for this as well as planned PRs.

This should fix issue #518


avalentino:
Hi @dashesy, nice enhancement, thank you.
It would be very appreciated if you could also add some unit test.

dashesy:
@avalentino added unit tests and 2 test files. One is a cell array inside a struct, the other is a cell array inside another cell array.

1. What do you think about adding limited support for reading VL strings in a similar way (boxing them as objects or perhaps `numpy.string_`)? 
2. What do you think about limited support for `MATLAB_class` attributes, at least for strings to return `numpy.string_` instead of objects?

avalentino:
Thank you very much @dashesy.
Sorry for taking so much time to merge.

Issue #529, opened 2016-03-11
Setting cache slots parameter less than 10 never removes items from the cache


grahamrjuk:
If I set a cache slots parameters to a value less than 10 - for example ```COND_CACHE_SLOTS``` - the number of items the ```NailedDict``` decides to remove is 0 (integer value of 10% of the parameter).

The cache then grows forever.

It is possible that such a small cache is not considered sensible but I don't see any documentation saying there is a minimum.

Suggest making the calculation of the number to remove always remove 1 cache entry, instead of none, or set a minimum cache slots of 10.

FrancescAlted:
Good point.  I am +1 on always removing 1 cache entry at least and +0 for a minimum cache slots.  Would you mind providing a PR?

FrancescAlted:
Fixed in #530.

PR #530, opened 2016-03-11
Ensure cache entries are removed if fewer than 10 entries in the cache - fix for #529

grahamrjuk:
Set ```entries_to_remove``` to ```1``` if ```self.maxentries // 10 < 0``` (using ```max```)


FrancescAlted:
Excellent.  Thanks!

Issue #534, opened 2016-03-29
Selecting rows using 'log10' in the query returns wrong results


153957:
I have PyTables (v3.2.2) running on multiple systems (OS X, Scientific Linux CERN, and Cent OS).
On the Mac numpy 1.9.4 and numexpr 2.4.4 are installed
On the SLC machine numpy 1.10.4 and numexpr 2.4.6 are installed.

I have a table (`sims`) with a column (`energy`) containing values ranging from `1e12` to `1e18`. Most values are at full powers of ten (e.g `1e13` and `1e14`) or in between (e.g. `10**13.5` and `10**15.5`). So to easily query these values I use the `log10` option for PyTables queries. The following query works on the Mac, but returns the wrong result on the SLC machine:

```python
import numpy
numpy.log10(sims.read_where('log10(energy) == 12')['energy'][:9])
# On Mac: array([ 12.,  12.,  12.,  12.,  12.,  12.,  12.,  12.,  12.], dtype=float32)
# On SLC: array([ 15.,  12.,  16.,  15.,  17.,  13.,  15.,  15.,  15.], dtype=float32)
```

So this may not be a PyTables bug perse, perhaps the fault lies with Numexpr. However, I encountered it while using PyTables. It would be good for PyTables to require a properly working version of Numexpr. Also, what is the source of this bug? Numexpr, Numpy, or something else?

We encountered a similar bug/inconsistency recently concerning the accuracy of the query when using `log10`: https://github.com/HiSPARC/sapphire/issues/114

(Our issue on this problem https://github.com/HiSPARC/sapphire/issues/116)

153957:
It appears that Numexpr 2.4.6 is the culprit. We can downgrade to 2.4.4, unfortunately upgrading to 2.5.0 is not an option because that also breaks some parts of the code: https://github.com/HiSPARC/publicdb/issues/131

FrancescAlted:
The problem seems to be in the [conda package](https://github.com/pydata/numexpr/issues/210) .  Closing this.

tomkooij:
This needs reopening.

When using Intel MKL with numexpr (default in conda from numexpr=2.4.6) all ```table.read_where()``` queries are broken.

Numexpr uses VML functions which need contiguous memory, but tables are ```numpy.rec.array()```s which are non-contiguous.

This can be fixed by getting and passing the right arguments to ```NumExpr.run()```.  I have a fix, will make PR.

PR #536, opened 2016-03-31
fix read_where() query with functions when numexpr uses Intel MKL

tomkooij:
When using numexpr with the Intel MKL (default for conda packages for numexpr>2.4.6) ALL ```table.read_where()``` queries which use functions are broken.

Issue: #534 
See also: pydata/numexpr#210 

numexpr (with MKL support) uses VML functions which require contiguous memory. tables are non-continguous so a copy operation is needed. Therefore the ex_uses_vml kwargs needs to be passed by ```conditions.call_on_recarr()```.

This fixes my testcase.

~~This breaks many unittests~~ 

ToDo:
- [x] Figure out if ```numexpr.GetContext``` needs to be called before ```numexpr.GetExprNames```
- [x] fix L428 of conditions.py
- [x] Fix other call_on_recarray() call
- [x] Investigate and fix failing unittest: test_queries.ScalarTableUsageTestCase.test_unsupported_op
- [x] Add unittest
- [x] flake8 / PEP8

### DIFF
--- a/tables/conditions.py
+++ b/tables/conditions.py
@@ -34,7 +34,7 @@ from __future__ import absolute_import
 import re
 from numexpr.necompiler import typecode_to_kind
 from numexpr.necompiler import expressionToAST, typeCompileAst
-from numexpr.necompiler import stringToExpression, NumExpr
+from numexpr.necompiler import stringToExpression, NumExpr, getExprNames
 from numexpr.expressions import ExpressionNode
 
 from .utilsextension import get_nested_field
@@ -333,7 +333,7 @@ class CompiledCondition(object):
                 idxvars.append(idxvar)
         return frozenset(idxvars)
 
-    def __init__(self, func, params, idxexprs, strexpr):
+    def __init__(self, func, params, idxexprs, strexpr, **kwargs):
         self.function = func
         """The compiled function object corresponding to this condition."""
         self.parameters = params
@@ -342,6 +342,8 @@ class CompiledCondition(object):
         """A list of expressions in the form ``(var, (ops), (limits))``."""
         self.string_expression = strexpr
         """The indexable expression in string format."""
+        self.kwargs = kwargs
+        """NumExpr kwargs (used to pass ex_uses_vml to numexpr)"""
 
     def __repr__(self):
         return ("idxexprs: %s\nstrexpr: %s\nidxvars: %s"
@@ -370,7 +372,8 @@ class CompiledCondition(object):
             exprs2.append((var, ops, tuple(limit_values)))
         # Create a new container for the converted values
         newcc = CompiledCondition(
-            self.function, self.parameters, exprs2, self.string_expression)
+            self.function, self.parameters, exprs2, self.string_expression,
+            **self.kwargs)
         return newcc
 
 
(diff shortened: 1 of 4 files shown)


tomkooij:
This is my testcase:
```python
import tables
import numpy as np
import numexpr

FILENAME = 'test.h5'

class Test(tables.IsDescription):
    a    = tables.Float32Col(dflt=1, pos = 2)  # single-precision
    b    = tables.Float32Col(dflt=1, pos = 2)  # single-precision
    c    = tables.Float32Col(dflt=1, pos = 2)  # single-precision

def make_table():
    with tables.open_file(FILENAME, "w") as data:
        table = data.create_table(data.root, "table", Test)

        row = table.row
        for i in range(100):
            row['a'] = i
            row['b'] = 200+i
            row['c'] = i
            row.append()
        table.flush()

if __name__ == '__main__':
    print "tables: ", tables.__version__
    print "numexpr: ", numexpr.__version__
    print "MKL/VML: ", numexpr.get_vml_version()

    make_table()

    with tables.open_file(FILENAME, "r") as data:
        result = data.root.table.read_where('log10(b) < 2')
        if len(result) == 0:
            print "Succes!"
        else:
            print "FAIL! log10(b) should be >> 2. result must be empty!", len(result['b'])

```
log10(200) is 2.3 so the query should always be empty!

result:
```
tables:  3.2.2.dev0
numexpr:  2.5.2.dev0
MKL/VML:  Intel(R) Math Kernel Library Version 11.3.1 Product Build 20151021 for Intel(R) 64 architecture applications
FAIL! log10(b) should be >> 2. result must be empty! 66
```

tomkooij:
I squashed the fix into a single commit, 

I added unittests for queries using functions to ```test_queries.py```. This tests many common functions using float32 and float64. Many of these fail using the current default ```conda``` packages for pytables and numexpr.

Testing complex functions is not implemented.

FrancescAlted:
Great work; this is very useful indeed.  Please keep us informed on the failures in test_indexvalues.  Thanks!!

tomkooij:
Thanks.
I reran the --heavy test suite on the current ```develop```: b5fa156
The same test in test_indexvalues seem to fail, these failures are unrelated to this PR. 

The test_indexvalues tests all seem to fail on the same assertion:
```
FAIL: None (tables.tests.test_indexvalues.FIHSV10bTestCase)
  File "tables/tests/test_indexvalues.py", line 2795, in test13g
    table1.where(condition, start=0, stop=10, step=1)]
  File "tables/table.py", line 1483, in where
    return self._where(condition, condvars, start, stop, step)
  File "tables/table.py", line 1522, in _where
    return row._iter(start, stop, step, chunkmap=chunkmap)
  File "tables/tableextension.pyx", line 783, in tables.tableextension.Row._iter (tables/tableextension.c:9486)
    self._init_loop(start, stop, step, coords, chunkmap)
  File "tables/tableextension.pyx", line 865, in tables.tableextension.Row._init_loop (tables/tableextension.c:10555)
    assert self.nrowsinbuf % self.chunksize == 0
AssertionError
```

tomkooij:
The test failures are unrelated to this PR, but they always seem to fail: I opened issue #538 

This seems done!

FrancescAlted:
:+1:  Merging.

PR #540, opened 2016-04-08
Update to use HDF5 1.8.16 (naming and minor API changes)

tadeu:
I'm updating it to work with the CMake built HDF5 1.8.16 in Windows.
I've also picked a commit from c-blosc that fixes it for HDF5 1.8.16.

### DIFF
--- a/RELEASE_NOTES.txt
+++ b/RELEASE_NOTES.txt
@@ -17,6 +17,8 @@ Improvements
 - It is now possible to  read reference types by dereferencing them as
   numpy array of objects (closes :issue:`518` and :issue:`519`).
   Thanks to Ehsan Azar
+- It is now also possible to use HDF5 with the new shared library naming
+  scheme (>= 1.8.10, hdf5.dll instead of hdf5dll.dll) on Windows (PR #540).
 
 
 Bugs fixed
--- a/c-blosc/hdf5/blosc_filter.c
+++ b/c-blosc/hdf5/blosc_filter.c
@@ -20,7 +20,15 @@
 
 #if H5Epush_vers == 2
 /* 1.8.x */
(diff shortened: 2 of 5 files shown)


avalentino:
Thank you @tadeu.
My only concern is that i would like to keep the compatibility with old HDF5 layout.
Ideas?

FrancescAlted:
The PR by @tadeu is just backporting something in blosc-hdf5 repository, so it should be really safe to merge it, unless you @avalentino has some specific concerns.

avalentino:
@FrancescAlted I was just worried about the library filename change hdf5dll.dll --> hdf5.dll
Honestly I didn't checked which is the last "official" binary distribution that uses the old naming.

FrancescAlted:
Good point @avalentino.  @tadeu, please could you make some experiments with earlier HDF5 versions and tell us which versions your modification will compatible with?  If you can provide some kind of backward compatibility with older version, that would be really great.  Thanks!

avalentino:
The old naming schema has been used up to HDF5 1.8.10 (November 2014) in "official" binary packages.
For PyTables the minimum supported version is HDF5 >= 1.8.4 (>=1.8.7 is strongly recommended).

If the patch for keeping backward compatibility is too complex we could just drop support for HDF5 < 1.8.11 (May 2013) on Windows.

It should not be a big issue IMO.

tadeu:
Well noted @avalentino, I'll try to make it backwards compatible.

tadeu:
I've modified the commit to add backward compatibility and tested it with both HDF 1.8.8 and 1.8.16

FrancescAlted:
Looks good to me.  @tadeu could you please update the RELEASE_NOTES accordingly please?  After that I will merge.  Thanks!

tadeu:
Done

PR #542, opened 2016-04-18
Fix docstring for create_table keyword obj

153957:
It is not true that native Python sequences are accepted.
Only numpy.ndarray arrays with named dtypes are allowed.

See the if statement after the docstring:

```python
if not isinstance(obj, numpy.ndarray):
    raise TypeError('invalid obj parameter %r' % obj)
```


FrancescAlted:
Good point.  Will merge.

PR #544, opened 2016-04-28
fix bench-pytables.sh 

rohitjamuar:
pytables_backend had hard-coded paths that are used as workspace by benchmark executer. The runtime makes an assumption about the existence of those hard-coded paths. These changes should ensure that those paths exist before benchmarks are executed.

@FrancescAlted

### DIFF
--- a/bench/pytables_backend.py
+++ b/bench/pytables_backend.py
@@ -19,7 +19,16 @@ class PyTables_DB(DB):
         self.filename += '-' + kind
         if docompress:
             self.filename += '-' + complib + str(docompress)
-        self.filename = datadir + '/' + self.filename + '.h5'
+        self.datadir = datadir
+        if not os.path.isdir(self.datadir):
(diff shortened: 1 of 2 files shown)


bench/pytables_backend.py
```diff
@@ -65,9 +74,12 @@ def fill_table(self, con):
 
     def index_col(self, con, column, kind, optlevel, verbose):
         col = getattr(con.root.table.cols, column)
+        tmp_dir = os.path.join(self.datadir, "scratch2", "faltet")
```
FrancescAlted: Provided that we are using a 'scratch2' in the datadir directory, it would be better to remote the trailing 'faltet'.

rohitjamuar:
@FrancescAlted, if this PR doesn't require any more alterations, could it be merged and closed?

FrancescAlted:
Thanks!

PR #550, opened 2016-06-17
Fix VLArray to host more than 32-bit long rows

FrancescAlted:
This is still preliminary because it segfaults when the number of elements is > 2**32.  Any volunteer to look into this?

### DIFF
--- a/src/H5VLARRAY.c
+++ b/src/H5VLARRAY.c
@@ -197,7 +197,7 @@ out:
 
 herr_t H5VLARRAYappend_records( hid_t dataset_id,
                                 hid_t type_id,
-                                int nobjects,
+                                size_t nobjects,
                                 hsize_t nrecords,
                                 const void *data )
 {
@@ -275,7 +275,7 @@ out:
 herr_t H5VLARRAYmodify_records( hid_t dataset_id,
                                 hid_t type_id,
                                 hsize_t nrow,
-                                int nobjects,
+                                size_t nobjects,
                                 const void *data )
 {
 
--- a/src/H5VLARRAY.h
+++ b/src/H5VLARRAY.h
@@ -23,14 +23,14 @@ herr_t H5VLARRAYmake( hid_t loc_id,
 
 herr_t H5VLARRAYappend_records( hid_t dataset_id,
                                 hid_t type_id,
-                                int nobjects,
+                                size_t nobjects,
                                 hsize_t nrecords,
                                 const void *data );
 
 herr_t H5VLARRAYmodify_records( hid_t dataset_id,
                                 hid_t type_id,
                                 hsize_t nrow,
-                                int nobjects,
+                                size_t nobjects,
                                 const void *data );
 
 herr_t H5VLARRAYget_info( hid_t   dataset_id,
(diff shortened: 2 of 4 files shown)


FrancescAlted:
You can reproduce the segfault by running [this test](https://github.com/PyTables/PyTables/pull/550/files#diff-45ebfb69c67a846065c3e0543f398863R261).  Please note that you will need a machine with at least 16 GB of RAM to comfortably run this test.

tomkooij:
I had a go with this.

Appending 2**32+1 rows to the vlarray is no problem. Appending it several times is no problem. 
reading the big vlarray with `vlarray[0]` causes the segfault.

I created a big vlarray (as in `test05` in this PR) and closed the file. 
Restarted IPython and opened the file in read-only mode.
`vlarray[0]` segfault.s

I narrowed it down to the call to `H5Dread()` in `hdf5extenstions.pyx:_read_array()` [line 2071](https://github.com/PyTables/PyTables/blob/develop/tables/hdf5extension.pyx#L2071)
The calls before `H5Dread` seem okay, at least as far as I can debug.

I tested on `HDF5-1.8.16`.

andreabedini:
For future reference the line is https://github.com/PyTables/PyTables/blob/ecc32b1fd30e4bd13077fee4664bd2a627b4df23/tables/hdf5extension.pyx#L2079 (by default GitHub links to line 2071 in the current version, which is subject to change, press `y` to reload the page with the canonical address).

andreabedini:
After another look, I can add a piece of the puzzle. `h5dump` segfaults on the temporary file created by

```
import numpy
import tables
h5 = tables.open_file('/tmp/tmp.h5', 'w')
N = int(2**32 + 1)  # > 2**32
vlarray = h5.create_vlarray(h5.root, 'vlarray2', atom=tables.Int8Atom(), filters=None, expectedrows=N)
x = numpy.zeros(N, dtype="i1")
vlarray.append(x)
h5.close()
```

so the problem might not be in `hdf5extenstions.pyx:_read_array` but elsewhere, since we are somehow creating a corrupted file.

tomkooij:
I think (guess) `h5dump` segfaults in the HDF5 library identical to the `_read_array` fails in the HDF5 lib (I traced that last one into the library).

That could ofcourse be caused by a corrupted file we create, but I could just as well be a library problem. It's probably an 32bit overflow somewhere, but where?

FrancescAlted:
Yeah, it would be nice if we could create a minimal C program creating the dataset, test that `h5dump` crashes on it and sending the report back to the HDF Group.  But if the problem is on the HDF5 side, then I'd merge this PR as-is.

ClimbsRocks:
Sounds like this might fix an [issue](https://github.com/pandas-dev/pandas/issues/2773) I've been running into for a while! 

In my case, the issue is caused by having too many string columns, which at some point hit some kind of a cumulative limit. 

Any updates on this PR?

louis925:
Can someone merge this pull request? This problem makes pandas not able to save some type of dataframe into hdf5 file. Thanks!

FrancescAlted:
I revisited this, and I can still reproduce the segfaults described here, even with HDF5 1.10.4, so I am not merging this until more investigation would be done.

PR #493, opened 2015-08-31
Single code base for Python 2 and 3

andreabedini:
Work in progress. This has to be thoroughly checked. From the top of my head surely the `range/xrange` dichotomy will cause problems.

PR #553, opened 2016-06-20
Python 3: Allow binary strings in VLStringAtom() 

### LINKED ISSUE #552: Python 3: VLStringAtom() reads a bytestring but write not allowed
When reading strings from VLArray, VLStringAtom() in python 3, a `read()` returns a `b` string (bytestring) but appending that same string throws a `TypeError` exception. (See testcase below).

This seems easily fixed by allowing bytestrings in VLStringAtom(). I will push a PR with a fix.

```python
"""
Write a string to VLArray of VLStringAtom.
Read it and append it again

Succes on python 2.7
Fails on python 3.5: TypeError: object is not a string

"""
import tables

with tables.open_file('vlarray.h5', 'w') as data:
    vlarray = data.create_vlarray(data.root, 'vlarray', tables.VLStringAtom(),
                                  'array of binary strings')
    string = 'Hello, World!'
    vlarray.append(string)

    read_string = vlarray.read()[0]
    print(read_string, type(read_string))  # b'Hello, World!' <class 'bytes'>
    vlarray.append(read_string)  # fails on python3

    print('reading again...')
    read_string = vlarray.read()[0]
    print(read_string, type(read_string))

```

Probably related: #499

tomkooij:
Fixes #552

Add  six.binary_types to allowed strings in VLStringAtom()

six.binary_types is 'str' in Python 2 and 'bytes' in Python 3.

### DIFF
--- a/tables/atom.py
+++ b/tables/atom.py
@@ -1119,7 +1119,8 @@ class VLStringAtom(_BufferedAtom):
     base = UInt8Atom()
 
     def _tobuffer(self, object_):
-        if not isinstance(object_, six.string_types):
+        if not isinstance(object_, six.string_types) and \
+                not isinstance(object_, six.binary_type):
             raise TypeError("object is not a string: %r" % (object_,))
         return numpy.string_(object_)
 


andreabedini:
Hi @tomkooij, thank you very much with this. I think this is part of a bigger problem: `VLStringAtom` doesn't work with non-ascii representable strings at all!

What I would like to see is a `VLStringAtom` that honours its name and can work with all strings. It would have a field to determine the encoding (defaulting to `sys.getdefaultencoding()`) which will have to be saved to disk. At that point `_tobuffer(object)` can be replaced with `object.encode(self.encoding)`.

Last time I tried to do this I ended up lost in refactoring hell and I gave up. Wanna give it a shot?

Alternatively, we should just make that method only accepts `bytes` or ascii-representable strings. This is contrary to its purpose but faithful to what it does at the moment.

tomkooij:
@andreabedini 

I took some time to try and understand string in python 2 and 3. I have formulated requirements for 'VLStringAtom' below. Let me know if this is acceptable.

My understanding of the string mess:

-  py2 `str` == py3 `bytes`: just a sequence of bytes. Basically useless data unless the encoding (serialisation) is known. 
- py2 `unicode` == py3 `str`: decoded strings. Internal encoding depends on python version and CPython (messy).

``` python
u = u'Hbar = \u210f'    # the u is optional in python 3
s = s.encode('utf-8')    # create bytestring with utf-8 encoding
type(u), type(s)
```
returns `(unicode, str)` on python2 and `(str, bytes)` on python 3.

`VLStringAtom` stores bytestring  (py2 `str` and py3 `bytes`) without storing encoding.
`VLUnicodeAtom` stores unicode (py2 `unicode` and py3 `str`) as UCS-4

From your desciption, I conclude you want to change `VLStringAtom` from storing only bytestrings to also accepting actual strings. Strings are encoded and stored in HDF5 as bytestring, but `read` will return `unicode` on py2 and `str` on py3. 

Requirements for `VLStringAtom`:
- stores encoding in a node attribute. For bytestrings encoding is None. When the attribute is missing (previous version) a bytestring is assumed.
- encodes `unicode` (py3 `str`) upon `append` and stores an encoded bytestring. Upon `read` the bytestring is decoded and return type is `unicode` (py2) or `str` (py3)
- stores bytestrings (py2 `str`, py3 `bytes`) with encoding 'None'. Upon `read` the returned type is `str` or `bytes` 

I'll be happy to provide a PR, if this seems to correct behaviour.

tomkooij:
However, the KISS solution is easiest to implement: 

> Alternatively, we should just make that method only accepts bytes or ascii-representable strings. This is contrary to its purpose but faithful to what it does at the moment.

We can *disallow* strings (py2 `unicode` and py3 `str`) and only allow bytestrings in `VLStringAtom()` and let the user keep track of and provide en/decoding.

FrancescAlted:
Yes, I agree that we should follow the KISS principle here.  @andreabedini do you see a problem in keeping the explicit separation between  `VLStringAtom` and `VLUnicodeAtom`?  I mostly agree with @tomkooij in that we could disallow py2 unicode and py3 str and only allow bytestrings for `VLStringAtom()`.

On the other hand, we could also disallow py2 strings and py3 bytestrings in `VLUnicodeAtom()`.  As the later stores things in UCS4 by default, we don't need to store the encoding anywhere (it is explicit).  However, as UCS4 takes 4 bytes per character, and VL types cannot be compressed in HDF5, I still see the merit in adding an (optional) encoding field for  `VLUnicodeAtom()`.  Would that make sense?

andreabedini:
> Yes, I agree that we should follow the KISS principle here.

Always happy to KISS!

> @andreabedini do you see a problem in keeping the explicit separation between VLStringAtom and VLUnicodeAtom? I mostly agree with @tomkooij in that we could disallow py2 unicode and py3 str and only allow bytestrings for VLStringAtom().

No problem at all, I think it's all just a question of naming, strings are not sequences of bytes (at least not without a specified encoding), so it is surprising having VLStringAtom store or retrieve bytes. That's why we have Python3! :)

> On the other hand, we could also disallow py2 strings and py3 bytestrings in VLUnicodeAtom(). 

I don't think there's much to do on Python2 since bytes and strings are coalesced into a same type. 

```
>>> isinstance("ciao", bytes)
True
```

> As the later stores things in UCS4 by default, we don't need to store the encoding anywhere (it is explicit). However, as UCS4 takes 4 bytes per character, and VL types cannot be compressed in HDF5, I still see the merit in adding an (optional) encoding field for VLUnicodeAtom(). Would that make sense?

(AFAIK UCS-4 is the same thing as UTF-32). It's before ☕ here but I would agree.

Perhaps VLStringAtom and VLUnicodeAtom should be renamed VLBytesAtom and VLStringAtom at least on PY3?

FrancescAlted:
@andreabedini Renaming the atoms for py3 is an option, but IMO that would create some API mess; and certainly this is not KISS, so I'm -1 on this.  In order to keep moving, @tomkooij would you mind to send a KISS PR so that we can discuss on it?

FrancescAlted:
@tomkooij Or just improve this one, which should be simpler indeed.

tomkooij:
Will do!

andreabedini:
> Renaming the atoms for py3 is an option, but IMO that would create some API mess; and certainly this is not KISS, so I'm -1 on this.

I agree it shouldn't be done light-heartedly, but eventually something like this should happen. Uses expect VLStringAtom to be able to contain a list of names (like René Magritte, Antonín Dvořák, Søren Kierkegård, Γιώργος Νταλάρας, 심은하, etc).

andreabedini:
Also, please note that this is PR to develop and there is no plan to make a release out of this branch just yet.

tomkooij:
A KISS proposal:
- Allow only bytestrings in `VLStringAtom`. Doesn't change much on py2, but fixes python 3.
- Update docstring to reflect changed behaviour. Give user responsability of en/decoding.
- Encoding is not stored. (Up to the user).

(Let me know if you'd like a single squased commit)

This does not seem to break old code, but does fix my problems. I have not written tests yet, but if this is acceptable, i'll add some tests to prevent regressions.

@andreabedini : Is a release planned at all? In the current `3.2.2` there is a nasty VML functions/numexpr API bug (fixed in 1da2d21) and the bytestrings bug. It would be really helpful if this can be fixed in a release (so we can use `conda` packages)

FrancescAlted:
This looks acceptable to me.  Could you add tests for this?  What about preventing bytestrings to be stored in VLUnicodeAtom()?  Would make sense to raise an Error (or maybe just a Warning) suggesting that VLStringAtom() should be used instead?

Also, Tom you suggested some time ago that a new release should be done.  I think the best way to ensure a prompt release is doing it for yourself.  Tell use in case you want to step ahead as a release manager and the rest of the PyTables crew will try to help as much as we can.

FrancescAlted:
Also note that this PR currently has some issues in for TravisCI.

andreabedini:
@tomkooij @FrancescAlted I'd be happy to facilitate or help working toward a release. I wasn't planning a release in the short term because develop has diverged significantly from master. E.g. with it contains #493 which I am not 100% whether it changes the API or not. There was an issue with `range`/`lrange`/`xrange` that is mentioned in the PR and, looking at it now, I just noticed a possible problem with `.next()` vs `__next__()`. Given there's a big refactoring coming up (the merge with h5py), I was thinking of leaving all the breakage to a future version 4.0.

The pragmatic approach would be to cherry-pick what you want from the develop branch, apply those changes to master, and later re-merge the master into develop (to make git happy about the double-commits). I think I already did this once, so I can offer advice.

That said, I have no intention to stop anyone from working on a release, rather the opposite. Happy to support Tom if he wishes to make a release.

tomkooij:
I'd like to prepare a release. However, this PR also breaks the API (although in a minor way, because it does not accept `str` in `VLStringAtom`.  

There are two options:
- Pragmatic: cherry pick the bugfixes and change this PR into a bugfix only, not changing the API. Leave #493 out.
- Look into the problems @andreabedini mentioned about #493 and release if possible.

I'll look into it.

andreabedini:
> I'd like to prepare a release. However, this PR also breaks the API (although in a minor way), because it does not accept str in VLStringAtom.

Yeah, I thought that too. I would avoid breaking the API until we have everything figured out, but this is just my opinion. Thanks for the effort @tomkooij, you're a hero!

avalentino:
Hi, I totally agree that it is a good idea to keep VLStringAtom only for byte strings.
An option could be to deprecate the current behaviour and print a warning in the next bugfix release (no API change) and  then remove totally the VLStringAtom support for str in the next feature release.

FrancescAlted:
+1 for the deprecation warning that suggeste @avalentino .  @tomkooij could you implement something like that?

tomkooij:
No problem. I'll keep the API as is and add a depreciation warning. 

I'll cherry pick as bugfixes from `develop` into a release branch and make sure the heavy tests pass first.

scopatz:
+1 All in favor of a new release!

Issue #545, opened 2016-05-03
HDF5 1.10 breaks pytables 3.2.2


dsteinberg:
On Arch Linux HDF5 was just upgraded to 1.10. This seems to break array creation. E.g.

```
>>> f2 = tables.open_file('test.hdf5', 'w')
>>> f2.create_array('/', 'test', obj=np.random.randn(100))
```

Gives the following error,

```
---------------------------------------------------------------------------
HDF5ExtError                              Traceback (most recent call last)
<ipython-input-4-509467cf97dc> in <module>()
----> 1 f2.create_array('/', 'test', obj=np.random.randn(100))

/home/dsteinberg/.virtualenvs/uncover-ml/lib/python3.5/site-packages/tables/file.py in create_array(self, where, name, obj, title, byteorder, createparents, atom, shape)
   1150         parentnode = self._get_or_create_path(where, createparents)
   1151         return Array(parentnode, name,
-> 1152                      obj=obj, title=title, byteorder=byteorder)
   1153 
   1154     createArray = previous_api(create_array)

/home/dsteinberg/.virtualenvs/uncover-ml/lib/python3.5/site-packages/tables/array.py in __init__(self, parentnode, name, obj, title, byteorder, _log, _atom)
    185         # Ordinary arrays have no filters: leaf is created with default ones.
    186         super(Array, self).__init__(parentnode, name, new, Filters(),
--> 187                                     byteorder, _log)
    188 
    189     def _g_create(self):

/home/dsteinberg/.virtualenvs/uncover-ml/lib/python3.5/site-packages/tables/leaf.py in __init__(self, parentnode, name, new, filters, byteorder, _log)
    260         # is a lazy property that automatically handles their loading.
    261 
--> 262         super(Leaf, self).__init__(parentnode, name, _log)
    263 
    264     def __len__(self):

/home/dsteinberg/.virtualenvs/uncover-ml/lib/python3.5/site-packages/tables/node.py in __init__(self, parentnode, name, _log)
    265             #   Create or open the node and get its object ID.
    266             if new:
--> 267                 self._v_objectid = self._g_create()
    268             else:
    269                 self._v_objectid = self._g_open()

/home/dsteinberg/.virtualenvs/uncover-ml/lib/python3.5/site-packages/tables/array.py in _g_create(self)
    217             # on
    218             (self._v_objectid, self.shape, self.atom) = self._create_array(
--> 219                 nparr, self._v_new_title, self.atom)
    220         except:  # XXX
[...]

PR #558, opened 2016-06-27
Fix vlstringatom on python3 after lib2to3

tomkooij:
This is a version of #553 against `master`.

This fixes: #552 

The bug is caused by `lib2to3` changing `basestring` to `str` on python3.

This fix also explicitly allows strings to make sure we do not break the API. 
This allows `bytes`  (which includes `str` in python 2) and `unicode` for python 2. 
This is converted to `bytes` and `str` for python 3 by `lib2to3`.

ToDo:
- [x] check if this breaks tests

### DIFF
--- a/tables/atom.py
+++ b/tables/atom.py
@@ -1167,7 +1167,7 @@ class VLUnicodeAtom(_BufferedAtom):
         elif isinstance(object_, bytes):
             warnings.warn("Storing bytestrings in VLUnicodeAtom is "
                           "deprecated.", DeprecationWarning)
-            return numpy.string_(object_)
+            return numpy.unicode_(object_)
         else:
             raise TypeError("object is not a string: %r" % (object_,))
 
--- a/tables/atom.py
+++ b/tables/atom.py
@@ -1143,7 +1143,7 @@ class VLUnicodeAtom(_BufferedAtom):
         # NumPy ticket #525).  Since ``_tobuffer()`` can't return an
         # array, we must override ``toarray()`` itself.
         def toarray(self, object_):
-            if not isinstance(object_, bytes):
+            if not isinstance(object_, basestring):
                 raise TypeError("object is not a string: %r" % (object_,))
             ustr = unicode(object_)
             uarr = numpy.array(ustr, dtype='U')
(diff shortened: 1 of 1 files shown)


tomkooij:
I guess issue #545 needs to be handled as well in this release.

tables/atom.py
```diff
@@ -1092,7 +1097,8 @@ class VLStringAtom(_BufferedAtom):
     base = UInt8Atom()
 
     def _tobuffer(self, object_):
-        if not isinstance(object_, basestring):
+        if not isinstance(object_, bytes) and \
+                not isinstance(object_, unicode):
```
FrancescAlted: I thought we wanted to issue a `DeprecationWarning` when someone is trying to feed a `VLString` with a `unicode` string.  I am missing sometthing?

tomkooij:
I was going to implement that in `develop` but It shouldn't be to difficult here. I'll add a commit.

tomkooij:
@FrancescAlted 5b97ee0 implements the `DeprecationWarning`.

FrancescAlted:
Good.  What about adding the complementary `DeprecationWarning` when trying to assign a bytestring into a `VLUnicodeAtom`?

tomkooij:
Fair enough... However this does change the API: bytestrings were not allowed in `VLUnicodeAtom` on Python 3 before 8a4337c now it throws a `DeprecationWarning` on both py2 and py3. 
I only tested on python 3, but Travis will check python 2. (Although there are no tests for bytestrings in VLUnicodeAtom)

FrancescAlted:
In my opinion, a DeprecationWarning is not an API change, but more precisely a warning about a future API change, so I think we are not doing something terribly bad here.  Also, I think preventing storing bytestrings to be stored in VLUnicode datasets is a good thing for the user in that it prevents wasting 4x disk space.  In another major release of PyTables (4.0?) we can rethink this string handling, as Andrea suggests.

tomkooij:
I changed the logic of throwing a DeprecationWarning to be DRY. (Which would have prevented the bug in 8a4337c)

I can squash all commits into a single commit, but I do not know PyTables policy on this.

FrancescAlted:
AFAIK there is not such a policy for the number of commits in a PR, so I think we are ready for merging.

PR #561, opened 2016-07-01
Python 3: Read objects form node attributes pickled on py2

### LINKED ISSUE #560: Incompatibility between pickled node attributes py2<>py3
Issue #368 contains a lot of useful background info. Most of the problems reported there were fixed (upstream), but a related issue still remains:

When porting code from py2 to 3 that stores classes in node attributes, I ran into unpickling errors, caused by incompatibility between python 2 and 3 pickles. (Although they are suppossed to be compatible).

Most of these problems were fixed in numpy-1.9 (see also #368) but some classes still fail to be unpickled correctly on python 3. It is easiest demonstrated with `datetime.datetime` objects (see code below).

The problem is non-ascii characters in the pickled string. Pickle throws an `UnicodeDecodeError` on these malformed pickled strings in python 3, so it should be possible to fix. I have a fix that I need to test.

```python
"""
write datetime object to node attribute on py2
read datetime from node attribute on py3
"""
import sys
import tables
from datetime import datetime

FILENAME = 'py2pickled.h5'

if sys.version_info[0] == 2:
    with tables.open_file(FILENAME, 'w') as data:
        data.set_node_attr('/', 'datetime', datetime.now())

if sys.version_info[0] > 2:
    with tables.open_file(FILENAME, 'r') as data:
        print('datetime = ', data.get_node_attr('/', 'datetime'))
```

tomkooij:
This fixes #560.

Objects pickled on python 2 with non-ascii characters in the pickled string can be unpickled on python>=3.4.

This requires python>=3.4. It throws an `LookupError` on python 3.3. On python 3.3 'latin-1' encoding (or any encoding that is limited to bytes, not utf-8) might work, but that hack is more dangerous. IMHO it's not worth supporting this on python 3.3. 

The included testcase fails on the current `master`.

`cPickle.loads(value, 'encoding=bytes')` throws a `TypeError` on python 2, so that should be safe.


tomkooij:
Travis fails on the newly added testcase in python 3.2 and 3.3. Good!
I'll skip the testcases on python <3.4. But leave them in python 2.

FrancescAlted:
This looks good to me.  Reading your comments I wonder whether it is worth to officially deprecate the support for Python 3.3.  Comments?

tomkooij:
@FrancescAlted I don't think it is worth deprecating support for python 3.3 because of this. Many classes are still pickled/unpickled correctly from python 2 to python 3 when stored in node attributes. This affects only some special classes *and* only affects usage between python 2 to 3.

Let's not base python 3.3 support on this. However, numpy drops python 2.6 and 3.3. in 1.12: [release notes](https://github.com/numpy/numpy/blob/master/doc/release/1.12.0-notes.rst). That should be a good reason!

FrancescAlted:
Maybe. My point is that Python 3.3 user base should be really tiny and also that many projects are starting to drop support for it (numpy is just one of them).  At any rate, we can always drop Python 3.3 support in next minor version (PyTables 3.3, which is is a bit ironic ;)

Do you want me to merge this or you want to do more things here?

tomkooij:
This is ready to merge!

FrancescAlted:
@tomkooij I just invited you to the Core team in the PyTables GitHub project so that you can perform the steps for releasing the package.  Also, I recommend you to subscribe to the https://groups.google.com/forum/#!forum/pytables-dev mailing list.

PR #563, opened 2016-07-13
Try 'latin1' encoding before 'bytes'

tomkooij:
This is an update to the fix for #560 

Classes get mangled by unpickling with encoding='bytes' because
attributes are unplicked as bytestrings. Try 'latin1' enconding
first, to increase the chance of succesful unpickling python2
pickles on python3.

It's not easy to construct a testcase that fails before e36dbd4 is applied. Classes are succesfuly unplicked by `encoding=bytes` but are mangled. They look like a proper `<class ... >` object, but are unusable. An assertion based on `isinstance` does not distinguish between the two. 
The test needs to assert that a certain attribute exists after unpickling.

ToDo:
- [x] extend testcase to check 'latin1' unpickle.


tables/attributeset.py
```diff
@@ -330,8 +330,12 @@ def __getattr__(self, name):
             except UnicodeDecodeError:
                 # Object maybe pickled on python 2 and unpickled on python 3.
                 # encoding='bytes' was added in python 3.4 to resolve this.
+                # However 'bytes' mangles class attributes as they are
+                # unplicked as bytestrings. Hence try 'latin1' first.
                 # Ref: http://bugs.python.org/issue6784
                 try:
+                    retval = cPickle.loads(value, encoding='latin1')
+                except TypeError:
                     retval = cPickle.loads(value, encoding='bytes')
```
avalentino: Do we need another nested try/except?

tomkooij: @avalentino: Yes, indeed! This is starting to get ridiculous, but an extra try/except is needed. 502c40c

andreabedini:
Much dislike that we have to go around this but solution look good to me.

tomkooij:
The testcase fails for the current master. (good!). This seems to fix issue #560 (hopefully..., but at least for now).

This is a PR to master. I'll merge this into master and develop, when I have time to do it properly.

Issue #564, opened 2016-07-30
Query error on indexed columns


ajenkins-cargometrics:
I've run into a situation where a query returns incorrect results when it involves columns with indexes.  It seems to be due to a bug in the `_get_idx_expr` function.  Specifically the `strexpr` that it returns can be wrong.  Here is a program which reproduces the problem with tables 3.2.3.1.

```python
from __future__ import print_function
import tables

handle = tables.open_file('/tmp/query_bug.h5', 'w')

schema = {
    'symbol': tables.StringCol(itemsize=60, pos=0),
    'open': tables.Float64Col(pos=1)
}

table = handle.create_table('/', 'symbols', schema, filters=tables.Filters(complevel=1))

# Commenting out the lines creating indexes makes the queries work correctly
table.cols.symbol.create_csindex()
table.cols.open.create_csindex()

num_rows = 2000

# Create a bunch of rows, with 2 duplicates of each row
table.append([('sym_{}'.format(j), i) for i in xrange(num_rows / 10) for j in (range(5) * 2)])
table.flush()

query1 = "(open == 50.0) & (symbol == 'sym_0')"
query2 = "(open == 100.0) & (symbol == 'sym_0')"

# These two queries return correct results
print(table.get_where_list(query1)) # returns [500, 505]
print(table.get_where_list(query2)) # returns [1000, 1005]

# query3 just combines the previous two queries with 'or', and should select the union of the results.
query3 = "({}) | ({})".format(query1, query2)

# Should return [500, 505, 1000, 1005], but instead returns []
print(table.get_where_list(query3))


handle.close()
```

[tables_query_bug.zip](https://github.com/PyTables/PyTables/files/391623/tables_query_bug.zip)

Stepping through the code in the debugger for the query with `query3`, it seems the problem is that the `_get_idx_expr` function incorrectly compiles `query3`.   The value of `query3` is

```python
"((open == 50.0) & (symbol == 'sym_0')) | ((open == 100.0) & (symbol == 'sym_0'))" 
```

In the compiled version, the `string_expression` part is `'(((e0 & e1) & e2) & e3)'`.  When this compiled expression is used in `_table__where_indexed` to create `chunkmap`, it does the wrong thing.  I tried changing the value of `string_expression` in the debugger to `'((e0 & e1) | (e2 & e3))'`, which is what I think it should have been, and then the query produces the correct results.  

[...]

PR #565, opened 2016-08-05
Use deepcopy for recursion to work

tomkooij:
This is a fix for #564

This function returns a tuple with references to `strexpr` and
`idxexprs` which a then passed recursively to this function while
traversing an expression AST. We need to make deepcopies of these
objects to prevent overwriting them. This fixes queries such as:
((a = 1 & b = 2) | (a = 2 & b = 1)).
Fixes gh-564. Thanks to ajenkens-cargometrics for reporting.

### DIFF
--- a/tables/conditions.py
+++ b/tables/conditions.py
@@ -31,6 +31,7 @@ Functions:
 """
 
 import re
+from copy import deepcopy
 from numexpr.necompiler import typecode_to_kind
 from numexpr.necompiler import expressionToAST, typeCompileAst
 from numexpr.necompiler import stringToExpression, NumExpr, getExprNames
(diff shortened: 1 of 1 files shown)


tomkooij:
I this passes Travis I will add the testcase provided in #564 and merge.

FrancescAlted:
Good detective work!

tomkooij:
Thanks! I'm not sure if this does not break the queries in some other way. I'll make sure there are sufficient testcases before merging.

FrancescAlted:
Yes, tests are your best  friend.

tables/conditions.py
```diff
@@ -253,8 +253,13 @@ def fix_invert(idxcmp, exprnode, indexedcols):
             return [expr]
 
     # Recursively get the expressions at the left and the right
-    lexpr = _get_idx_expr_recurse(left, indexedcols, idxexprs, strexpr)
+    lexpr = deepcopy(_get_idx_expr_recurse(left, indexedcols, idxexprs,
+                                           strexpr))
+    if len(lexpr) == 2 and lexpr != not_indexable:
```
tomkooij: This is a bit of a hack. This is now necessary since `strexpr` is no longer a (reference to) a variable in a high scope (a sort of global).
It can probably be fixed by refactoring `add_expr` below.

tomkooij:
This is not a real fix. Fixing #564 requires a rewrite of  `_get_idx_expr` that correctly traverses a parsed expression tree.

Closing this, because it does not contribute to the solution.

PR #566, opened 2016-08-07
MNT: fix bad import

tacaswell:
No function H5Rget_object_type exists in hdf5 (as of 1.10).  There is a
H5Rget_obj_type.

This function is not used in pytables.


andreabedini:
any idea why this failed?

tacaswell:
Tried restarting the tests (still failed).  Closed/opened to target against current version of h5py branch (maybe I created this PR when the target branch was failing?)  Then went and checked the CI on the h5py branch:

https://travis-ci.org/PyTables/PyTables/builds/150586625

Which is failing with the same error.

PR #426, opened 2015-03-20
Appveyor support

joshmoore:
See: https://github.com/PyTables/PyTables/issues/424

This migrates the http://ci.openmicroscopy.org/job/PyTables build to Appveyor. The PR should be built automatically as with travis (and be green!) for merge.

PR #569, opened 2016-08-31
Use Appveyor CI

tomkooij:
Use Appveyor CI
Based on previous work by @joshmoore gh-426

Explicitly build wheels, to be able to upload wheels to PyPI
Build wheel's for python 2.7 and 3.5 (64bit)
Install from wheel using pip
Run tests

Minor changes, based on anaconda pytables recipe
blosc: Fix _xgetbv for MSVC 9.0 (Python 2.7)
Remove zlib1.dll and szip.dll from package


tomkooij:
This works!

The AppVeyor badge in the README (from 5e9ecda) seems to be wrong. I can't find the correct link on AppVeyor. Probably need to login as PytablesBot.

c-blosc/blosc/shuffle.c
```diff
@@ -123,10 +123,9 @@ static uint64_t _xgetbv(uint32_t xcr) {
     which means we can get away with returning a hard-coded value from
     this implementation of _xgetbv. */
 
-static inline uint64_t
-_xgetbv(uint32_t xcr) {
+static __inline uint64_t _xgetbv(uint32_t xcr) {
     /* A 64-bit OS must have XMM save support. */
-    return xcr == 0 ? (1UL << 1) : 0UL;
+    return (xcr == 0 ? (1UL << 1) : 0UL);
```
FrancescAlted: Hmm, is this something that we should modify in C-Blosc upstream?  Is this a problem with some Visual Studio compiler?  If so, which one?

tomkooij: Without the patch building fails on  MSVC2008 (=Visual C++ 9.0), which is used for CPython 2.7.

It's an inline defined for x64 targets with MSVC2010 SP1. So it might fail on MSVC2010 as well, but I haven't tested it.

I found the fix in the anaconda (and conda-forge) pytables recipe.

tomkooij:
This is now correctly checking PRs and building wheels for downloading. I'm going to check if the wheels (built against the conda hdf5 package) work with a "clean" windows install with the hdfgroup HDF5 dll's. Or if the .dll can be packaged as in: 
https://groups.google.com/forum/#!topic/pytables-dev/bvjIuE1J3J4

It would be very nice to have wheels on PyPI that just work. Especially for windows as building them is such a pain.

andreabedini:
+1 for bundling the dll on windows (so we pick what version we want to ship).

c-blosc/blosc/shuffle.c
FrancescAlted: Ok, not sure why python-blosc did not complain about this, but fixed in: https://github.com/Blosc/c-blosc/commit/54abf15278270dbb80578ba5760069c0a27f913e.  Let's try to update C-Blosc for the forthcoming release of PyTables.

tomkooij:
The `cp35-cp35m-win_amd64.whl` (x64 python 3.5) wheel works "out of the box" without an existing HDF5 installation. (`hdf5.dll` is bundled in the package).  I have to figure out how to package `zlib1.dll` and `szip.dll`.

tomkooij:
This now packages `zlib.dll` and `bzip2.dll` from conda. But there is no `szip.dll` in conda. Conda hdf5 seems to be built without szip.

The resulting wheels seem good enough for PyPI: There is BZIP2 support, but no LZO. All dependencies are included except for the MSVC Runtime.

andreabedini:
> This now packages zlib.dll and bzip2.dll from conda. But there is no szip.dll in conda. Conda hdf5 seems to be built without szip.

AFAIK there's some legal issue redistributing szip, is that right @PyTables/maintainers ?

FrancescAlted:
Yes, definitely . I don't think that distributing szip would be a good idea. Plus, I don't think there is a lot of people (not even one) using it. So I wouldn't worry - at all - with missing szip binaries in PyTables wheels.

Sent from my phone

El 5 set 2016, a les 2:05, Andrea Bedini <notifications@github.com> va escriure:

cgohlke:
Re SZIP: you could use [libaec](https://www.dkrz.de/redmine/projects/aec), which is a BSD licensed [drop-in replacement for SZIP](https://www.dkrz.de/redmine/projects/aec/wiki/SZIP).

tomkooij:
@cgohlke: Thanks! I'm not going to include `libaec` since szip is hardly used, but it's nice to know. I'll include a link it in the documentation.

PR #576, opened 2016-09-15
Make non-ascii returning __str__ work on Python2

### LINKED ISSUE #514: Groups and tables named with unicode
I want groups (tables) named with unicode

``` python
import tables
f = tables.open_file('test.h5', mode='w')
g0 = f.create_group(f.root, u"table_0")
g1 = f.create_group(f.root, u"table_1")
```

Then

``` python
f.list_nodes(f.root)
```

returns

```
[/table_0 (Group) ''
   children := [], /table_1 (Group) ''
   children := []]
```

Ok, but after group creation with

``` python
gb = f.create_group(f.root, u"table_β")
```

the code 

``` python
f.list_nodes(f.root)
```

raises exception

```
...
/usr/local/lib/python2.7/dist-packages/tables/node.pyc in newmethod(self)
     46             addr = hex(id(self))
     47             return '<closed %s.%s at %s>' % (cmod, cname, addr)
---> 48         return oldmethod(self)
     49     newmethod.__name__ = oldmethod.__name__
     50     newmethod.__doc__ = oldmethod.__doc__

/usr/local/lib/python2.7/dist-packages/tables/group.pyc in __repr__(self)
   1122         childlist = '[%s]' % (', '.join(rep))
   1123 
-> 1124         return "%s\n  children := %s" % (str(self), childlist)
   1125 
   1126 

UnicodeEncodeError: 'ascii' codec can't encode character u'\u03b2' in position 7: ordinal not in range(128)
```

The code

```
f.get_node(f.root, name=u"table_β")
```

doesn't work too. But [here](http://www.pytables.org/release-notes/RELEASE_NOTES_v3.0.x.html) it is written

> The File.get_node() method now accepts unicode paths (closes gh-203)

Is it is PyTables limitation or bug?

The only way to get groups I found is something like this

``` python
import h5py
f = h5py.File('test.h5')
for k in f.keys(): print k
```

returns

```
table_0
table_1
table_β
```

Versions: `python 2.7.6` on Ubuntu 14.04 (x64) with `tables=3.2.2` and `h5py==2.2.1`.

P.S. This issue is originated from [my question on stackoverflow](http://stackoverflow.com/questions/33344471/pandas-hdf5-tables-named-with-unicode). It was suggested as [possible bug](http://stackoverflow.com/questions/33344471/pandas-hdf5-tables-named-with-unicode#comment54598532_33344471) and it was adviced to post here.

andreabedini:
Uses class decorator `@python_2_unicode_compatible` from `six`.

```
A class decorator that takes a class defining a __str__ method. On
Python 3, the decorator does nothing. On Python 2, it aliases the
__str__ method to __unicode__ and creates a new __str__ method that
returns the result of __unicode__() encoded with UTF-8.
```

Closes #514

### DIFF
--- a/tables/attributeset.py
+++ b/tables/attributeset.py
@@ -72,6 +72,7 @@ def issysattrname(name):
         return False
 
 
+@six.python_2_unicode_compatible
 class AttributeSet(hdf5extension.AttributeSet, object):
     """Container for the HDF5 attributes of a Node.
 
--- a/tables/description.py
+++ b/tables/description.py
@@ -313,6 +313,7 @@ Time64Col = Col._subclass_from_prefix('Time64')
 
 # Table description classes
 # =========================
+@six.python_2_unicode_compatible
 class Description(object):
     """This class represents descriptions of the structure of tables.
 
--- a/tables/exceptions.py
+++ b/tables/exceptions.py
@@ -23,6 +23,7 @@ import warnings
 import traceback
 
 
+@six.python_2_unicode_compatible
 class HDF5ExtError(RuntimeError):
     """A low level HDF5 operation failed.
 
--- a/tables/file.py
+++ b/tables/file.py
@@ -576,6 +576,7 @@ class NodeManager(object):
                 node._f_close()
 
 
+@six.python_2_unicode_compatible
 class File(hdf5extension.File, object):
     """The in-memory representation of a PyTables file.
 
--- a/tables/filters.py
+++ b/tables/filters.py
@@ -57,6 +57,7 @@ _bitshuffle_flag = 0x8
 
 # Classes
 # =======
+@six.python_2_unicode_compatible
 class Filters(object):
     """Container for filter properties.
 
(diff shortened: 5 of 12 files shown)


andreabedini:
It still fails the tests. The [error](https://travis-ci.org/PyTables/PyTables/jobs/160344876#L1424) looks weird to me, I'll need to debug a bit more.

tomkooij:
Why did I approve the commit, when the test fail???

Anyway, the decorator `@six.python_2_unicode_compatible` messes up `group.__dict__`: (or the representation of the objects in the dict). 

I added `print(self.h5file.root.__dict__)` to the failing testcase and all the keys of the __dict__ are wrong.

The `__str__` function seems to point to the wrong class. (Decorator get's applied only the first time upon object/class creation?)

andreabedini:
@tomkooij I might have fixed this, any chance it can make it to 3.4.0 (if the tests pass?)

PR #578, opened 2016-09-24
setup.py: gracefuly handle cpuinfo failure

keszybz:
When comping on aarch64, setup.py fails with:
Exception: py-cpuinfo currently only works on X86 and some ARM CPUs.
Assume that the CPU does not support SSE2 or AVX in importing cpuinfo fails
and continue.

(Package builds fine and tests pass with this patch:
https://kojipkgs.fedoraproject.org//packages/python-tables/3.3.0/1.fc26/data/logs/aarch64/build.log)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,8 +32,13 @@ import distutils.spawn
 from setuptools.command.build_ext import build_ext
 
 # For guessing the capabilities of the CPU for C-Blosc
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Nice patch.  Thanks!

PR #582, opened 2016-10-31
HDF5 1.10

avalentino:
This PR fixes compatibility with HDF5 v1.10 using the old 1.8 API.

This PR fixes issue #545.

The fix is backward compatible and should be almost complete.

A minor issue remains: it seems that the naming convention for hdf5 libraries is changed.
Now (v1.10 ?) when one builds HDF5 from code libraries are installed with the `-shared` and `-static` suffixes. The current setup.py does not play well with this naming but the easy workaround is to create symlinks with old names.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -105,8 +105,7 @@ optional.
 Installation
 ------------
 
-1. Make sure you have HDF5 version 1.8.4 or above. HDF5 1.10.x is not
-supported.
+1. Make sure you have HDF5 version 1.8.4 or above.
 
    On OSX you can install HDF5 using `Homebrew <http://brew.sh>`_::
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,8 @@
 environment:
-
-  # Make setup.py include conda specific dll's in wheels
+  PACKAGE_NAME: tables
+  BUILD_DEPENDS: "numpy>=1.8.0 numexpr>=2.5.2 six cython bzip2"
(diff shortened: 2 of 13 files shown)


andreabedini:
This looks great @avalentino! 💥 LGTM

tomkooij:
Nice work! +1

I think we need to test against HDF5 1.8 and 1.10 on our CI systems at least until HDF5 1.8 is phased out. Or at least extend to test matrix to test python 3.5 against HDF5 1.10, while keeping HDF5 1.8 in the matrix. 

I'll try to do that for Appveyor, when time permits.

tomkooij:
2bbfa36 implements testing against HDF5 v1.10.x (as well as v1.8.x) on appveyor/windows. This is done by installing HDF5 v1.10 conda packages from a custom channel (my channel). 
Installing HDF5 binaries is not an option on windows: There are no python 2.7 (vc9) compatible binaries. Building the lib on windows is not trivial: I have build the HDF5 v1.10 packages using a forked conda-forge feedstock. 

Testing against HDF5 v1.10 on travis is not implemented yet. There are no apt packages for v1.10 on travis/ubuntu AFAIK. We could easily build the lib from source or find an apt repo that has v1.10 or even switch to conda. I'm leaving that one for someone else :-)

avalentino:
Which is the state of this PR?
Someone is going to merge it?

tomkooij:
I'm +1 for merging.

avalentino:
Merged in c7b13097f9a5c0f36cd41448f8e3d31026dd1d0a

PR #606, opened 2017-04-07
Add cython to requirements list

gut:


tomkooij:
Thanks for taking the time for sending a PR! However, I cannot merge this.

`cython` is not a runtime requirement. It is only needed for building from source. If we put `cython` in `requirements.txt` it will be installed by `pip` even when a wheel is used for installation.

gut:
Right! the `requirements.txt` list is only for runtime dependencies. 

Sorry about that and thanks for taking the time to reply it.

PR #581, opened 2016-10-26
accept windows drive letter in path

oscar6echo:
Every time I use pytables (manipulating HDF5 files from pandas) I have to manually change these 2 lines. May be they could be added to the 'master' branch.

What do you think ?

PR #611, opened 2017-04-12
Release 3.4.1

tomkooij:
This is against `develop`... 👎 Will fix if CI pass.

Add CI tests for pt utils
Fix identation in ptrepack.py (but I think it will still fail)


tomkooij:
#581 still fails on windows.  Reverted.

tomkooij:
There seems to be a segfault in one off the Appveyor builds (python 3.6+hdf5 1.10). The identical "branch" build passes. I'm guessing this is an appveyor thing and merging.

PR #614, opened 2017-04-13
Find packages in conda env

tomkooij:
~~set HDF5_DIR to $CONDA_PREFIX if HDF5_DIR is undefined.~~

This tries to find packages in conda env if package location (HDF5_DIR...) is undefined.

TODO:
 - [x] Test on Windows.
 - [x] Test on Linux
 - [x] ~~figure out LZO package~~ not worth the trouble.
 - [x] add some kind of `--no-conda` flag to disable this feature

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,7 +42,7 @@ install:
   # as well as pip, conda-build, and the binstar CLI
   - powershell .\\ci\\appveyor\\install.ps1
   - powershell .\\ci\\appveyor\\missing-headers.ps1
-  - "set PATH=%PYTHON%;%PYTHON%\\Scripts;%PYTHON%\\Library\\bin;%PATH%"
+  - "set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
 
 build_script:
   - conda config --add channels conda-forge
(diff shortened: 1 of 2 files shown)


tomkooij:
This breaks the following case:

When `setup.py` is ran inside a conda env, but without the conda hdf5 package, it will no longer look for the system default HDF5 in PATH. 

This can be interpreted as a feature: Inside a conda env, you will want to use the conda hdf5 package.
This can be interpreted as a bug: If the hdf5 package is purposely missing, you will now need to explicitly set HDF5_DIR to point to the system hdf5lib.

It is certainly a change, but IMHO it is a feature. It will probably make it easier to build pytables inside conda envs (readthedocs).

tomkooij:
This seems finished.

This is tested on Windows through Appveyor. 

Travis / Linux does not use appveyor, so we need to be careful. I've tested on Ubuntu 16.04 LTS with and without conda (and using both conda libs and the system defaults libs from within conda). It looks like this is well-behaved.

tacaswell:
I trick I have used on other project is to use the patching mechanism of conda-build to insert the place holders that get replaced during conda's install process

ex https://github.com/NSLS-II/lightsource2-recipes/tree/master/recipes-tag/filestore

doesn't work with source-insall in a conda-env, but for built packages 'just works' with out having to identify that you are in a conda env etc.

tomkooij:
@tacaswell : Thanks. 

I'm looking for a way to deal with installing from source in conda-env.

Main problem is that conda packages do not seem to have `pkg-config` *.pc information. (At least hdf5 and bzip2 lack these).

I really don't like this solution: It is not the "KISS" way of doing things.
 I'd be much more happy with being able to use `pkg-config`. I'll have a look after the Easter holidays.

Don't merge!

tomkooij:
Closing this in favor of #615

PR #615, opened 2017-04-14
Detect conda env try to use conda packages

tomkooij:
Use installed package if location not otherwise specified.
Add `--no-conda` flag to `setup.py` to disable.
Use CONDA_PREFIX\\Library on windows.

PEP8 fixes to setup.py
Remove HDF5_DIR etc from appveyor

This does not use `conda list`. This is the KISS way.  

Issues:
 - It breaks finding packages using `pkg-config` inside a conda env, when the package is not in conda env but is installed on the system (with pkg-config info)


tomkooij:
I'll add the docs.

Issue #616, opened 2017-04-15
ptrepack arguments: Absolute Windows path provokes crash


oscar6echo:
If you run ptrepack on a windows machine with absolute source and destination filenames, say F:\path\to\this\folder and F:\path\to\that\folder, then the command will fail as the args.src.split(':') 
(lines 425-426 in PyTables/tables/scripts/ptrepack.py) will separate the F from the rest of the path.
On Windows the .split(':') is irrelevant as it cannot separate file names, so a change could be something like:

````
# Catch the files passed as the last arguments
if 'win' in sys.platform():
    # the file separator for windows is deemed ','
    src = args.src.split(',')
    dst = args.dst.split(',')
else:
    # the file separator for osx, linux for  is deemed ':'
    src = args.src.split(':')
    dst = args.dst.split(':')
`````

instead of current lines lines 424-426:
````
# Catch the files passed as the last arguments
src = args.src.split(':')
dst = args.dst.split(':')
`````

oscar6echo:
Clarification about the issue description.  

Absolute Windows path in ptrepack, pttree, ptdump will provoke crash.  
Because an absolute Window path contains an extra colon - as opposed to Linux/macOS.  
e.g. "F:\path\to\file".  

It is NOT about having different separators for Windows. It is about allowing Absolute Windows paths as arguments for ptrepack, pttree, ptdump.

PR #617, opened 2017-04-15
Allow Windows absolute file paths in ptrepack args

oscar6echo:
Allow different file separators based on OS: ":" for
Linux/macOS and "," for Windows. Clarify documentation
(available via "ptreapck --help") accordingly.
Not test included. Need help on that. Any example ?

Resolves: #616
See also: commit 476a2f956322569efefa5d779b457d639b90a552

### DIFF
--- a/tables/scripts/ptrepack.py
+++ b/tables/scripts/ptrepack.py
@@ -422,8 +422,15 @@ def main():
         )
 
     # Catch the files passed as the last arguments
-    src = args.src.split(':')
-    dst = args.dst.split(':')
+    if 'win' in sys.platform():
+        # the file separator for windows is deemed ','
+        src = args.src.split(',')
+        dst = args.dst.split(',')
+    else:
+        # the file separator for osx, linux for  is deemed ':'
+        src = args.src.split(':')
+        dst = args.dst.split(':')
+
     if len(src) == 1:
         srcfile, srcnode = src[0], "/"
     else:
--- a/tables/scripts/ptrepack.py
+++ b/tables/scripts/ptrepack.py
@@ -422,14 +422,8 @@ def main():
         )
 
     # Catch the files passed as the last arguments
-    if 'win' in sys.platform():
-        # the file separator for windows is deemed ','
-        src = args.src.split(',')
-        dst = args.dst.split(',')
-    else:
-        # the file separator for osx, linux for  is deemed ':'
-        src = args.src.split(':')
-        dst = args.dst.split(':')
+    src = args.src.rsplit(':', 1)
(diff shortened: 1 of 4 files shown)


tomkooij:
Nice work. Thanks! 

I'll add a basic test (probably after easter) andere let you finish.

Thanks for contributing!

tables/scripts/ptrepack.py
```diff
@@ -422,8 +422,15 @@ def main():
...
     # Catch the files passed as the last arguments
-    src = args.src.split(':')
-    dst = args.dst.split(':')
+    if 'win' in sys.platform():
+        # the file separator for windows is deemed ','
+        src = args.src.split(',')
+        dst = args.dst.split(',')
+    else:
+        # the file separator for osx, linux for  is deemed ':'
+        src = args.src.split(':')
```
tomkooij: What about: `src = args.src.rsplit(':', 1)`.

That should work on windows and posix.

tomkooij: I just realised `rsplit(':', 1')` was your original commit 476a2f956322569efefa5d779b457d639b90a552

Sorry about that, but I think it's the way to go...

oscar6echo: If ptrepack arguments are sourcefile and destfile, yes I agree.  

If on Windows and ptrepack arguments are sourcegroup and destgroup each a sequence of absolute paths starting with a drive letter e.g. F:/path/to/file then it seems to me splitting on ":" won't work. But personally I always use ptrepack with sourcefile and destfile. Not sure it is a frequent case...

Anyway you're the maintainer and decide. So I've updated the PR as you suggested.   
I'll wait to see how you write the test to learn how for next time.  

Thx

tomkooij:
I added testcases for `ptrepack`, `ptdump` and `pttree`. These utils all have the same windows drive letter (extra semi colon) issue.

Would you mind fixing ptdump and pttree as well?

The testcases are a bit over-engineered, but I could not think of an easier way to do this, without all the patching.

tomkooij:
Argh! @oscar6echo: Please work from a branch instead of `develop`. I accidentally pushed your branch into `PyTables:develop` instead of `oscar6echo:develop`. This caused the merge...

I already reset the state of `PyTables:develop`. But I cannot force push `oscar6echo:develop`.

I hope you can still force push this. If you reset it back to your last commit, I will add the new commits.

oscar6echo:
Ok, I restarted from scratch and followed your last suggestion.
I opened a new PR #618 and clarified issue #616 description.
PR #618 does not contain the tests. Please add them again.

Hope that'll work !

PR #618, opened 2017-04-16
Allow Windows absolute paths in ptutils args

oscar6echo:
ptdump, pttree and ptrepack use 'path:/group/node' as input.
On windows this could be: 'drive:\folder\filename:/group/node' (extra colon)
No testcase included.

Resolves: #616
See also: commit 476a2f956322569efefa5d779b457d639b90a552

### DIFF
--- a/RELEASE_NOTES.txt
+++ b/RELEASE_NOTES.txt
@@ -12,7 +12,10 @@ Changes from 3.4.1 to 3.4.x
 ===========================
 
 
-
+Bugs fixed
+----------
+ - Fixed windows absolute paths in ptrepack, ptdump, ptree.
+   :issue:`616`. Thanks to oscar6echo.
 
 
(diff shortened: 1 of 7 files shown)


tomkooij:
Thanks for creating the new PR!

I added the testcases. Let's wait and see what the CIs think.

tomkooij:
The testcase fail on the patched `print` on python 2.7. They pass on my local py 2.7. Anyway, I'll have a look. (But not today ;) )

tomkooij:
Python 2.7 on appveyor fails again... I'm going to skip this testcase on python 2.7 and just go with `unittest.mock`.

tomkooij:
Finally!

Thanks @oscar6echo.

oscar6echo:
@tomkooij You're welcome. Good !

PR #624, opened 2017-04-24
feat(group): autocomplete valid python-identifiers

ankostis:
Implement `__dir__()` on group nodes to include in the list of autocompletions also those children that are named as valid python identifiers.

PR #625, opened 2017-04-25
feat(attributes): __dir__() autocompletes also valid py-indentifiers

ankostis:
Similar to #624, implement also `AttributeSet.__dir__()`  to return in its autocompletion list those attributes that are named as valid python identifiers.

+ Docs updated.
+ Simple test-case.
+ Not suitable for PY-2 (no `__dir__()` in datamodel).

### DIFF
--- a/RELEASE_NOTES.txt
+++ b/RELEASE_NOTES.txt
@@ -12,9 +12,9 @@ Changes from 3.4.2 to 3.4.x
 
 Improvements
 ------------
- - Autocomplete group-children that are named as valid python identifiers
-   on interactive python sessions. 
-   :issue:`624` thanks to ankostis.
+ - On interactive python sessions, group/attribute  `__dir__()` method
+   autocompletes children that are named as valid python identifiers.
+   :issue:`624` & :issue:`625` thanks to ankostis.
 
 
 Changes from 3.4.1 to 3.4.2
--- a/tables/attributeset.py
+++ b/tables/attributeset.py
@@ -26,6 +26,7 @@ from .exceptions import ClosedNodeError, PerformanceWarning
 from .path import check_attribute_name
 from .undoredo import attr_to_shadow
 from .filters import Filters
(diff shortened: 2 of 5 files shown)


tomkooij:
Nice improvement to your previous PR.

I stumbled onto this: http://stackoverflow.com/questions/5474008/regular-expression-to-confirm-whether-a-string-is-a-valid-identifier-in-python
TL;DR: in python 3 identifiers can be unicode. Can you add the unicode thing in your `tables.utils.is_python_identifier`

Perhaps add that `SYS%` (that is currently failing) to the doctest for `tables.utils.is_python_identifier`. There are some other nice tests in the SO post linked above.

ankostis:
Thanks for the infos.
I had to filter all attributes not to return invalid identifiers - so there was a bug lurking in the lib :-)

My own bug (also included in #624) was that I had copied my code from an ambitious plan, where I was hopping to smooth-out *all* attribute names in `dir`, and retrofit `__getattr__()` to translate them.
But it was too much fuzz.

Travis pass ([225996694](https://travis-ci.org/PyTables/PyTables/builds/225996694)), we now have to wait for the usual appveyor's sequential bottleneck.

ankostis:
Oups, now generating double-items!
Must augment terst-case and edit rewrite once more.

tomkooij:
Unfortunately I cannot kill the appveyor jobs. (No access...) I need to implement an automatic skip-if-new sometime.

ankostis:
It's fine now.

ankostis:
Or should it bail out on PY-2?
Currently i expect pressing tab on PY2 to fail with exception of is identifier attribute not found.

tomkooij:
I tested the py2 wheel (from appveyor) on my local machine. It's fine: Pressing tab does not work, but does not throw an exception (in IPython).

I think this is ready to merge. Any additions planned, or shall I merge this?

ankostis:
No other ideas, :-) it's ready!
Thanks.

Issue #270, opened 2013-08-01
tables.group.Group class should have __getitem__ attribute


vshkolyar:
Reason for bugs or in other words inconsistency with DRY (don't repeat yourself), SSOT (Single Source of Truth) or whatever it's called software principles:
If developer creates some object, property, in his code that he can refer to it latter in his code, he wand there will be only one generic way to do that. In case of the group, if I create for example root group in HDF by using string value, I will create some static field that I will use to creation and retrieving of the group/table. Or for some generic use the name could be passed to me as var by user of my method and I in my code should not know what is the name, I only will want to pass the name to root group and retrieve the son node by name. Or for example in refactoring of some large code base: I will want to change only the name of the node in the static string field value and not to remember to seek in code for places where the string used as code and not the string.
Sorry if the description is not the best, English is not my mother tongue :)

FrancescAlted:
Closing as I don't understand that very well either.

PR #628, opened 2017-05-05
Define `Group.__getitem__()` method

ankostis:
Since `tables.group.Group` class defines `__contains__()` and `iter__()` methods, it should also define `__getitem__()`, and work as a proper *container*.  

That would make the code below to work intuitively:

```python
>>> 'group 1' in hfile.root
>>> g1 = hfile.root['group 1']
>>> g1['some &^%&^* named group']._v_pathname 
/group 1/some &^%&^* named group
```
For the record, that must have been the purpose of #270.

### Bonus commit: 
Improv unique checks on `__dir__()` TCs.

### DIFF
--- a/RELEASE_NOTES.txt
+++ b/RELEASE_NOTES.txt
@@ -15,6 +15,9 @@ Improvements
  - On interactive python sessions, group/attribute  `__dir__()` method
    autocompletes children that are named as valid python identifiers.
    :issue:`624` & :issue:`625` thanks to ankostis.
+ - Implement `Group.__getitem__()` to have groups act as python-containers,
+   so code like this works: ``hfile.root['some child']``.
+   :issue:`628` thanks to ankostis.
 
 
 Changes from 3.4.1 to 3.4.2
--- a/tables/group.py
+++ b/tables/group.py
@@ -434,6 +434,16 @@ class Group(hdf5extension.Group, Node):
             return False
         return True
 
+    def __getitem__(self, childname):
+        """Return the (visible or hidden) child with that `name` ( a string).
+
+        Raise IndexError if child not exist.
+        """
+        try:
+            return self._f_get_child(childname)
+        except NoSuchNodeError:
+            raise IndexError(childname)
+
     def _f_walknodes(self, classname=None):
         """Iterate over descendant nodes.
 
@@ -827,6 +837,9 @@ be ready to see PyTables asking for *lots* of memory and possibly slow I/O."""
         if name in self._c_lazy_children_attrs:
             self._g_add_children_names()
             return self.__dict__[name]
+        return self._f_get_child(name)
+
+    def __item__(self, name):
         return self._f_get_child(name)
 
     def __setattr__(self, name, value):
--- a/tables/tests/test_attributes.py
+++ b/tables/tests/test_attributes.py
@@ -639,6 +639,10 @@ class CreateTestCase(common.TempFileMixin, TestCase):
         self.assertNotIn(bad_user, completions)
         self.assertEqual(completions.count(user_attr), 1)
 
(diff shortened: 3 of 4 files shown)


tomkooij:
Thanks for the extra testcases.

About `Group.__getitem__`:

` >>>'group 1' in hfile.root`
This already works.

`>>> g1 = hfile.root['group 1']`
This is very ugly. 
The syntax `g1 = hfile.get_node('/group1')` is much more clean IMHO.

I won't **block** `__getitem__` but I do not feel that is it a necessary extension. I'd like to see what other devs think. pinging @PyTables/core

ankostis:
In python it is very much expected if `foo in group` works, also `group[foo]` to work;  it is the read-only ["container" data-model](https://docs.python.org/3/reference/datamodel.html#emulating-container-types).  What is more puzzling, is that `atttribute_sets` work already like containers, but `groups` not.

> ```>>> g1 = hfile.root['group 1']```
> This is very ugly.

It's strange that you say so :-)
I would say, having everything working as a containers is the "zen" of python, but that's my opinion...

Anyway, I will implement it, and you may decide about it.
Personally I'm always monkey-patching `Group` class, not to have to remember in which cases indexing works or not.

ankostis:
I forgot to explain that I referring to interactive use, e.g. Jupter notebooks, where you use the `[Tab]` completion to review the existing children, and then type as few as possible keystrokes to access the child. And work your way gradually doen the hierarchy.
I had this moment today where I was presenting my notebook to a colleague and had to explain why simple indexing did not function, and dig my way through `_v_...` methods.

FrancescAlted:
Yeah, I tend to agree that PyTables philosophy is to treat children as attributes (i.e. using 'natural naming'), and not as entries in a dictionary.  Having said that, practicality beats purity, and if people like using the `Group` class as a dictionary, then I don't see strong reasons so as to not allowing that. So +1 on accepting this PR.

@ankostis I don't think I get your point in what the trouble is using `[TAB]` in the Jupyter notebook.  This should work in PyTables, or it does not for you?

ankostis:
Well, I was a bit confusing in my previous posting.
I had irregular-named groups, so `[Tab]` would not reveal them, but console was printing them, so I was trying expressions like `root['some child'].other_child.more['(g2)']` and they didn't work, and had a hard time finding which method to use to achieve this "gradual descend".  

I fixed my posting above.

RELEASE_NOTES.txt
```diff
@@ -24,6 +24,9 @@ Improvements
 ------------
  - setup.py detects conda env and uses installed conda (hdf5, bzip2, lzo
    and/or blosc) packages when building from source.
+ - On interactive python sessions, group/attribute  `__dir__()` method
```
tomkooij: This was not released yet. `v3.4.2` does not include PR 624 and PR 625. Please move this to "Changes from 3.4.2 to 3.4.x".

ankostis:
Release-changes fixed as requested.

tomkooij:
Thanks!

tables/group.py
```diff
@@ -829,6 +839,9 @@ def __getattr__(self, name):
             return self.__dict__[name]
         return self._f_get_child(name)
 
+    def __item__(self, name):
+        return self._f_get_child(name)
+
```
avalentino: Sorry, probably I'm missing something, but what is for  `__item__` exactly? If I'm not it is not a standard dunder.

tomkooij: O... I missed that in my review... 

I'll remove it.

tomkooij: Removed in cc1175425397b4906ef909538345d4daa815029a

avalentino: thanks

PR #630, opened 2017-05-15
speedup in `_where`

llllllllll:
I saw my code spending a lot of time in the coords = [p.nrow ...] comprehension, this change should speed up the creation of the coordinates array dramatically. The monotonic check removal is because it took a long time to create the array to do the check, it seems to be fine to just skip it because the savings are lost by allocating and doing the elementwise compare.

The idea of the change is to do the iteration directly in cython to produce a coordinate array.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ tables/*.so
 
 # specific files
 src/version.h
+
+
+tmp/*
+*.egg-info/*
--- a/tables/table.py
+++ b/tables/table.py
@@ -1480,9 +1480,9 @@ very small/large chunksize, you may want to increase/decrease it."""
 
         """
 
-        return self._where(condition, condvars, start, stop, step)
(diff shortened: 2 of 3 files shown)


scopatz:
This seems reasonable to me.  Any idea what kind of speedup is seen?

llllllllll:
Not yet, I haven't tested too much. I wanted to make sure this seemed like a reasonable change before investing too much time. I will try to test this with the different `__next__` implementations with different sized tables.

FrancescAlted:
The problem that I see with this approach is that you are malloc'ing a potentially big `out` array (as long as the length of the table) for the coordinates, and this might require a lot of memory.  I agree that as Linux does overcommit, this is rarely a problem, but this does not apply to Mac OSX and Windows.  Maybe you can start with a small `out` array and enlarge it as needed?

PR #634, opened 2017-06-19
Arraycore

albertosm27:
First implementation of the new array using h5py backend. The objective of this branch is to pass the basic tests in tables/tests/test_array.py.

### DIFF
--- a/tables/core/array.py
+++ b/tables/core/array.py
@@ -9,7 +9,11 @@ class Array(Leaf):
     def __init__(self, _atom=None, **kwargs):
         super().__init__(**kwargs)
         self.atom = _atom
+        if _atom is None or _atom.shape == ():
+            self.atom = Atom.from_dtype(self.dtype)
         self.nrow = None
+        # Provisional for test
+        self.flavor = 'numpy'
         # TODO iterators?
 
     @property
--- a/tables/core/array.py
+++ b/tables/core/array.py
@@ -1,5 +1,61 @@
+import sys
+import numpy
 from .leaf import Leaf
+from tables import Atom
+from tables.utils import byteorders
 
 
 class Array(Leaf):
-    pass
+    def __init__(self, _atom=None, **kwargs):
+        super().__init__(**kwargs)
+        self.atom = _atom
+        self.nrow = None
+        # TODO iterators?
+
+    @property
+    def rowsize(self):
+        "The size of the rows in bytes in dimensions orthogonal to *maindim*."
+        maindim = self.maindim
+        rowsize = self.atom.size
+        for i, dim in enumerate(self.shape):
+            if i != maindim:
+                rowsize *= dim
+        return rowsize
+
+    @property
+    def size_in_memory(self):
+        """The size of this array's data in bytes when it is fully loaded into
+        memory."""
+        return self.nrows * self.rowsize
+
+    def _g_create(self):
+        """Save a new array in file."""
+
+    def get_enum(self):
+        if self.atom.kind != 'enum':
+            raise TypeError("array ``%s`` is not of an enumerated type"
+                            % self._v_pathname)
+
+        return self.atom.enum
+
+    def read(self, start=None, stop=None, step=None):
+        (start, stop, step) = self._process_range_read(start, stop, step)
+        if self.nrows == 1:
+            return numpy.array(self[()])
+        else:
+            return self[start:stop:step]
+
+    def iterrows(self, start=None, stop=None, step=None):
+        start, stop, step = self._process_range(start, stop, step,
+                                                warn_negstep=False)
+        # Scalar dataset
+        if self.nrows == 1:
+            self.nrow = 0
+            yield numpy.array(self[()])
+        else:
+            self.nrow = start
+            for r in self[start:stop:step]:
+                self.nrow += step
+                yield r
+
+    __iter__ = iterrows
--- a/tables/core/group.py
+++ b/tables/core/group.py
@@ -163,7 +163,7 @@ class File(HasChildren, Node):
         super().close()
 
     def __enter__(self):
-        self.open()
+        return self
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
         self.close()
--- a/tables/core/group.py
+++ b/tables/core/group.py
@@ -35,11 +35,6 @@ class HasChildren:
             yield child.name
 
     def __getitem__(self, item):
-        # Try cache first
-        nmanager = self._file._node_manager
-        node = nmanager.get_node(item)
-        if node:
-            return node
         # No luck, so use the backend to lookup the item
         value = self.backend[item]
         if isinstance(value, abc.Group):
@@ -61,7 +56,8 @@ class HasChildren:
         elif isinstance(old, str):
             self.backend.rename_node(old, new_name)
         else:
-            raise TypeError("Expecting either the name of the node to rename or the node itself")
+            raise TypeError(
+                "Expecting either the name of the node to rename or the node itself")
 
     def remove_node(self, *args):
         """ This method expects one argument (node) or two arguments (where, node) """
@@ -74,8 +70,8 @@ class HasChildren:
                 self.backend.remove_node(name)
             else:
                 raise TypeError("Expecting either the name of the node "
-                        "to rename or the node itself when called with "
-                        "one argument")
+                                "to rename or the node itself when called with "
+                                "one argument")
         elif len(args) == 2:
             where, name = args
             where.remove_node(name)
@@ -156,8 +152,6 @@ class File(HasChildren, Node):
         super().__init__(**kwargs)
         # TODO (re) make this configurable
         # node_cache_slots = params['NODE_CACHE_SLOTS']
-        node_cache_slots = 10
-        self._node_manager = NodeManager(nslots=node_cache_slots)
         # TODO only show Filters the inputs it wants
         self._filters = Filters(**self.backend.params)
         # Bootstrap the _file attribute for nodes
@@ -165,7 +159,7 @@ class File(HasChildren, Node):
 
     def close(self):
         # Flush the nodes prior to close
-        self._node_manager.flush_nodes()
+        self.backend.flush()
         super().close()
 
     def __enter__(self):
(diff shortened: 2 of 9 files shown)


FrancescAlted:
Looks good to me so far.  I specially like the first attempt at getting rid of the node manager as h5py should be the place to handle this.

tomkooij:
Nice work!

tables/core/group.py
```diff
@@ -92,12 +96,46 @@ def filters(self, filters):
...
+        if byteorder is None:
+            _byteorder = np_byteorders['irrelevant']
+        else:
+            _byteorder = np_byteorders[byteorder]
+
+        if obj is None:
+            if atom is None or shape is None:
+                raise TypeError('if the obj parameter is not specified '
+                                '(or None) then both the atom and shape '
+                                'parametes should be provided.')
```
FrancescAlted: parameters

tables/core/array.py
```diff
@@ -42,20 +45,46 @@ def get_enum(self):
...
+            bytes_required = self.rowsize * nrowstoread
+            # if buffer is too small, it will segfault
+            if bytes_required != out.nbytes:
+                raise ValueError(('output array size invalid, got {0} bytes, '
+                                  'need {1} bytes').format(out.nbytes,
+                                                           bytes_required))
+            if not out.flags['C_CONTIGUOUS']:
+                raise ValueError('output array not C contiguous')
+
+            np.copyto(out, arr)
```
FrancescAlted: You can avoid the `arr` temporary by using [`read_direct(array, source_sel=None, dest_sel=None)`](http://docs.h5py.org/en/latest/high/dataset.html).

tables/core/group.py
```diff
@@ -92,12 +96,46 @@ def filters(self, filters):
...
+            else:
+                # Making strides=(0,...) below is a trick to create the
+                # array fast and without memory consumption
+                dflt = np.zeros((), dtype=atom.dtype)
+                obj = np.ndarray(shape, dtype=atom.dtype, buffer=dflt,
+                                 strides=(0,) * len(shape))
+        else:
+            flavor = flavor_of(obj)
+            # use a temporary object because converting obj at this stage
+            # breaks some test. This is soultion performs a double,
```
FrancescAlted: "This is soultion performs " -> "This fix performs".

tables/core/group.py
```diff
@@ -92,12 +96,46 @@ def filters(self, filters):
...
+                                'parametes should be provided.')
+            else:
+                # Making strides=(0,...) below is a trick to create the
+                # array fast and without memory consumption
+                dflt = np.zeros((), dtype=atom.dtype)
+                obj = np.ndarray(shape, dtype=atom.dtype, buffer=dflt,
+                                 strides=(0,) * len(shape))
+        else:
+            flavor = flavor_of(obj)
+            # use a temporary object because converting obj at this stage
```
FrancescAlted: [STYLE] Start comments in its own line with an uppercase letter.

tables/core/leaf.py
```diff
@@ -35,7 +41,10 @@ def nrows(self):
             return int(self.shape[self.maindim])
         # Scalar dataset
         else:
-            return 1
+            if self.shape == ():
+                return 1
+            else:
+                return len(self)
```
FrancescAlted: For compactness you can make use of the ternary operator there:

```
return 1 if self.shape == () else len(self)
```

tacaswell:
Where did the work from the Perth workshop end up?

tacaswell:
:sheep: Never mind, I sorted it out (that code is on the pt4 branch which this is a PR into).

tables/utils.py
```diff
@@ -33,6 +33,12 @@
     '|': 'irrelevant',
 }
 
+np_byteorders = {
+    'big': '>',
+    'little': '<',
+    'irrelevant': '|'
```
tacaswell: 'irrelevant' should be 'native'

tables/core/group.py
```diff
@@ -187,17 +220,117 @@ def __contains__(self, item):
...
+            basepath = where
+            nodepath = join_path(basepath, name or '') or '/'
+            node = self.root[nodepath]
+        else:
+            raise TypeError(
+                "``where`` must be a string or a node: %r" % (where,))
+
+        # Finally, check whether the desired node is an instance
+        # of the expected class.
+        if classname:
```
tacaswell: might be worth checking `classname is not None` so other 'falsy' values (which should probably raise) do not safely pass.

tables/core/group.py
```diff
@@ -187,17 +220,117 @@ def __contains__(self, item):
...
 
-    def create_table(self, where, name, desc, *args, **kwargs):
+    def create_table(self, where, name, desc, *args, createparents=False, **kwargs):
+        if not hasattr(where, 'create_table'):
+            where = self._get_or_create_path(where, createparents)
         return where.create_table(name, desc, *args, **kwargs)
 
-    def get_node(self, where):
-        return self.root[where]
+    def get_node(self, where, name=None, classname=None):
```
tacaswell: This is to match the API on `tables.file.File.get_node` ?

tacaswell:
https://github.com/albertosm27/PyTables/pull/1 <- getting copy to work.

tacaswell:
```
$ py.test tables/tests/test_array.py 

============ 64 failed, 307 passed in 6.19 seconds ============
```

on the PR into @albertosm27 's branch.

tables/utils.py
FrancescAlted: 'irrelevant' is meant for ASCII strings and the like, where the order is, well, irrelevant.

tables/tests/test_carray.py
```diff
@@ -2209,7 +2209,7 @@ def test00b_zeros(self):
             print("First row-->", ca[0])
             print("Defaults-->", ca.atom.dflt)
         self.assertTrue(allequal(ca[0], numpy.zeros(N, 'S3')))
-        self.assertTrue(allequal(ca.atom.dflt, numpy.zeros(N, 'S3')))
+        self.assertTrue(allequal(ca.atom.dflt, b""))
```
tacaswell: Why change the tests like this?

albertosm27: Multidimensional dtypes are giving me a hard time. 
I thought h5py as numpy was putting the dtype.shape inside the array.shape so I did that for the Atom.dflt (equivalent to fillvalue in h5py) and thats why I needed to change the test like that.

I still do not know if multidimensional dtypes are fully supported in h5py, for example in tables/tests/array_mdatom.h5 you can get '/arr' and it will be a multidimensional array but trying it to copy (with create_dataset) will raise an error like 'Can't broadcast shape (5, 5, 5) to (5, 5, 5, 3)' the 3 comes from the dtype shape.

Using the h5py copy method from Group will copy the array correctly with its multidensional dtype (still reading the array returns it as numpy with the dtype shape inside the array shape)

tacaswell: What do you mean by 'multidimensional dtypes'?  

`h5py` should support all dtypes that hdf5 supports (and if any are missing it should be fixed in h5py).

albertosm27: I mean dtypes with shape like this ```numpy.dtype('(3,)i4')```.

tacaswell: I actually think this is a numpy bug

```python
In [58]: np.zeros(5, dtype=np.dtype('(3,)i4')).shape
Out[58]: (5, 3)

In [59]: np.zeros(5, dtype=np.dtype('(3,)i4,f4')).shape
Out[59]: (5,)

In [60]: np.zeros(5, dtype=np.dtype('(3,)i4')).dtype
Out[60]: dtype('int32')

In [61]: np.zeros(5, dtype=np.dtype('(3,)i4,f4')).dtype
Out[61]: dtype([('f0', '<i4', (3,)), ('f1', '<f4')])

```

tacaswell: Maybe:

```python
In [62]: np.zeros(5, dtype=np.dtype('i4,i4,i4')).dtype
Out[62]: dtype([('f0', '<i4'), ('f1', '<i4'), ('f2', '<i4')])

In [63]: np.zeros(5, dtype=np.dtype('i4,i4,i4')).shape
Out[63]: (5,)

```

tacaswell: Ok, so `'(3,)i4'` and `3*'i4'` are definitely different datatypes, the question is if we expect the promotion (demotion?) to a int32 array with `shape + (3,)` or not.

https://docs.scipy.org/doc/numpy/user/basics.rec.html#structured-arrays

Sorry for the stream of conscious comments as I sort this out :sheep:

FrancescAlted: @tacaswell I remember discussing with Travis the 'demotion' of multidimensional dtypes in NumPy for atomic dtypes (i.e. not compound or structured) quite long time ago and IIRC while he agreed that the demotion was not consistent, he argued easiness of implementation to keep the current behaviour.  Perhaps the NumPy guys can change that, but IMO, this ship sailed long time ago so it is not a good idea to ask for that change now.

tables/core/array.py
```diff
@@ -136,8 +140,10 @@ def copy(self, newparent=None, newname=None,
             newparent = self.root._get_or_create_path(newparent, createparents)
         if self.__class__.__name__ == 'Array':
             create_function = newparent.create_array
-        else:
+        elif self.__class__.__name__ == 'CArray':
```
avalentino: Probably it would be saner to use isinstance to test the condition, but you need to reverse the order of tests because CArraiy *is* an Array and EArray *is* a CArray.

albertosm27: I tried, but with isinstance I need to import CArray in array.py and CArray also imports Array which makes it crash.

avalentino: @albertosm27 yes, you are right. Now I remember the issue :/

tables/tests/test_vlarray.py
```diff
@@ -1323,8 +1324,8 @@ def test05b_VLStringAtom(self):
         # Modify values
         vlarray[0] = "as4"
         vlarray[1] = "aaanc"
-        self.assertRaises(ValueError, vlarray.__setitem__, 1, "shrt")
-        self.assertRaises(ValueError, vlarray.__setitem__, 1, "toolong")
+        #self.assertRaises(ValueError, vlarray.__setitem__, 1, "shrt")
+        #self.assertRaises(ValueError, vlarray.__setitem__, 1, "toolong")
```
FrancescAlted: Make sure that you add new tests for the assignation of values that are of different length than the old values.

FrancescAlted: Also, do not forget to add an entry abut this new feature in the CHANGELOG for PyTables 4.0.

PR #640, opened 2017-07-18
Add PEP 519 support

mruffalo:
PEP 519 defines a filesystem path protocol, and enhances Python builtin I/O functions to support "path-like" objects such as `pathlib.Path`s. As part of this PEP, the `os` module also gained a `fspath` function, which takes an arbitrary path-like object and returns a string.
    
If the os.fspath function is available (i.e. Python >= 3.6), use it to convert `filename` to a string inside `utilsextension.encode_filename`. This allows using `Path` objects in `tables.open_file`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ tables/*.so
 
 # specific files
 src/version.h
+
+# Build artifacts
+a.out
+tmp/
--- a/tables/tests/test_basics.py
+++ b/tables/tests/test_basics.py
@@ -1987,6 +1987,51 @@ class UnicodeFilename(common.TempFileMixin, TestCase):
         group = self.h5file.create_group(root, 'vertex_data')
 
 
+@unittest.skipIf(sys.version_info < (3, 6),
+                 'PEP 519 was implemented in Python 3.6')
+class PathLikeFilename(common.TempFileMixin, TestCase):
+
+    def _getTempFileName(self):
+        from pathlib import Path
+        return Path(tempfile.mktemp(suffix='.h5'))
+
+    def setUp(self):
+        super(PathLikeFilename, self).setUp()
+
+        self.test = self.h5file.create_array('/', 'test', [1, 2])
+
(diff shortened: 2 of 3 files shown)


mruffalo:
Very strange; I have no idea what happened with Python 3.5:

```
FAIL: None (tables.tests.test_array.PointSelection4)
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/tables/tests/test_array.py", line 2119, in test02a_write
    "NumPy array and PyTables modifications does not match.")
AssertionError: False is not true : NumPy array and PyTables modifications does not match.
```

I didn't touch anything close to that test, and if I actually *had* somehow introduced an issue with basic indexing and data integrity, I find it very strange that it would only manifest on 3.5 and not 2.7, 3.4, or 3.6.

tomkooij:
This is a known bug in numpy 1.13. Will be fixed in numpy 1.13.1. The
failure is pretty random, but always in that test. Just disregard.

Thanks fort the PR. I cannot review atm, because i'm on mobile.

Op wo 19 jul. 2017 00:38 schreef Matt Ruffalo <notifications@github.com>:

mruffalo:
I was just reminded of this pull request -- apparently I wrote some code that depends on these changes, and didn't realize I'd done this until I tried to run that code on a system with a released version of PyTables.

Are there any problems with this PR before it could be merged?

tomkooij:
Sorry about not reviewing. I'll try to review this somewhere this week.

mruffalo:
Ping -- while modifying some of my analysis code that I haven't touched in a while, I also took this opportunity to save as HDF5 instead of pickling Pandas objects. I absentmindedly tried to pass a `Path` object to `pandas.HDFStore`, since all other Pandas I/O functions support path-like objects in addition to strings, and I momentarily forgot that paths still need to be wrapped in `str` since they're passed directly to PyTables.

tomkooij:
Merging. Thanks for contributing.

PR #642, opened 2017-08-22
No track times

alex-cobb:
As mentioned on the mailing list, these are changes to optionally disable recording of ctime (metadata creation time) when creating datasets.  This makes it possible to get bitwise identical output from repeated runs of the same application, which is useful for regression testing, and also for working with build tools that use hashes for change detection.

Commit 6eace4d does the real work, accepting a track_times flag on dataset constructors and then using H5Pset_obj_track_times() to optionally disable time tracking when creating datasets.

One of the commits here (19df430) adds a .track_times property to the Leaf class so that users can retrieve the value, but I am not sure this is actually a good idea.  The property uses H5Pget_obj_track_times() to get the value of this flag after retrieving the dataset creation property list, but H5Pget_obj_track_times() does not seem to give a correct value for the flag after the HDF5 file has been closed and reopened.  H5Py uses this function too, and has the same issue.  It might be better to just drop this property.   Users could simply check for a non-zero ctime for the dataset (using Node._get_obj_timestamps() from 69f6d41).  Dropping the .track_times property would mean reverting the changes in 19df430 and some minor changes to the tests created in 69f6d41.

### DIFF
--- a/src/H5ARRAY.c
+++ b/src/H5ARRAY.c
@@ -27,6 +27,7 @@
  * Date: March 19, 2001
  *
  * Comments: Modified by F. Alted. November 07, 2003
+ *           Modified by A. Cobb. August 21, 2017 (track_times)
  *
  *-------------------------------------------------------------------------
  */
@@ -44,6 +45,7 @@ hid_t H5ARRAYmake(  hid_t loc_id,
                     char  *complib,
                     int   shuffle,
                     int   fletcher32,
+		    hbool_t track_times,
                     const void *data)
 {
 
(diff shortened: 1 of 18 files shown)


tomkooij:
Nice feature! I like it.

I cannot review this right now, but I'll try to do this in the near future. (But please don't wait for me, I someone else is able to review).

alex-cobb:
FWIW, I asked on hdf-users about the issue of track_times not being recoverable after dataset close/reopen --- it is now with the HDF Helpdesk as issue HDFFV-10281.

FrancescAlted:
Yes, after a quick glance over it, I like this too.  My only comment is that there is an ongoing effort to put PyTables over h5py (see https://github.com/PyTables/PyTables/pull/634), and this new functionality would mean to re-base this work on top of h5py itself.  I am +1 on adding this now, but we should keep an eye on how to adapt this for the future h5py backend.

tacaswell:
h5py already supports turning off time-tracking so it should just be a bit of re-routing to make sure the option ends up in the right place.

FrancescAlted:
Ok.  Then I don't see any reason why we should not merge this.  This PR comes with tests, so this will easy that this functionality would be respected when the pt4 branch would be merged.

alex-cobb:
For a h5py backend it would mean passing the flag along to the h5py.File.create_dataset method: `this_file.create_dataset(..., track_times=False)`

PR #645, opened 2017-09-22
DOC small documentation fixes

rth:
A few small fixes in the documentation,
 * PEP8 in an example
 * fixed broken URL to the sourceforge mailing list archive
 * removed the "Old trac site" link as it is not working (but maybe there is a way to fix it)?


avalentino:
thanks @rth

Issue #649, opened 2017-09-30
Documentation error/is it really practical to inherit from Table?


molpopgen:
The [documentation](http://www.pytables.org/cookbook/custom_data_types.html) on using your own custom types inheriting from Table is incorrect.  It is outdated with respect to the current API and it has a few "quirks".  I've re-factored it to be up to date and a bit simpler:

```py
from __future__ import print_function
import numpy as np

import tables
from tables import File, Table
from tables.file import _checkfilters

from tables.parameters import EXPECTED_ROWS_TABLE


class DerivedFromTable(Table):
    _c_classId = 'DerivedFromTable'

    def __init__(self, parentNode, name, description=None,
                 title="", filters=None,
                 expectedrows=EXPECTED_ROWS_TABLE,
                 chunkshape=None, byteorder=None, _log=True):
        super(DerivedFromTable, self).__init__(parentNode, name,
                                          description=description, title=title,
                                          filters=filters,
                                          expectedrows=expectedrows,
                                          chunkshape=chunkshape, byteorder=byteorder,
                                          _log=_log)

    def read(self, start=None, stop=None, step=None, field=None):
        print("HERE!")
        data = Table.read(self, start=start, stop=stop, step=step,
                          field=field)
        return data


def createDerivedFromTable(self, where, name, data, title="",
                           filters=None, expectedrows=10000,
                           chunkshape=None, byteorder=None,
                           createparents=False):
    parentNode = self._get_or_create_path(where, createparents)

    _checkfilters(filters)
    return DerivedFromTable(parentNode, name, data,
                            title=title, filters=filters,
                            expectedrows=expectedrows,
                            chunkshape=chunkshape, byteorder=byteorder)


File.createDerivedFromTable = createDerivedFromTable


if __name__ == '__main__':
    x = np.array(np.random.rand(100))
    x=np.reshape(x,(50,2))
    x.dtype=[('x',np.float),('y',np.float)]
    h5file = tables.open_file('tester.hdf5', 'w')
    mtab = h5file.createDerivedFromTable(h5file.root, 'random', x)

    h5file.flush()
    print(type(mtab))
    mtab_read = mtab.read()
    h5file.close()
    h5file = tables.open_file('tester.hdf5', 'r')
    mtab = h5file.root.random

    print(type(mtab))
    mtab_read2 = mtab.read()
    print(np.array_equal(mtab_read, mtab_read2))
```

[...]

PR #650, opened 2017-10-05
Docs: cookbook: update inherit from Table

tomkooij:
Thanks to @molpopgen gh-649

### DIFF
--- a/doc/source/cookbook/custom_data_types.rst
+++ b/doc/source/cookbook/custom_data_types.rst
@@ -13,14 +13,12 @@ such as :class:`tables.Leaf`).
 This can be useful for storing a specialized type of data or presenting a
 customized API.
 
-Here's one way to do it, taken from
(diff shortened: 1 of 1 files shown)


molpopgen:
The PR has a bit of a problem.  It contains text describing the issue at the end (see last paragraph).  While the PR fixes one issue (the old example didn't work easily), it does not address the bigger issue, which is that it seems wrong to inherit from Table.

tomkooij:
I reopened the issue. The solution seems to be to documented as "do not do this", with the example showing why.

molpopgen:
Ok.  Is there a recommended way to support custom types?  There are allusions to inheriting from "isDescription" in the Table docs--is that related?  If so, is there an example?

tomkooij:
I'm moving this from this (closed) PR to the original issue, so we can track this #649

PR #663, opened 2018-01-08
fixing Deprecation Warning from numpy.issubdtype

oleksandr-pavlyk:
Changed `numpy.issubdtype(dt, int)` to `numpy.issubdtype(dt, np.signedinteger)` as suggested by the deprecation warning:

```text
In [2]: np.issubdtype(np.int32, int)
/localdisk/work/miniconda3/envs/np114/bin/ipython:1: FutureWarning: Conversion of the second argument of issubdtype from `int` to `np.signedinteger` is deprecated. In future, it will be treated as `np.int64 == np.dtype(int).type`.
  #!/localdisk/work/opavlyk/miniconda3/envs/np114/bin/python
Out[2]: True
```

### DIFF
--- a/tables/array.py
+++ b/tables/array.py
@@ -722,7 +722,7 @@ class Array(hdf5extension.Array, Leaf, six.Iterator):
         # truncate data if least_significant_digit filter is set
         # TODO: add the least_significant_digit attribute to the array on disk
         if (self.filters.least_significant_digit is not None and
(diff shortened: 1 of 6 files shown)


tomkooij:
Thanks! This (and your other PR) solves many (many!) DeprecationWarnings against numpy 1.14.

PR #664, opened 2018-01-08
Fixing about 200 test errors arising when upgrading to Numpy 1.14

oleksandr-pavlyk:
replaced hand-coded list of lists Python object inputs to `numpy.record.array` in test files `test_tables.py` and `test_tablesMD.py` to address failures with NumPy 1.14. 

See https://github.com/numpy/numpy/issues/10344 for more details. 

All the changes amount to replacing list of lists, specifying record array, with the list of tuples.

Also adjusted shape parameter of auto-generated record array in two instances, as that was off, and was causing test errors as well.


tomkooij:
Thanks! I'll try to review ASAP.

oleksandr-pavlyk:
@tomkooij To maybe help out with the review, let outline the changes I've made.

Let me state with a premise that NumPy's recommended way to input a vector of structures using Python objects is by [using a list of tuples](https://docs.scipy.org/doc/numpy-1.13.0/user/basics.rec.html), where each tuple corresponds to fields of the structure.

An effort is being made in `numpy.record.array` code to [recover](https://github.com/numpy/numpy/blob/master/numpy/core/records.py#L676) if the Python object input is given as a list of lists, rather than list of tuples.

Due to [a change](https://github.com/numpy/numpy/pull/6053) in NumPy, this recovery logic started to fail, and it is not clear whether one can always implement a robust recovery. It was suggested that input should be rewritten to conform to the recommended specs. 

This PR implements that directive, by changing inputs to `numpy.record.array` from list of lists to list of tuples, both where such inputs were specified in tests explicitly (see [example](https://github.com/PyTables/PyTables/blob/4a4d194430667be5de80c1ad62a34897900395d7/tables/tests/test_tables.py#L3003)) as well as where they were constructed programmatically (see [example](https://github.com/PyTables/PyTables/blob/4a4d194430667be5de80c1ad62a34897900395d7/tables/tests/test_tables.py#L5180)).

I likely have missed some, for [example](https://github.com/PyTables/PyTables/blob/4a4d194430667be5de80c1ad62a34897900395d7/tables/tests/test_tables.py#L204) the line 204, code should be changed from `buflist.append(tmplist)` to `buflist.append(tuple(tmplist))`.

I was guided by test failures though.

The other two changes I made were 

1. in [`test_tables.py#L146`](https://github.com/PyTables/PyTables/blob/4a4d194430667be5de80c1ad62a34897900395d7/tables/tests/test_tables.py#L146), where `range(self.expectedrows + 1)` was changed to `range(self.expectedrows)` to agree with value of keyword `shape=self.expectedrows` given on [L207](https://github.com/PyTables/PyTables/blob/4a4d194430667be5de80c1ad62a34897900395d7/tables/tests/test_tables.py#L207).

2. in [`test_tablesMD.py#L1088`](https://github.com/PyTables/PyTables/blob/4a4d194430667be5de80c1ad62a34897900395d7/tables/tests/test_tablesMD.py#L1088), where `shape=300` was corrected to be `shape=600`.

I hope this outline will help to get the changes in quicker.

oleksandr-pavlyk:
Ping.

tomkooij:
Merging. Thanks for contributing to PyTables!

PR #667, opened 2018-01-25
[FEAT] Internal C-Blosc sources updated to 1.13.3

FrancescAlted:

PR #668, opened 2018-01-26
[FEAT] Internal C-Blosc sources updated to 1.13.4

FrancescAlted:
This fixes a nasty bug introduced in C-Blosc 1.13 series as reported in:

https://github.com/Blosc/hdf5-blosc/issues/19

This overrides PR #667.


FrancescAlted:
Don't know why the appveryor tests failed for the branch, but they succeed on the PR, so I'd say that this is safe to be merged.

PR #672, opened 2018-02-23
[UPDATE] C-Blosc sources bumped to 1.14.0

FrancescAlted:
The 1.14.0 C-Blosc release is critical because it is the first one that enforces a forward compatibility policy for persistent compressed data.  More info in: http://blosc.org/posts/new-forward-compat-policy/


FrancescAlted:
All tests pass.  Merging.

Issue #651, opened 2017-10-18
Weird behavior of tables.expression.Expr when slice is outside of data samples


jackdbd:
I noticed a strange behavior when evaluating an expression on a slice which is outside of the data samples.

I would have expected to get an empty numpy array. Instead I got a single scalar. Is this a bug or the expected behavior?

Here is my environment:

```shell
Windows 10 64 bit
Python 2.7 32 bit
hdf5 1.8.18
tables 3.4.2
numexpr 2.6.4
numpy 1.13.1+mkl
```

I wrote these tests to show this issue.

```python
import tables
import unittest
import numpy as np


class TestNumexpr(unittest.TestCase):

    def test_slice(self):
        with tables.open_file("my_file.h5", mode="w", title="Test") as h5file:
            group = h5file.create_group(h5file.root, "columns", "Group Name")
            num_elements = 10
            arr_a = np.ones(num_elements)
            a = h5file.create_array(group, "a", arr_a, "Array a")
            arr_b = np.ones(num_elements) * 2.0
            b = h5file.create_array(group, "b", arr_b, "Array b")
            expr = tables.Expr("2 * a + b * 3")
            start, stop, step = (0, 7, None)
            expr.set_inputs_range(start, stop, step)
            result = expr.eval()
            expected = np.ones(stop) * 8
            self.assertListEqual(result.tolist(), expected.tolist())

    # This test fails: expr.eval() returns a scalar (8.0), not an empty np.array
    def test_empty_slice_for_binary_function_returns_empty_array(self):
        with tables.open_file("my_file.h5", mode="w", title="Test") as h5file:
            group = h5file.create_group(h5file.root, "columns", "Group Name")
            num_elements = 10
            arr_a = np.ones(num_elements)
            a = h5file.create_array(group, "a", arr_a, "Array a")
            arr_b = np.ones(num_elements) * 2.0
            b = h5file.create_array(group, "b", arr_b, "Array b")
            expr = tables.Expr("2 * a + b * 3")
            start, stop, step = (0, 0, None)
            # we get the same behavior if we pick a slice outside of the array
            # start, stop, step = (100, 1000, None)
            expr.set_inputs_range(start, stop, step)
            result = expr.eval()
            expected = np.array([])
            self.assertListEqual(result.tolist(), expected.tolist())

    # This test fails: expr.eval() returns a scalar (8.0), not an empty np.array
    def test_empty_slice_for_unary_function_returns_empty_array(self):
[...]

PR #674, opened 2018-03-14
Fix slicing outside of data samples

jackdbd:
Hopefully this fixes issue #651 

Actually I think this is more a workaround than a real fix...

I created the test `test02_out` in `test_expression.py`.

### DIFF
--- a/tables/tests/test_expression.py
+++ b/tables/tests/test_expression.py
@@ -149,6 +149,25 @@ class ExprTestCase(common.TempFileMixin, TestCase):
         self.assertTrue(common.areArraysEqual(r1, r2),
                         "Evaluate is returning a wrong value.")
 
+    def test02_out(self):
+        """Checking that expression is correctly evaluated when slice is
+        outside of data samples (`out` param)"""
+        expr = tables.Expr(self.expr, self.vars)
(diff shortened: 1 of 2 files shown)


tables/tests/test_expression.py
```diff
@@ -149,6 +149,25 @@ def test01_out(self):
         self.assertTrue(common.areArraysEqual(r1, r2),
                         "Evaluate is returning a wrong value.")
 
+    def test02_out(self):
+        """Checking that expression is correctly evaluated when slice is
+        outside of data samples (`out` param)"""
+        expr = tables.Expr(self.expr, self.vars)
+        # maybe it's better to use the leading dimemsion instead?
+        maxshape = max(self.shape)
+        start, stop, step = (maxshape + 1, maxshape + 2, None)
```
jsancho-gpl: This test is a good idea for checking outside ranges, but you are missing the current issue, when start and stop are identical. Maybe you could test it when start == stop too.

jsancho-gpl:
The issue appears not only when start and stop are the same, but when start is greater than stop too. I've changed it in your code. You are right, this is a workaround for the real problem, but you've done a good work anyway. Thanks a lot.

PR #677, opened 2018-03-18
Skip the DLL check when running from a frozen app

### LINKED ISSUE #675: Fix for #177 still appears broken for frozen apps
The relevant PyInstaller issue is here: https://github.com/pyinstaller/pyinstaller/issues/3407

I believe the original suggestion in #177 to skip the `find_library` checks in https://github.com/PyTables/PyTables/blob/develop/tables/__init__.py#L26-L86 would be better for frozen apps. Ideally, testing on an un-frozen version should have already established that the HDF5 library is correctly installed.

PyInstaller _is_ doing some monkey patching which interacts badly with PyTables, which I pointed out in the above linked issue. Again, just skipping the check when `sys.frozen` is `True` would bypass the problem.

jwiggins:
Fixes #675

### DIFF
--- a/tables/__init__.py
+++ b/tables/__init__.py
@@ -22,10 +22,13 @@ from __future__ import absolute_import
 
 
 import os
+import sys
+
 
 # On Windows, pre-load the HDF5 DLLs into the process via Ctypes
 # to improve diagnostics and avoid issues when loading DLLs during runtime.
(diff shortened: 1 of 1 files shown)


tomkooij:
@jwiggins : Thanks. Merging.

PR #684, opened 2018-04-22
Fix versionadded tags for v3.4.3

avalentino:
Flags for time tracking control have been added in v3.4.3, not in 3.4.


jsancho-gpl:
Good eye. I proceed with the merge immediately.

PR #685, opened 2018-04-25
Remove unused import and fix ImportError with numexpr 2.6.5.dev0

cgohlke:
`is_cpu_amd_intel` was recently removed from [`numexpr.necompiler`](https://github.com/pydata/numexpr/blob/master/numexpr/necompiler.py#L16)

### DIFF
--- a/tables/table.py
+++ b/tables/table.py
@@ -29,8 +29,7 @@ from . import tableextension
 from .lrucacheextension import ObjectCache, NumCache
 from .atom import Atom
 from .conditions import compile_condition
-from numexpr.necompiler import (
-    getType as numexpr_getType, double, is_cpu_amd_intel)
+from numexpr.necompiler import getType as numexpr_getType, double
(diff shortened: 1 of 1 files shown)


FrancescAlted:
It is unfortunate that `is_cpu_amd_intel` is gone from numexpr, although I agree that as it is not used in PyTables, it should go away (it is probably the remains of previously deleted code).  I suppose that this is forcing us to do another release of PyTables in order to be prepared for a forthcoming release of numexpr.  But frankly, I'd have preferred `is_cpu_amd_intel` not going away in numexpr.

FrancescAlted:
I just asked `is_cpu_amd_intel` to be back again in numexpr.  Please do not merge this until we have made a decision in numexpr.

robbmcleod:
Re-added a dummy variable `is_cpu_amd_intel=False` to NumExpr 2.6.5dev0 so you should be ok to accept this now.

Issue #688, opened 2018-05-02
Fault on Tests


MichaelOVertolli:
I linked a file with the dump.
[output.txt](https://github.com/PyTables/PyTables/files/1967809/output.txt)

MichaelOVertolli:
The system gets a segmentation fault if I try to run the PyTables from the master branch of the git.

PR #689, opened 2018-05-02
MAINT: Fixed test failures with Python 2.7 and NumPy 1.14.3

oleksandr-pavlyk:
While fixing fall-outs of changes to `numpy.rec.array`
(see https://github.com/numpy/numpy/issues/10344) I evidently missed
a spot which is now causing trouble. See

https://github.com/PyTables/PyTables/pull/664

I think I noted this in the aforementioned PR in a comment.

This PR is to make sure that `initRecArray` method produces list of
tuples, rather than list of lists, as an argument to `numpy.rec.array` constructor.

### DIFF
--- a/tables/tests/test_tables.py
+++ b/tables/tests/test_tables.py
@@ -201,7 +201,7 @@ class BasicTestCase(common.TempFileMixin, TestCase):
                 else:
                     tmplist.append(1 + float(i)*1j)
 
-            buflist.append(tmplist)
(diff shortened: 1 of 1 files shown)


scopatz:
LGTM, thanks @oleksandr-pavlyk

jsancho-gpl:
Thanks @oleksandr-pavlyk, this is directly related with #688 issue.

PR #730, opened 2019-03-22
Remove mock from requirements.txt and add to setup.py

### LINKED ISSUE #729: Mock dependency
Mock was recently added to requirements.txt as a dependency. Setup.py copies requirements.txt to install_requires thus making mock an install time requirement.

The commit shows, and searching the code verifies, that mock is only required for test: https://github.com/PyTables/PyTables/commit/b721f8b19f5e6e48a84bc80ba5c3fd8be95c6b93

Should this really be made an install time requirement? Wouldn't tests_require make more sense?

kylekeppler:
Fixes #729

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,3 @@
 numpy>=1.9.3
 numexpr>=2.6.2
 six>=1.9.0
-mock>=2.0
--- a/setup.py
+++ b/setup.py
@@ -1060,6 +1060,7 @@ interactively save and retrieve large amounts of data.
                 'sphinx_rtd_theme',
                 'numpydoc',
                 'ipython']},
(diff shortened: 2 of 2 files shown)


tomkooij:
Thanks for taking the time to submit a PR. However, I'm not sure if this is what's best for most users. (See my comment in #729)

kylekeppler:
No problem, I'd expect that to a pretty rare use case where you could just suggest the user to install mock. Main problem here is the conda-forge package has not picked up the mock dependency yet, reported in https://github.com/conda-forge/pytables-feedstock/issues/49.

tomkooij:
Superseded by conda-forge/pytables-feedstock#50

Issue #744, opened 2019-07-11
Python 3.8 ImportError: cannot import name 'clock' from 'time'


sergiopasra:
`time.clock` will be removed in Python 3.8 after being deprecated in Python 3.3

```
$ python3.7 -c 'from time import clock; clock()'
-c:1: DeprecationWarning: time.clock has been deprecated in Python 3.3 and will be removed from Python 3.8: use time.perf_counter or time.process_time instead
```
The recommendation is to change `clock` by either `time.perf_counter`or `time.process_time`

The differences between both functions are subtle, but it seems that `timeit`is using `time.perf_counter`, so I would use that function

See also: https://www.webucator.com/blog/2015/08/python-clocks-explained/

tomkooij:
Fixed in #745

pruthviraja786:
Thanks for the suggestion for the time module, "from time import perf_counter" works without any error. Try using perf_counter to count the time elapsed to execute the code.

PR #745, opened 2019-07-11
Replace deprecated time.clock with time.perf_counter

sergiopasra:
`time.clock` will be removed in Python 3.8 after being deprecated in Python 3.3

```
$ python3.7 -c 'from time import clock; clock()'
-c:1: DeprecationWarning: time.clock has been deprecated in Python 3.3 and will be removed from Python 3.8: use time.perf_counter or time.process_time instead
```

The Fedora linux distribution is starting to build Python packages with beta releases of Pyhton 3.8 in other to detect possible problems. This PR substitutes calls to `time.clock`, that will been removed in Python 3.8, with `time.perf_counter`

The related issue is #744

### DIFF
--- a/bench/bsddb-table-bench.py
+++ b/bench/bsddb-table-bench.py
@@ -239,16 +239,16 @@ if __name__ == "__main__":
     # Catch the hdf5 file passed as the last argument
     file = pargs[0]
 
-    t1 = time.clock()
+    t1 = time.perf_counter()
     psyco.bind(createFile)
     (rowsw, rowsz) = createFile(file, iterations, recsize, verbose)
-    t2 = time.clock()
+    t2 = time.perf_counter()
     tapprows = round(t2 - t1, 3)
 
-    t1 = time.clock()
+    t1 = time.perf_counter()
     psyco.bind(readFile)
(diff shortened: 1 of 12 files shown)


tirkarthi:
It has been removed in Python 3.8 with https://github.com/python/cpython/pull/13270 . Given that the project supports Python 2.7 one solution would be to try importing `time.perf_counter` and fallback to `time.clock` when not available.

tomkooij:
Thanks for the PR. Fixing this for py2.7 should be easy... (But we may want to drop py2.7 in the meantime...) We'll merge this soon!

scopatz:
Or maybe just use clock on Python 2 for the moment.

tomkooij:
> Or maybe just use clock on Python 2 for the moment.

Yes. But unfortunately that means a lot of conditional imports. Seems like a waste of time. I'm going to get rid of the python 2.7 CIs and merge this. (And maybe create a py27 branch or something)

tomkooij:
We decided to drop python 2.7 support (for an upcoming release that will support python 3.8). This can be merged into master even with the python 2.7 tests (obviously) failing.

PR #746, opened 2019-07-12
setup.py: remove improperly escaped regexps

keszybz:
python3.8 now warns about those. One solution would be to change each
of them to r'\s+' or '\\s+', but it turns out that using re is not
necessary at all, since the plain .split() does the exact same thing.


tomkooij:
Thanks!

PR #750, opened 2019-07-19
Use assertIsInstance in more tests

sethtroisi:
Replace all assertTrue(isinstance(...)) with assertIsInstance(...)
and also some assertFalse with assertNotIsIntance


scopatz:
Thanks a ton @sethtroisi!

PR #751, opened 2019-07-20
Cleanup more asserts

sethtroisi:
Thanks for pulling these changes, they are not the most glamorous but make debugging tests much easier


tomkooij:
Thanks again, @sethtroisi

PR #752, opened 2019-07-20
Final asserts changes

sethtroisi:
Note about half of these change functionality. I assume they were mistyped originally. 

self.assertTrue(type(b), numpy.ndarray)
=>
self.assertIsInstance(b, numpy.ndarray)

self.assertTrue(a1.maindim, self.maindim)	
=>
self.assertEqual(a1.maindim, self.maindim)


tomkooij:
I agree on the asserts! Thanks!

PR #755, opened 2019-08-12
Travis CI: test against python 3.8-dev

tomkooij:
Add python 3.8 dev to travis
Cleanup .travis.yml: Use xenial for all builds


avalentino:
Thanks @tomkooij

PR #767, opened 2019-10-02
setup.py fix egg/wheel on windows: correct dll loc

### LINKED ISSUE #766: WIN: Python 3.8rc0 : can't find hdf5.dll in egg
EDIT: No this was all wrong:
`python setup.py bdist_wheel` should included (vendored) DLLs in the wheel.

`python setup.py install | develop` should NOT include DLLs in egg/installation. 

This means from python 3.8 onwards, we cannot reliably use eggs on windows. I will update docs.

tomkooij:
The current `setup.py` works for (windows) wheels. (We use manylinux/auditwheel for Linux/Mac). But eggs are broken!

Specify *.dll in `package_data` so that wheels and eggs
include hdf5.dll in the correct folder. (site-packages\tables)
This is required for python 3.8 on windows due to stricter dll resolution.

TODO:
 - [x] Check that hdf5.dll is correctly included in wheels/eggs
 - [ ] Check wheelbuilding: will `bzip2.dll` or `blosc.dll` be included when building wheels at MacPython/pytables-wheels with this branch?
 - [ ] Fix `dll_files` var in `setup.py`: it is unused now, check that it can be removed altogether. (Mac/Linux?)
 - [ ] test `zlib.dll` for the wheelbuilder (conda `hdf5.dll` is dynamically linked against `zlib.dll`)
 - [ ] test `blosc.dll` for eggs and wheels

Fixes #766

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,7 +126,6 @@ if __name__ == '__main__':
     lib_dirs = []
     inc_dirs = [os.path.join('hdf5-blosc', 'src')]
     optional_libs = []
-    data_files = []    # list of data files to add to packages (mainly for DLL's)
 
     default_header_dirs = None
     default_library_dirs = None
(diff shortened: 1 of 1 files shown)


tomkooij:
Never mind.

`eggs` / `setuptools` drive me crazy with seemingly inconsistent behaviour. I might change docs to *always use wheels*. 

I can't get this to work. `python setup.py install` is probably broken for python>=3.8 on Windows. Because py3.8 will not search for DLLs in PATH, and copying the DLLs is hard. 
`python setup.py develop` will work when using the Anaconda patched python in a conda env, but not on plain python.

PR #781, opened 2019-12-04
Fix slow import of tables

cgohlke:
On my Windows workstation, `import tables` takes 3.3 s, about 3 s of which are spent in `import pkg_resources`.

Importing `pkg_resources` is known to be excruciatingly slow. See https://github.com/pypa/setuptools/issues/510

This PR moves the import of pkg_resources to the single test function where pkg_resources is used.

### DIFF
--- a/tables/nodes/tests/test_filenode.py
+++ b/tables/nodes/tests/test_filenode.py
@@ -18,8 +18,6 @@ import shutil
 import tempfile
 import warnings
 
-from pkg_resources import resource_filename
-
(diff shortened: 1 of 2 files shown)


avalentino:
Thanks @cgohlke.
The PR looks good to me but tests fail on windows.
I'm pretty sure that the problem is not related to changes in this PR. Need some more investigation.

tomkooij:
Windows / Appveyor is unrelated. Merged.

PR #783, opened 2019-12-27
Adding Travis-Ci Support For Arm64

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 job in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 arch:
     - amd64
-    - arm64
 language: python
 dist: xenial
 
@@ -9,6 +8,11 @@ python:
   - 3.6
   - 3.7
   - 3.8-dev
+
+jobs:
+  include:
+  - python: 3.7
+    arch: arm64
 
 addons:
   apt_packages:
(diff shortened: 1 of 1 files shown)


mohneesh9797-puresoftware:
Ping @tomkooij can you please have a look at this PR.

tomkooij:
LGTM. The appveyor error is unrelated. (Add to the relevant issue).

@mohneesh9797-puresoftware : Is there an actual use case for `arm64` (There are hdf5 packages in Arch Linux for arm64)

No problem to merge, but **why** do we want to test against this arch??? If it is used somewhere, I'm fine.

rhenwood-arm:
@tomkooij  - I have been working with @ossdev07 on libraries that are consumed in the Anaconda python distribution. This is one of them. Given there is a dependency on an underlying library (hdf5) external to python, it is valuable for AArch64 users that this library is tested routinely on AArch64.

Generally, I have observed that in some cases, AArch64 test results are slow to return (I assume this is because of missing wheels, that will hopefully be addressed soon by manylinux2014?). It might make sense to limit the scope of testing to one python version (3.7?) initially.

PR #792, opened 2020-02-20
Fix for Python3.9 on Debian

Marco-Sulla:
This way, the directories found by `pkg-config` have precedence over the other dirs. With Python3.9 on Debian, without this change the `setup.py` does not work.


scopatz:
This seems reasonable to me

tomkooij:
Yes. LGTM. I added it to RELEASE_NOTES.txt. Will merge.

PR #802, opened 2020-04-21
Fix light index optlevel > 6 chunksize

leolchat:
Using a light index with optlevel above 6 result in a float chunksize value, which gets pytables to crash soon after.
This fixes it.


FrancescAlted:
Well spotted.

PR #804, opened 2020-05-24
Update tutorials.rst

dearkarm:
array(pressure) is confusing as it implies array() is part of tables, which leads to "module not callable" error. Since the tutorial introduces numpy import as "import numpy", numpy.array() is preferred.


avalentino:
Thanks @dearkarm

PR #810, opened 2020-07-20
Solve numpy tostring DeprecationWarning

sroet:
Since `numpy=1.19` `array.tostring()` is deprecated and should be replaced with `array.tobytes()` ([numpy release note reference](https://numpy.org/devdocs/release/1.19.0-notes.html#numpy-ndarray-tostring-is-deprecated-in-favor-of-tobytes))

This PR changes all of them, it passes the small test suite. 

No bump of the minimal `numpy` version is needed, as `tobytes()` was implemented in `numpy>=1.9`, which is already the minimal requirement.

### DIFF
--- a/tables/atom.py
+++ b/tables/atom.py
@@ -1130,7 +1130,7 @@ class VLStringAtom(_BufferedAtom):
         return numpy.string_(object_)
 
     def fromarray(self, array):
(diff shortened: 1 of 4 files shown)


avalentino:
Thanks @sroet

PR #816, opened 2020-09-12
Update super usage

rahulporuri:
On Python 3, `super` doesn't need to called with arguments where as on Python 2, `super` needs to be called with a class object and an instance as the two arguments e.g. `super(EArray, self)`.

Now that the pytables codebase is python 3 only, the `super` usage can be updated to remove the need to pass args.

Note that these changes were done using automated regex-based search and replace. `(super)+\(+(\w+)(,)+\s(self)+\)+`. The regex searches for the word `"super"`, followed by the character `(`, followed by another word, followed by the comma character `,`, followed by the word `"self"`, followed by the character `)`.

After the automated search and replace was done, each change was individually checked and added to the commit. Each change was checked to ensure that the `super` was being called with the current class and not a super class. For example, in `CArray`, the usage `super(Array, self)` has not been updated. A few other such uses have been observed. I am not sure if they are intentional or if they are hidden bugs.

Note to reviewer : These changes were not prompted by any specific issue in the repository but something I noticed while I was going through the codebase. This is a large changeset and I fully understand if the maintainer/maintainers decide to not merge this PR/close this PR/brake it up into smaller chunks.

Ref : https://docs.python.org/3/library/functions.html#super

### DIFF
--- a/cpuinfo.py
+++ b/cpuinfo.py
@@ -65,7 +65,7 @@ if platform.system().lower() == 'windows':
 				# to get --onefile mode working.
 				os.putenv('_MEIPASS2', sys._MEIPASS)
 			try:
-				super(_Popen, self).__init__(*args, **kw)
+				super().__init__(*args, **kw)
 			finally:
 				if hasattr(sys, 'frozen'):
 					# On some platforms (e.g. AIX) 'os.unsetenv()' is not
--- a/doc/source/cookbook/custom_data_types.rst
+++ b/doc/source/cookbook/custom_data_types.rst
@@ -34,12 +34,12 @@ Submitted by Kevin R. Thornton.
                      title="", filters=None,
                      expectedrows=EXPECTED_ROWS_TABLE,
                      chunkshape=None, byteorder=None, _log=True):
-            super(DerivedFromTable, self).__init__(parentNode, name,
-                                              description=description, title=title,
-                                              filters=filters,
-                                              expectedrows=expectedrows,
-                                              chunkshape=chunkshape, byteorder=byteorder,
-                                              _log=_log)
+            super().__init__(parentNode, name,
+                             description=description, title=title,
+                             filters=filters,
+                             expectedrows=expectedrows,
+                             chunkshape=chunkshape, byteorder=byteorder,
+                             _log=_log)
 
         def read(self, start=None, stop=None, step=None, field=None):
             print("HERE!")
--- a/examples/multiprocess_access_benchmarks.py
+++ b/examples/multiprocess_access_benchmarks.py
@@ -40,7 +40,7 @@ def create_file(array_size):
 class PipeReceive(multiprocessing.Process):
 
     def __init__(self, receiver_pipe, result_send):
-        super(PipeReceive, self).__init__()
+        super().__init__()
         self.receiver_pipe = receiver_pipe
         self.result_send = result_send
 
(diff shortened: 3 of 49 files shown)


avalentino:
Hi @rahulporuri, thank you for the huge work.
Personally I'm all in favour of code modernization.
All tests seems to pass but I would like some other maintainer to commend on this before merging.

avalentino (approved):

rahulporuri:
> Hi @rahulporuri, thank you for the huge work.
Personally I'm all in favour of code modernization.
All tests seems to pass but I would like some other maintainer to commend on this before merging.

I fully understand. Thanks for the review.

FrancescAlted:
+1 I am onboard on this too!

FrancescAlted (approved):

rahulporuri:
@FrancescAlted @avalentino I noticed these when I was making the changes and I want to bring them to your attention. I didn't include them in this PR because they are byclass the superclass. (I am not at all familiar with the internals of pytables so these might be intentional).

The automated search and replace made these three changes which I chose to not include because they are off. They are off because

* we call `super(Array, self)` inside the class `CArray`. Ref : https://github.com/PyTables/PyTables/blob/dd0e7e5f4d75cc26ee2e717f1580996f4fb9456e/tables/carray.py#L219-L220
* we call `super(Array, self)` inside the classes `CacheArray` and `IndexArray`. Ref : https://github.com/PyTables/PyTables/blob/dd0e7e5f4d75cc26ee2e717f1580996f4fb9456e/tables/indexesextension.pyx#L586 and https://github.com/PyTables/PyTables/blob/dd0e7e5f4d75cc26ee2e717f1580996f4fb9456e/tables/indexesextension.pyx#L1489

<details><summary>git diff of changes which werent included</summary>

```
diff --git a/tables/carray.py b/tables/carray.py
index 887cdc07..4e392c88 100644
--- a/tables/carray.py
+++ b/tables/carray.py
@@ -216,7 +216,7 @@ class CArray(Array):
                 self._v_chunkshape = tuple(SizeType(s) for s in chunkshape)

         # The `Array` class is not abstract enough! :(
-        super(Array, self).__init__(parentnode, name, new, filters,
+        super().__init__(parentnode, name, new, filters,
                                     byteorder, _log, track_times)

     def _g_create(self):
diff --git a/tables/indexesextension.pyx b/tables/indexesextension.pyx
index b076cd3d..f9d8a47c 100644
--- a/tables/indexesextension.pyx
+++ b/tables/indexesextension.pyx
@@ -583,7 +583,7 @@ cdef class CacheArray(Array):
     return

   def _g_close(self):
-    super(Array, self)._g_close()
+    super()._g_close()
     # Release specific resources of this class
     if self.mem_space_id > 0:
       H5Sclose(self.mem_space_id)
@@ -1486,7 +1486,7 @@ cdef class IndexArray(Array):


   def _g_close(self):
-    super(Array, self)._g_close()
+    super()._g_close()
     # Release specific resources of this class
     if self.mem_space_id > 0:
       H5Sclose(self.mem_space_id)
```

</details>

avalentino:
@rahulporuri the call to `` super(Array, self).__init__(...)`` in ``carray.py`` is made on purpose (see also the comment on the previous line ``# The `Array` class is not abstract enough! :(``).

Regarding the two calls to ``super(Array, self)._g_close() `` in ``indexesextension.pyx``, IMHO that lines can be converted to ``super()._g_close()`` because ``Array`` don't seem to implement the ``_g_close`` method directly.
Also I tested that change an all seem to work.

In conclusion, please add a new commit with the updated ``super`` call for ``_g_close`` methods only (``indexesextension.pyx``).

avalentino:
Thanks @rahulporuri

PR #817, opened 2020-09-12
Remove utf8 coding cookies in source files

rahulporuri:
On Python 3, the default source file encoding is utf8 so the utf8 coding cookies in source files are not necessary.
Note that there are source files with latin-1 coding cookies which have not been touched in this commit.

Ref : https://docs.python.org/3/tutorial/interpreter.html#source-code-encoding


avalentino:
Not sure about this.
Is there any editor around that relies on coding cookies?

rahulporuri:
>  Not sure about this.
Is there any editor around that relies on coding cookies?

I don't know if there are any and why there would be. I don't see any issues with the source code in VS Code, PyCharm and vim but then again, that isn't really an answer to your question.

FrancescAlted:
I am +0 with this change. Probably removing latin1 coding cookies is fine too.

avalentino (approved):

avalentino:
thanks @rahulporuri

PR #818, opened 2020-09-13
Remove u string prefixes

rahulporuri:
On Python 2, `u"..."` string prefix is used to denote a unicode string where as on Python 3, all strings are unicode strings so the `u"..."` string prefix isnt needed. The changes were made by manually searching and replacing the occurances in the codebase.

### DIFF
--- a/tables/tests/test_vlarray.py
+++ b/tables/tests/test_vlarray.py
@@ -1288,7 +1288,7 @@ class TypesTestCase(common.TempFileMixin, TestCase):
         # Check for ticket #62.
         self.assertRaises(TypeError, vlarray.append, [b"foo", b"bar"])
         # `VLStringAtom` makes no encoding assumptions.  See ticket #51.
-        self.assertRaises(UnicodeEncodeError, vlarray.append, u"asd\xe4")
+        self.assertRaises(UnicodeEncodeError, vlarray.append, "asd\xe4")
 
         if self.reopen:
             name = vlarray._v_pathname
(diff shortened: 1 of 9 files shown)


tables/tests/test_vlarray.py
```diff
@@ -1359,7 +1359,7 @@ def test06a_Object(self):
 
         vlarray = self.h5file.create_vlarray(
             '/', "Object", atom=ObjectAtom())
-        vlarray.append([[1, 2, 3], "aaa", u"aaa���"])
+        vlarray.append([[1, 2, 3], "aaa", "aaa���"])
```
rahulporuri, line 1362: These are interesting - the file has the coding cookie - `# -*- coding: latin-1 -*-` and if I reload the file using the Latin-1 encoding in VS Code, these bytes are decoded correctly.

The surprising thing is that GitHub is able to decode the bytes before this change was made and I'm not sure how. If I load the file on master with encoding set to UTF-8. I just see garbage - not the accented c-like characters that GitHub is displaying.

rahulporuri: https://github.com/PyTables/PyTables/commit/8bbe84f91cf81e4228e68a4ffa7084592e3a3861

avalentino (approved):

avalentino:
LGTM

tables/tests/test_vlarray.py
FrancescAlted: Ah, yes.  I forgot that the internal supported coding for `VLArray` is just `latin-1`.  So better don't change anything more here.

rahulporuri: i'm not exactly sure what you mean by that @FrancescAlted . Do you mean not remove the coding cookie in the file?

rahulporuri:
> @rahulporuri can you please rebase against the current master?

Done.

tables/tests/test_vlarray.py
FrancescAlted: Yes, I mean exactly that.  It is a test file, so I don't think this would be problematic.

FrancescAlted (approved):

avalentino:
Thanks @rahulporuri

PR #829, opened 2020-11-16
doctest: adjust __init__ TypeError formatting

keszybz:
Python3.10 includes the class name in the TypeError string
for __init__(). This change allows the tests to pass in the new version,
but obviously breaks with older Pythons.

This patch is enough for the tests to pass in Fedora rawhide with the new python version, but obviously you need something that also works with python <= 3.9. So I'm opening this to get the discussion started ;)

### DIFF
--- a/tables/atom.py
+++ b/tables/atom.py
@@ -569,10 +569,10 @@ class Atom(metaclass=MetaAtom):
             >>> atom3 = atom1.copy(shape=(2, 2))
             >>> print(atom3)
             Int32Atom(shape=(2, 2), dflt=0)
-            >>> atom1.copy(foobar=42)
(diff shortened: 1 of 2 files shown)


avalentino:
Hi @keszybz sorry for the late reply, and thanks for your PR.
Honestly I'm not sure how the problem can be solved keeping compatibility with all python versions.
Probably we should just skip those specific tests (``# doctest: +SKIP``)

keszybz:
It took me a while... but here's a new version. The tests pass with python3.9 and python3.10.

avalentino (approved):

avalentino:
@keszybz very nice indeed.
Thanks a lot, merging ...

Issue #823, opened 2020-10-09
release wheels for python 3.9


xmatthias:
Currently, [pypi](https://pypi.org/project/tables/#files) does not contain wheels for python 3.9.

I know python 3.9 just came out - but it would be great if the last release (or a new release - considering the last release has been over a year ago) could be updated to include Wheels for python 3.9 to simplify installation.

amotl:
Dear @avalentino, @tomkooij, @PyTables and @numfocus,

first things first: Thank you so much for conceiving and maintaining PyTables.

Binary wheels for Python 3.9 for all of lxml, h5py, Pandas and Numpy are already available on PyPI these days. At @earthobservations, we just have been able to upgrade to the most recent releases on [Wetterdienst](https://github.com/earthobservations/wetterdienst) and it works like a charm, see https://github.com/earthobservations/wetterdienst/pull/202#issuecomment-727190718.

It would be awesome if you could also build and upload `cp39` artefacts to PyPI to make it easier for downstream packages and their users to upgrade to Python 3.9.

With kind regards,
Andreas.

avalentino:
OK i managed to build wheels for linux locally.
@tomkooij  if you do not have plans to build wheels via [pytables-wheels](https://github.com/MacPython/pytables-wheels) I could try to upload mines.

amotl:
Dear Antonio,

thanks a bunch already. While I am just getting started to learn about the package publishing infrastructure behind the different Python packages for scientific computing, there is nothing that I could really contribute here other than pointing out what others are doing, especially in the light of switching away from Rackspace and staging packages through either AppVeyor/Travis/Anaconda or GitHub Actions these days.

In this manner, the discussion over at https://github.com/pydata/numexpr/issues/367 has progressed and you might want to look into what @robbmcleod and @mattip (thank you!) have been doing / are suggesting there.

Specifically, [cibuildwheel](https://github.com/joerick/cibuildwheel) by @joerick, @YannickJadoul and contributors seems to be a nice way to build wheels these days and supports the full stack of Python versions and CI platforms.

With kind regards,
Andreas.

avalentino:
Dear Andreas,
thank you for the pointers. I also used cibuildwheel to build wheels locally. I agree with you, it is a very nice tool.
I have just uploaded wheels for GNU/Linux.
Unfortunately I do not have too much time in this moment to setup a CI job to generate and upload wheels automatically.
[...]

PR #833, opened 2020-12-11
Add Python 3.9 support

eumiro:
Add Python 3.9 (and 3.8) to appveyor and travis CIs.

Did you consider dropping Python 3.5 support (since after EOL)?

And in https://github.com/PyTables/PyTables/blob/master/setup.py#L98-L103 the minimum required version is 3.4. Is this still valid as CIs are testing only against Py3.5+?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ python:
   - 3.5
   - 3.6
   - 3.7
-  - 3.8-dev
+  - 3.8
+  - 3.9
 
 jobs:
   include:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,6 +21,14 @@ environment:
       PYTHON_VERSION: "3.7"
       PYTHON_ARCH: "64"
 
+    - PYTHON: "C:\\Miniconda38-x64"
+      PYTHON_VERSION: "3.8"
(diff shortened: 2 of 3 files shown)


avalentino (approved):

avalentino:
Dear @eumiro thanks for your effort. Unfortunately windows builds are still not working.
I'm keen t merge this PR as soon as the issue with windows builds is fixed.

Anyway I would like to inform you the there is in the air the idea of switching to Git Hub Actions (GHA), see also #823.
It is still not happened (and I don't if and when it will) mainly due to lack of developer time.
The main reason of the switch is that travis is becoming super-slow and probably soon it will be no longer available at all.

Would you be interested in submitting a PR implementing CI fpr PyTables in GHA?

eumiro:
@avalentino this will require some assistance from you since as a unprivileged PR contributor I cannot test GHA if there is no GHA activated in the `master` branch. You can see whether #834 merged into `master` runs the tests and then we'll have to fix possible bugs and add the `deploy` section as well.

avalentino:
@eumiro great! having GHA would be fantastic. Thank you very much.
I will comment on #834 for thehnical details.

avalentino:
Closing.
Superseded by #834.

PR #834, opened 2020-12-12
Add GitHub Actions for Tests

eumiro:
Convert the Test CI from Travis to GitHub Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
@@ -31,3 +31,14 @@ jobs:
     - name: Source distibution
       run: |
         python setup.py sdist
+    - name: Installation
+      run: |
+        pip install dist/*.tar.gz
+    - name: 'Run test'
+      run: |
+        cd .. && python -m tables.tests.test_all
+        pt2to3 -h
+        ptrepack -h
+        ptdump -h
+        pttree -h
+


avalentino:
OK, we cannot see GHA tests until GHA itself is not activated in the repo.
I suggest to merge a first PR with a dummy GHA setup, e.g. just checkout sources (more or less what you already have).
At that point we will have the  GHA enabled and we can setup a new PR with the complete implementation.
Does it make sense for you?

cc @amotl

eumiro:
The GHA results are [here](https://github.com/eumiro/PyTables/actions) but I am learning the syntax on the fly, so it still does not work correctly. Anyway we don't need to merge to `master` right now. If you have any ideas, how to make it work, let me know.

eumiro:
FYI, @avalentino I found a YAML formatting error, fixed it, and now it runs apparently for more than 20 minutes (it still did not appear in the list of builds), which sounds like it should do something now. If that fails again, I will limit the list of Python versions to a single one to fail faster.

Oh, there's a gitter.im chat if you want to follow…

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/checkout@v2
+      - name: 'Set up Python ${{ matrix.python-version }}'
+        uses: actions/setup-python@v2
+        with:
+          python-version: '${{ matrix.python-version }}'
+      # - run: sudo apt-get install libbz2-dev libhdf5-serial-dev liblzo2-dev
+      - name: 'Run test'
+        run: |
+          pip install cython
+          pip install dist/*.tar.gz
```
avalentino, line 28: IMHO before running
```
pip install dist/*.tar.gz
```
there should be something like
```
python3 setup.py sdist
```
or
```
make dist
```

avalentino (changes requested):

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+          - ubuntu-latest
+        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
+    steps:
+      - name: 'Check out repo content'
+        uses: actions/checkout@v2
+      - name: 'Set up Python ${{ matrix.python-version }}'
+        uses: actions/setup-python@v2
+        with:
+          python-version: '${{ matrix.python-version }}'
+      # - run: sudo apt-get install libbz2-dev libhdf5-serial-dev liblzo2-dev
```
avalentino, line 24: Without this PyTables is not going to build

avalentino (changes requested):

.github/workflows/build.yml
eumiro: The long run is still now over. I will now change that line to `python3 setup.py dist` and limit Python versions to 3.5.

avalentino:
> If that fails again, I will limit the list of Python versions to a single one to fail faster.

I agree, it would accelerate a lot things for the initial development.

.github/workflows/build.yml
eumiro: Thanks, I added that line to the long list of `run` in `Run test`.

eumiro:
Did I break GitHub? Even after more than an hour there is no response and adding new workflow files, even very basic, does not trigger any new builds. :man_shrugging:

avalentino:
> Did I break GitHub? Even after more than an hour there is no response and adding new workflow files, even very basic, does not trigger any new builds.

Indeed there is something strange.
Please let go the CI on this dummy workflow and I will merge it.
After that we shall be able to see GHA stuff also in the PR and on the main repo.

avalentino:
@eumiro  I can confirm that also on my personal fork I'm not able to start GHA, there must be some problem on GH side.
I will try again later.

eumiro:
> > Did I break GitHub? Even after more than an hour there is no response and adding new workflow files, even very basic, does not trigger any new builds.
> 
> Indeed there is something strange.
> Please let go the CI on this dummy workflow and I will merge it.
> After that we shall be able to see GHA stuff also in the PR and on the main repo.

If you don't want to pollute `master` with our attempts, create a new branch on the main repository and we'll define the GHA against this branch. Then we can develop the GHA configuration and PR it against it as well. When everything works, we'll repeat it at once with `master`.

I'll wait now until there's a working dummy GHA on the main repository.

eumiro:
> @eumiro OK, please rebase on the current master

Rebased and I updated the `ci.yml` file. There is still a problem with the tests, but at least they get launched, even in a PR!

.github/workflows/ci.yml
```diff
@@ -1,6 +1,12 @@
 name: CI
 
-on: [push]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
avalentino, line 9: Not sure but I think that this one is the cause of GHA not triggered on PR (not matching "master").
Probably using the following:
```
on: [push pull_request]
```
is OK.
After all we do not have an high commit rate.

avalentino (changes requested):

avalentino (approved):

avalentino:
@eumiro when you are fine with your changes please remember to update the PR title and description.

avalentino:
Fantastic!
Thanks a lot @eumiro

PR #837, opened 2020-12-16
Out of order datatypes in Dataset #836

### LINKED ISSUE #836: Out of order datatypes in Dataset
I was working with .h5 files when I noticed some weird behaviour. There is no requirement (I believe!) for the datatypes in a dataset to be in order. For instance, running `h5ls -rfv` on one of these out-of-order h5 files:

```
Storage:   30 logical bytes, 65520 allocated bytes, 0.05% utilization       
Type:      struct {                                                         
               "test_5"           +25   5-byte null-terminated ASCII string 
               "test_10"          +15   10-byte null-terminated ASCII string
               "test_15"          +0    15-byte null-terminated ASCII string
           } 30 bytes                                                       
```

So the data is actually stored in rows as test_15, test_10 and test_5 but added to the datatypes out of order. Some other tools that I use correctly display these types of .h5 files. PyTables doesn't error and gives erroneous/missing data when viewing.

Attaching a .h5 if anyone would like to verify.

`get_node("/group/table")[0]`
Gives:
`(b'*****', b'---------', b'....\x00%@\x00\x00\x00\x00\x00\x001@')`

When what is actually stored is:
`(b'**************', b'---------', b'....')`

I've been working locally with a small change to get around this issue. Not sure if it's correct or the most efficient as I don't usually use PyTables. I'll submit a PR with a test incase it is helpful!

[out_of_order_types.h5.zip](https://github.com/PyTables/PyTables/files/5705853/out_of_order_types.h5.zip)

ste-pool:
Fixes #836


avalentino (approved):

avalentino:
Looks good to me, tanks @ste-pool

FrancescAlted:
Good work @ste-pool !

PR #838, opened 2020-12-18
Apply `pyupgrade --py36-plus` to the Python code

eumiro:
This commit is done only automatically using  the following command:

```bash
pyupgrade --py36-plus {bench,c-blosc,contrib,doc,examples,tables}/**/*.py cpuinfo.py setup.py
```

Since we're at Python 3.6+, the code can be modernized using the automatic tool https://github.com/asottile/pyupgrade using the `--py36-plus` switch, which does the following:

- remove the old `from __future__` imports
- convert string formattings from `%` to `.format` to f-string depending on what is shorter
- declare classes without the unnecessary `(object)` inheritance
- remove the unnecessary `# -*- coding: UTF-8 -*-` for source files
- `yield from`
- literal dicts
- shorted `super()`
- etc.

### DIFF
--- a/c-blosc/conanfile.py
+++ b/c-blosc/conanfile.py
@@ -49,7 +49,7 @@ class CbloscConan(ConanFile):
             else:
                 return
             with tools.environment_append({'CTEST_OUTPUT_ON_FAILURE': '1'}):
-                self.run("%s ctest %s" % (prefix, test_args))
+                self.run(f"{prefix} ctest {test_args}")
 
     def package(self):
         self.copy("blosc.h", dst="include", src="blosc")
--- a/c-blosc/tests/check_symbols.py
+++ b/c-blosc/tests/check_symbols.py
@@ -1,6 +1,5 @@
 """Checks the exported symbols."""
 
-from __future__ import print_function, unicode_literals
 
 import argparse
 import re
@@ -61,7 +60,7 @@ def check_symbols(path):
       del symbol_type
       if re.match(allowed_symbols_pattern, name):
         continue
-      print('ERROR: Found unexpected symbol in %s: %s' % (filename, name))
+      print(f'ERROR: Found unexpected symbol in {filename}: {name}')
       failed = True
   return not failed
 
--- a/cpuinfo.py
+++ b/cpuinfo.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: UTF-8 -*-
 
 # Copyright (c) 2014-2018, Matthew Brennan Jones <matthew.brennan.jones@gmail.com>
 # Py-cpuinfo gets CPU info with pure Python 2 & 3
@@ -79,7 +78,7 @@ if platform.system().lower() == 'windows':
 
 	forking.Popen = _Popen
 
-class DataSource(object):
+class DataSource:
 	bits = platform.architecture()[0]
 	cpu_count = multiprocessing.cpu_count()
 	is_windows = platform.system().lower() == 'windows'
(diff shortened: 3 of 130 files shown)


c-blosc/conanfile.py
```diff
@@ -49,7 +49,7 @@ def launch_tests(self):
             else:
                 return
             with tools.environment_append({'CTEST_OUTPUT_ON_FAILURE': '1'}):
-                self.run("%s ctest %s" % (prefix, test_args))
+                self.run(f"{prefix} ctest {test_args}")
```
avalentino, line 52: Files in c-blosc are imported from an external repo, we should not touch them.

c-blosc/tests/check_symbols.py
```diff
@@ -1,6 +1,5 @@
 """Checks the exported symbols."""
 
-from __future__ import print_function, unicode_literals
```
avalentino, line 3: Imported from an external repo (see above)

cpuinfo.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: UTF-8 -*-
```
avalentino, line 2: Also cpuinfo.py is an external tool vendor in pytables. Please do not modify it.

tables/tests/test_utils.py
```diff
@@ -4,7 +4,7 @@
 try:
     from unittest.mock import patch
 except ImportError:
-    from mock import patch
+    from unittest.mock import patch
```
avalentino, line 8: Here probably the entire try/except shall be removed, and just leave
```
from unittest.mock import patch
```

tables/tests/test_vlarray.py
```diff
@@ -1,5 +1,3 @@
-# -*- coding: latin-1 -*-
-
 import sys
```
avalentino, line 1: Please revert this.
I have had a lot of problems in the past for that coding cookie :/

avalentino (changes requested):

avalentino:
A general comment: with this PR we are explicitly breaking compatibility with python 3.5, I'm not against it but 
a. I would like to have some feedback from at lease another @PyTables/maintainers
b. we need to document it and update setup.py, classifier etc.

Finally, as noted in my review comments, there are some files/directories that are imported form external repositories and shall not be modified IMHO.

eumiro:
This should probably start as an issue to discuss and plan and become a PR only when the time has arrived. It is important to prepare it well, but then running the command is very cheap.

1. I'll remove the `c-blocs` directory and the `cpuinfo.py` file (which had tabs instead of spaces, which irritated me) from the call
2. After you discuss it with the maintainers, there should be some last version for Python 3.5.

tables/tests/test_vlarray.py
eumiro: Wouldn't it be safer to replace those 17 occurrences of the `REPLACEMENT CHARACTER` with `\ufffd` or `\N{REPLACEMENT CHARACTER}` or a variable?

tables/tests/test_utils.py
eumiro: That should be done in a different PR.

eumiro:
I have created a less aggressive PR #839 upgrading the code (without the imported sections) to Python 3 only and thus keeping the 3.5 support. I have also fixed the encoding issue in the `test_vlarray.py` file. Now this PR can be closed.

tables/tests/test_vlarray.py
eumiro: WIP in #839

avalentino:
Thanks a lot for your effort.
Just curious, apart for the excluded files, which is he delta w.r..t #839? Only f-strings?

eumiro:
> Just curious, apart for the excluded files, which is he delta w.r..t #839? Only f-strings?

Yes. I applied `--py36-plus` to the `--py3-plus`ified code and all changes were only f-strings.

PR #839, opened 2020-12-18
Apply `pyupgrade --py3-plus` to the Python codebase

eumiro:
Modernize the whole Python codebase (except for imported `c-blosc` and `cpuinfo.py`) using the automated command:

```bash
pyupgrade --py3-plus {bench,contrib,doc,examples,tables}/**/*.py setup.py
```

This should have happened before upgrading everything to Python 3.6+ in #838 and thus dropping the Python 3.5 support. Now only old Python 2 code is deleted or converted. Dropping Python 3.5 support should be planned and announced.

The second commit fixes the issue in the file `tables/tests/test_vlarray.py`.
`pyupgrade` automatically removed the `# -*- coding: latin-1 -*-` header which was needed to interpret correctly 17 occurrences of the `REPLACEMENT CHARACTER` (or `\ufffd`). Now this character has been replaced within the Python string with `\ufffd` and interpreted correctly.

### DIFF
--- a/tables/nodes/tests/test_filenode.py
+++ b/tables/nodes/tests/test_filenode.py
@@ -9,7 +9,6 @@
 ########################################################################
 
 """Unit test for the filenode module."""
-from __future__ import absolute_import
 
 import os
 import shutil
@@ -404,7 +403,7 @@ class ReadFileTestCase(TempFileMixin, TestCase):
         except ImportError:
             # PIL not available, nothing to do.
             pass
-        except IOError:
+        except OSError:
             self.fail(
                 "PIL was not able to create an image from the file node.")
 
--- a/tables/nodes/tests/test_filenode.py
+++ b/tables/nodes/tests/test_filenode.py
@@ -374,15 +374,9 @@ class ReadFileTestCase(TempFileMixin, TestCase):
     def test00_CompareFile(self):
         """Reading and comparing a whole file node."""
 
-        # Try to use hashlib (included from Python 2.5 on)
-        try:
-            import hashlib
-            dfiledigest = hashlib.md5(self.datafile.read()).digest()
-            fnodedigest = hashlib.md5(self.fnode.read()).digest()
-        except ImportError:
-            import md5
-            dfiledigest = md5.new(self.datafile.read()).digest()
-            fnodedigest = md5.new(self.fnode.read()).digest()
+        import hashlib
+        dfiledigest = hashlib.md5(self.datafile.read()).digest()
+        fnodedigest = hashlib.md5(self.fnode.read()).digest()
 
         self.assertEqual(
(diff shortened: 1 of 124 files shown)


eumiro:
Sorry, I have tried that `test_vlarray` cookie again:

```python
# -*- coding: latin-1        
                             
a = '�'                      
b = '\ufffd'                 
c = '\xef\xbf\xbd'           
print(a == b, b == c, c == a)
```
and only `a == c`. That's why the `REPLACEMENT CHARACTER` has to be replaced by a sequence of three latin-1 bytes that come from:
```python
print(''.join(f'\\x{ord(x):02x}' for x in '�'.encode().decode('latin1')))
```

tables/tests/test_utils.py
```diff
@@ -4,7 +4,7 @@
 try:
     from unittest.mock import patch
 except ImportError:
-    from mock import patch
+    from unittest.mock import patch
```
avalentino, line 7: Forgot this?
Please just replace:
```
try:
    from unittest.mock import patch
except ImportError:
    from mock import patch
```
with
```
from unittest.mock import patch
```

avalentino (changes requested):

avalentino:
Many thanks @eumiro 
The PR looks very good, I have just requested a small change.

tables/tests/test_utils.py
eumiro: Right, thanks. There were two other cases of now-useless `ImportError` hacks. Fixed all of them.

tables/nodes/tests/test_filenode.py
```diff
@@ -375,15 +374,9 @@ def tearDown(self):
...
-            import hashlib
-            dfiledigest = hashlib.md5(self.datafile.read()).digest()
-            fnodedigest = hashlib.md5(self.fnode.read()).digest()
-        except ImportError:
-            import md5
-            dfiledigest = md5.new(self.datafile.read()).digest()
-            fnodedigest = md5.new(self.fnode.read()).digest()
+        import hashlib
+        dfiledigest = hashlib.md5(self.datafile.read()).digest()
+        fnodedigest = hashlib.md5(self.fnode.read()).digest()
```
avalentino, line 380: Good catch! Thanks

tables/scripts/ptrepack.py
```diff
@@ -14,17 +14,12 @@
...
 import sys
 import time
 import os.path
 import argparse
 import warnings
-try:
-    from time import process_time as cputime
-except ImportError:
-    from time import clock as cputime
+from time import process_time as cputime
```
avalentino, line 22: +1

avalentino (approved):

avalentino:
Very good job.
Thanks @eumiro

PR #844, opened 2020-12-23
[DO-NOT-MERGE] Update gzguts.h

sroet:
This ports over the fix from https://github.com/Blosc/c-blosc/commit/5fc66f7cdcff969b47332ed843be5f344a5543cc and https://github.com/Blosc/c-blosc/commit/406f5960da86f0932bd3609dae0bc70d5c785919 . 

Which is how #828 was resolved for the main blosc repo (https://github.com/Blosc/python-blosc/issues/229). As mentioned in that issue, all credit goes to @xytxytxyt and @FrancescAlted.

### DIFF
--- a/c-blosc/internal-complibs/zlib-1.2.8/gzguts.h
+++ b/c-blosc/internal-complibs/zlib-1.2.8/gzguts.h
@@ -2,6 +2,9 @@
  * Copyright (C) 2004, 2005, 2010, 2011, 2012, 2013 Mark Adler
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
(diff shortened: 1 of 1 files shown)


FrancescAlted:
I think a better approach is to update to latest c-blosc sources so as to fix this.  I'll try to find time for doing this in the next few days.

sroet:
@FrancescAlted  +1 for that approach. I will keep this open until that has happened as a reminder (mainly for me to update `git` install instead of a `pip` install), but feel free to close this whenever you want.

FrancescAlted:
Overridden by #847

PR #846, opened 2020-12-28
A bunch of minor code quality refactorings

eumiro:
This is a bunch of independent code-improvement refactorings. Please feel free to discuss/accept/reject any number of them. I'd like to know your opinion on them.

1. Remove `.keys()` calls on dicts when the keys are needed, since the dict provides them implicitly
2. Change loops with `yield` to `yield from`
3. Bugfix: `if … in ('instancemethod')`. That's not a tuple and therefore it tested against the string `in 'instancemethod'`
4. `"""Docstrings use triple double quotes."""`
5. Remove one try/except with a method changed in Python 3.5
6. Simplify chained conditions
7. classmethods are usually called on `cls` variables
8. Change a mutable (list) default argument to a tuple and refactor the method
9. Nested repeating `for i in …` replaced with more meaningful variable names to avoid confusion
10. refactor `get_nrows` to have less code to read
11. `isinstance(obj, (cls1, cls2, …))` accepts multiple classes
12. simplify range loops to get the variables immediately and not through index

### DIFF
--- a/tables/tests/test_basics.py
+++ b/tables/tests/test_basics.py
@@ -2423,8 +2423,7 @@ class TestDescription(TestCase):
         descr = Description(d)
         self.assertEqual(sorted(descr._v_names), sorted(d.keys()))
         self.assertIsInstance(descr._v_dtype, numpy.dtype)
-        self.assertTrue(sorted(descr._v_dtype.fields.keys()),
-                        sorted(d.keys()))
+        self.assertTrue(sorted(descr._v_dtype.fields), sorted(d.keys()))
 
 
 class TestAtom(TestCase):
--- a/bench/indexed_search.py
+++ b/bench/indexed_search.py
@@ -28,13 +28,10 @@ prec = 6  # precision for printing floats purposes
 
 
 def get_nrows(nrows_str):
-    if nrows_str.endswith("k"):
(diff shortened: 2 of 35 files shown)


tables/tests/test_basics.py
```diff
@@ -2421,10 +2421,9 @@ def test_str_names(self):
         # see gh-42
         d = {'name': tables.Int16Col()}
         descr = Description(d)
-        self.assertEqual(sorted(descr._v_names), sorted(d.keys()))
+        self.assertEqual(sorted(descr._v_names), sorted(d))
```
avalentino, line 2424: Maybe in this specific case using ``.keys()`` is more easy to understand for the reader.

avalentino:
LGTM, thanks a lot @eumiro

tables/tests/test_basics.py
eumiro: Okay, I will revert that.

avalentino (approved):

avalentino:
Thank you very much @eumiro.
You work is very appreciated.
Would you be interested in becoming a PyTables maintainer?
In the case please feel free to contact me in private (antonio dot valentino at tiscali dot it) or on the PyTables mailing list.

PR #847, opened 2020-12-29
Update to latest C-Blosc sources (1.21.0)

FrancescAlted:
This should override #844.  Note that Snappy codec is not distributed anymore in C-Blosc sources (maintaining its C++ sources was becoming pretty difficult).  However, it should be possible to link with Snappy if the library is installed in the system (I think conda does distribute a package for Snappy).  This is mentioned in the RELEASE_NOTES.txt.


avalentino (approved):

avalentino:
Thanks @FrancescAlted

PR #849, opened 2020-12-30
Doc code and typos

eumiro:
Fixed some typos in the doc. Modernized the example code snippets.

Then generated everything with `make html`, copied to the `pytables.github.com` repo, but the diff is much larger than expected due to some IE/class/… changes. The new version is also `3.6.2.dev0`, so probably we should run this update when the new release is ready.

### DIFF
--- a/doc/scripts/pickletrouble.py
+++ b/doc/scripts/pickletrouble.py
@@ -4,24 +4,24 @@ import tables
 class MyClass:
     foo = 'bar'
 
+
 # An object of my custom class.
 myObject = MyClass()
 
-h5f = tables.open_file('test.h5', 'w')
-h5f.root._v_attrs.obj = myObject  # store the object
-print(h5f.root._v_attrs.obj.foo)  # retrieve it
-h5f.close()
+with tables.open_file('test.h5', 'w') as h5f:
+    h5f.root._v_attrs.obj = myObject  # store the object
+    print(h5f.root._v_attrs.obj.foo)  # retrieve it
 
 # Delete class of stored object and reopen the file.
 del MyClass, myObject
 
-h5f = tables.open_file('test.h5', 'r')
-print(h5f.root._v_attrs.obj.foo)
-# Let us inspect the object to see what is happening.
-print(repr(h5f.root._v_attrs.obj))
-# Maybe unpickling the string will yield more information:
-import pickle
-pickle.loads(h5f.root._v_attrs.obj)
-# So the problem was not in the stored object,
-# but in the *environment* where it was restored.
-h5f.close()
+with tables.open_file('test.h5', 'r') as h5f:
+    print(h5f.root._v_attrs.obj.foo)
+    # Let us inspect the object to see what is happening.
+    print(repr(h5f.root._v_attrs.obj))
+    # Maybe unpickling the string will yield more information:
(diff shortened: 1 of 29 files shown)


doc/scripts/pickletrouble.py
```diff
@@ -4,24 +4,25 @@
...
+    print(h5f.root._v_attrs.obj.foo)
+    # Let us inspect the object to see what is happening.
+    print(repr(h5f.root._v_attrs.obj))
+    # Maybe unpickling the string will yield more information:
+    import pickle
+
+    pickle.loads(h5f.root._v_attrs.obj)
+    # So the problem was not in the stored object,
+    # but in the *environment* where it was restored.
+    h5f.close()
```
avalentino, line 28: no need to close if one uses the with statement

doc/source/usersguide/tutorials.rst
```diff
@@ -1227,18 +1227,18 @@ can be passed to this method::
...
 
             ########### Detectable errors start here. Play with them!
-            particle['pressure'] = array(i*arange(2*3)).reshape((2,4))  # Incorrect
-            #particle['pressure'] = array(i*arange(2*3)).reshape((2,3)) # Correct
+            particle['pressure'] = np.array(i * np.arange(2 * 3)).reshape((2, 4))  # Incorrect
+            #particle['pressure'] = np.array(i * np.arange(2 * 3)).reshape((2, 3)) # Correct
             ########### End of errors
 
-            particle['temperature'] = (i**2)     # Broadcasting
+            particle['temperature'] = i ** 2     # Broadcasting
```
avalentino, line 1242: Personally I'm not used to see spaces around the ``**`` operator but it is OK

avalentino (approved):
Just a couple of minor comments

avalentino:
> The new version is also 3.6.2.dev0, so probably we should run this update when the new release is ready.

Probably yes but in the end there are not too many differences between 3.6.1. and 3.6.2.dev0 so also updating now would be fine to me.

doc/source/usersguide/tutorials.rst
eumiro: Would we consider introducing some sort of automatic formatter for the code, such as `black`? I believe `black` puts spaces there, but with a formatter (and a `pre-commit` rule that would apply all those changes to any commit) there would be no need to take special care of it.

doc/scripts/pickletrouble.py
eumiro: Thank you, fixed and repushed.

doc/source/usersguide/tutorials.rst
avalentino: Honestly I'm not a big fan of pre-commit hooks the modify the code.
I have always checked the code with [flake8](https://flake8.pycqa.org) if we want to switch to black I'm fine with it but I would prefer to to use this tools offline.

eumiro: They are meant to check whether the code has been formatted/linted properly and ask the contributor to do so. `flake8` is one of the options, of course.

eumiro:
Pushed to https://github.com/PyTables/pytables.github.com and the new version is live. Thank you!

PR #850, opened 2020-12-30
Drop Python 3.5 support, use f-strings and underscores in integers 

### LINKED ISSUE #840: Dropping Python 3.5 support
Python 3.5 is after end-of-life since 2020-09-30. Dropping its support would allow introduction of f-strings and keep the development modern. In #839 the codebase has already been `pyupgrade`d to Python 3 and some old Python 2 import hacks have been removed.

To abandon Python 3.5 properly, one of the next versions of PyTables should announce the deprecation of Python 3.5.

eumiro:
This drops Python 3.5 support from CIs and PyPI metadata. PyTables will be then Python 3.6+ and this introduces following new syntax features: 

- PEP 498, formatted string literals.
- PEP 515, underscores in numeric literals.
- PEP 526, syntax for variable annotations.
- PEP 525, asynchronous generators.
- PEP 530: asynchronous comprehensions.

The first two can be added automatically using `pyupgrade` and manually. Variable annotations should be discussed. Async stuff is not relevant for PyTables yet.

Closes #840

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         # os: [ubuntu-latest, macos-latest, windows-latest]
         os: [ubuntu-latest]
-        python: ['3.5', '3.6', '3.7', '3.8', '3.9']
+        python: ['3.6', '3.7', '3.8', '3.9']
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python }}
--- a/bench/chunkshape-bench.py
+++ b/bench/chunkshape-bench.py
@@ -44,7 +44,7 @@ b = a.copy(f.root, "b", chunkshape=newchunkshape)
 tcpy = round(time() - t1, 3)
 thcpy = round(dim1 * dim2 * 8 / (tcpy * 1024 * 1024), 1)
 print("Chunkshape for row-wise chunkshape array:", b.chunkshape)
-print("Time to copy the original array: {} sec ({} MB/s)".format(tcpy, thcpy))
+print(f"Time to copy the original array: {tcpy} sec ({thcpy} MB/s)")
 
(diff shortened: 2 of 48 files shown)


avalentino (approved):
LGTM.
Thanks @eumiro

avalentino:
Ups, I haven't realized this is still in draft, sorry

eumiro:
Sorry, @avalentino these were the two first steps. The third one (manually converting the rest of the string formatting to f-strings) is under way, but of course takes a little bit more time (and the review will probably too be more work). But I will already merge this and do the rest in a new PR.

eumiro:
@avalentino the https://github.com/PyTables/PyTables/runs/1629625410 failed because it wanted to `Building cp35-manylinux_x86_64 wheel` and we are now beyond 3.5 with the f-strings.

In the file https://github.com/PyTables/PyTables/blob/master/.github/workflows/wheels.yml there's no mention of 3.5, but there is the line 40: `CIBW_SKIP: "cp27-* pp*"`. Is it possible, that `cp35` is on some list and we have to skip it here?

avalentino:
> In the file https://github.com/PyTables/PyTables/blob/master/.github/workflows/wheels.yml there's no mention of 3.5, but there is the line 40: CIBW_SKIP: "cp27-* pp*". Is it possible, that cp35 is on some list and we have to skip it here?

@eumiro  yes, correct.
Adding "cp55-*" to CIBW_SKIP should be enough.

Sorry the workflow for building wheels is only on master and releases because it takes very long time to run

eumiro:
Oh yes, the list is here: https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip

I replaced `CIBW_SKIP` with `CIBW_BUILD` to have an explicit list of versions to support. A future us will have easier time to add/drop Python versions.

PR #851, opened 2021-01-01
Replace round() with proper f-string

eumiro:
Most calls to `round()` are used to prepare the values to format into a string. This can be replaced by proper `.Xf` format in an f-string.

The amount of potential new f-strings in the code is very high and I have recognized some patterns that could be refactored later. I'll add new code features in similar batches.


avalentino (approved):
LGTM

eumiro:
Thank you, @avalentino. I spotted a tiny typo in one format definition, fixed and repushed it. Now I'll merge it.

PR #853, opened 2021-01-02
Add itersorted test for step=-1 and fix #740

### LINKED ISSUE #740: PyTables does not return the first row when using itersorted with a negative step
'm using the itersorted method in PyTables. All rows are returned correctly when the step is set to a positive value, but the first row is not returned when using a negative step value. I asked about this on StackOverflow and one reply suggests it could be a bug.

Below is test code that illustrates the problem:

	import tables

        class DBEntry(tables.IsDescription):
		id = tables.Int16Col()

	class Test(object):

		def __init__(self, parent = None):
			pass

		def create_db_entries(self):
			db = tables.open_file("py_db_test.h5", mode = "w", title = "db")
			table = db.create_table('/', 'test', DBEntry, "test")
			entry = table.row
			for i in range(0, 5):
				entry['id'] = i
				entry.append()

			table.flush()
			table.cols.id.create_csindex()
			db.close()

		def get_db_entries(self):
			db = tables.open_file("py_db_test.h5", mode = "r")
			entries = db.root.test
			print('ascending (all rows showing)')
			for row in entries.itersorted('id', start = None, stop = None, step = 1):
				print(row['id'])
			print('decending (missing first row)')
			for row in entries.itersorted('id', start = None, stop = None, step = -1):
				print(row['id'])
			db.close()

	if __name__ == '__main__':
		test = Test()
		test.create_db_entries()
		test.get_db_entries()`

eumiro:
The current test `test04_itersorted9` tested with `start=None, stop=None, step=-5`, so it did not always have to return the first row. Add a new `test04_itersorted10` test with `start=None, stop=None, step=-1` and fixed a line in `tableextensions.pyx` to include that last missing row.

Fixes #740

### DIFF
--- a/tables/tableextension.pyx
+++ b/tables/tableextension.pyx
@@ -1050,7 +1050,7 @@ cdef class Row:
         self._finish_riterator()
     elif 0 > self.step:
       #print("self.nextelement = ", self.nextelement, self.start, self.nrowsread, self.nextelement <  self.start - self.nrowsread + 1)
-      while self.nextelement - 1 > self.stop:
+      while self.nextelement > self.stop:
         if self.nextelement < self.start - (<long long> self.nrowsread) + 1:
           if 0 > self.nextelement - (<long long> self.nrowsinbuf) + 1:
             tmp = self.coords[0:self.nextelement + 1]
--- a/tables/tests/test_indexes.py
+++ b/tables/tests/test_indexes.py
@@ -2062,7 +2062,7 @@ class CompletelySortedIndexTestCase(TempFileMixin, TestCase):
         self.assertTrue(allequal(sortedtable, sortedtable2))
 
     def test04_itersorted9(self):
-        """Testing the Table.itersorted() method with a negative step."""
+        """Testing the Table.itersorted() method with a negative step -5."""
 
         # see also gh-252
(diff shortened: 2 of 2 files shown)


avalentino (approved):

avalentino:
Thanks @eumiro, very good catch!

eumiro:
Thank you, @avalentino . This was my very first Cython bugfix, and the only way to test I found today was to put `print`s and `python setup.py develop` each time. What would be the best way to do it with less overhead?

And what is your opinion on our labels? I had difficult time choosing the right label for this PR.

avalentino:
> Thank you, @avalentino . This was my very first Cython bugfix, and the only way to test I found today was to put prints and python setup.py develop each time. What would be the best way to do it with less overhead?

I also use ``print`` most of the times (with ``python3 setup.py build_ext --inplace`` which should be equivalent to your method).
But usually I do not debug complex things in cython.
Anyway if you are interested the official documentation is here [1].

[1] https://cython.readthedocs.io/en/latest/src/userguide/debugging.html

> And what is your opinion on our labels? I had difficult time choosing the right label for this PR.

For this PR I would have used "defect"

PR #855, opened 2021-01-05
Refactor `__str__` and `__repr__` with f-strings

eumiro:
This adds some new f-strings to the codebase, refactoring all `__str__` and `__repr__` methods. Here I have a question about https://github.com/PyTables/PyTables/blob/master/tables/file.py#L2788 (the `Last modif.` line in the `str(h5f)` output), that is formatted as `repr(time.asctime(…))` which looks like `'Mon Sep 20 12:40:47 2004'`. Would be an `isoformat()` alternative (e.g. `2021-01-05 12:34:56+00:00`) more appropriate here? This is the only place where we print/format some datetime info.

The bonus tiny commit changes one percentage variable to a `ratio` variable (0.–1. instead of 0–100) and formats it accordingly. The

### DIFF
--- a/tables/attributeset.py
+++ b/tables/attributeset.py
@@ -675,21 +675,17 @@ be ready to see PyTables asking for *lots* of memory and possibly slow I/O"""
         # Get this class name
         classname = self.__class__.__name__
         # The attribute names
-        attrnumber = len([n for n in self._v_attrnames])
-        return "%s._v_attrs (%s), %s attributes" % \
-               (pathname, classname, attrnumber)
+        attrnumber = sum(1 for _ in self._v_attrnames)
+        return f"{pathname}._v_attrs ({classname}), {attrnumber} attributes"
 
     def __repr__(self):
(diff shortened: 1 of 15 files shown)


avalentino (approved):
LGTM

avalentino:
> Here I have a question about https://github.com/PyTables/PyTables/blob/master/tables/file.py#L2788 (the Last modif. line in the str(h5f) output), that is formatted as repr(time.asctime(…)) which looks like 'Mon Sep 20 12:40:47 2004'. Would be an isoformat() alternative (e.g. 2021-01-05 12:34:56+00:00) more appropriate here? This is the only place where we print/format some datetime info.

yes, IMHO using isoformat is a good idea.
Thanks @eumiro

eumiro:
Thank you, @avalentino . Please approve the third commit and when the tests are green, I'll merge it. (and thanks for the labels, I'll remember next time)

avalentino (approved):

PR #856, opened 2021-01-06
Consistent elapsed time measuring both real and cpu

eumiro:
Consistent import and usage of elapsed time measuring tools in the whole repository:
- `time.perf_counter() as clock()` for elapsed time
- `time.process_time() as cpuclock()` for CPU time

Both methods return `float` that are relative to some point within the process, so they should be used only for calculating deltas and not as UNIX timestamp. They are immune to system clock updates (such as NTP), and therefore the deltas are always positive.

There was previously a bug that used `time.perf_counter()` for CPU time. This has been fixed here.

### DIFF
--- a/bench/LRU-experiments.py
+++ b/bench/LRU-experiments.py
@@ -1,7 +1,7 @@
 # Testbed to perform experiments in order to determine best values for
 # the node numbers in LRU cache. Tables version.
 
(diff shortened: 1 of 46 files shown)


avalentino (approved):
LGTM

PR #857, opened 2021-01-06
[WIP/draft] Modernize setup.py

eumiro:
This is an attempt to modernize `setup.py` (f-strings, pathlib, formatting, …), but it does a little bit more, a bugfix and cleanup of old code. It has not been completely tested, so **PLEASE DO NOT MERGE.**

During the editing I changed quite a lot of stuff, then reformatted with `black`, so the diff was huge. Then I took the `setup.py` from the `master`, applied `black -l 79 setup.py && isort setup.py`, reformatted some comments, and rebased my changes on that, so after the first commit the file's functionality didn't change and the diff in the second commit is what I'd like to discuss with you.

1. `pathlib.Path` for file/directory manipulation. I have a relatively good feeling about the`posix` part, but little idea about the `nt` part. Anyway, there's the project's `ROOT` Path object where all project's files are based on. The `glob`bing and file reading is more compact. The main reason is to distinguish between strings and directory/file names. There were too many string hacks and now it should be clearer where we are working with file system. 
2. f-strings because the earlier `pyupgrade --py36-plus` didn't change everything, so I am updating anything I touch
3. fix typo `extras_required` in the `setup()` call
4. and entered a few `TODO`s where I was not sure what to do

I am not sure how `setup.py` should generally work, but this one prints quite a lot. So even for `python setup.py --version` where one would expect a single like with the version, it always prints some *debugging* info, such as directories with headers.

I am looking forward to your feedback!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,27 +4,26 @@
 
 import ctypes
 import distutils.spawn
-import glob
 import os
-import platform
 import subprocess
 import sys
 import tempfile
 import textwrap
 from distutils.ccompiler import new_compiler
-from distutils.core import Extension
 from distutils.dep_util import newer
 from distutils.util import convert_path
 from distutils.version import LooseVersion
-from os.path import exists, expanduser
+from pathlib import Path
 
 import pkg_resources
-# Using ``setuptools`` enables lots of goodies
-from setuptools import find_packages, setup
+from setuptools import find_packages, setup, Extension
+
 # We need to avoid importing numpy until we can be sure it's installed
 # This approach is based on this SO answer http://stackoverflow.com/a/21621689
 # This is also what pandas does.
 from setuptools.command.build_ext import build_ext
+
+ROOT = Path(__file__).resolve().parent
 
 if __name__ == "__main__":
     # `cpuinfo.py` uses multiprocessing to check CPUID flags. On Windows, the
@@ -35,16 +34,15 @@ if __name__ == "__main__":
 
         cpu_info = cpuinfo.get_cpu_info()
         cpu_flags = cpu_info["flags"]
-    except Exception as e:
-        print("cpuinfo failed, assuming no CPU features:", e)
+    except Exception as exc:
+        print("cpuinfo failed, assuming no CPU features:", exc)
         cpu_flags = []
 
     # The name for the pkg-config utility
     PKG_CONFIG = "pkg-config"
 
     # Fetch the requisites
-    with open("requirements.txt") as f:
-        requirements = f.read().splitlines()
+    requirements = (ROOT / "requirements.txt").read_text().splitlines()
 
     class BuildExtensions(build_ext):
         """Subclass setuptools build_ext command
@@ -81,7 +79,6 @@ if __name__ == "__main__":
         tw = textwrap.TextWrapper(
             initial_indent="   ", subsequent_indent="   "
         )
-
         print(f".. {kind.upper()}:: {head}")
         for line in tw.wrap(body):
             print(line)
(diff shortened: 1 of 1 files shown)


eumiro:
Sorry for the delay and CI build e-mails flood. There were some issues I couldn't identify on my host.

One of them was `error: each element of 'ext_modules' option must be an Extension instance or 2-tuple`. Thanks to 
https://stackoverflow.com/questions/21594925/error-each-element-of-ext-modules-option-must-be-an-extension-instance-or-2-t I changed the import of `Extension` and now it works. Appveyor helped me to fix some issues on the `nt` host.

The `Path` conversion could be still optimized. The idea is to convert from `str` to `Path` as soon as possible and from `Path` to `str` as late as possible. And maybe those libraries could start accepting `os.PathLike`.

setup.py
```diff
@@ -2,204 +2,207 @@
...
         default_header_dirs = []  # no default, must be given explicitly
         default_library_dirs = []  # no default, must be given explicitly
         default_runtime_dirs = [  # look for DLL files in ``%PATH%``
-            _path for _path in os.environ['PATH'].split(';')]
+            Path(_path) for _path in os.environ["PATH"].split(";")
+        ]
         # Add the \Windows\system to the runtime list (necessary for Vista)
-        default_runtime_dirs.append('\\windows\\system')
+        # TODO: should that go to C:\ ?
+        default_runtime_dirs.append(Path("\\windows\\system"))
```
avalentino, line 174: I'm not a windows user.
If I remember well I have seen someone with windows installed on ``D:\\``.
Probably the safer option is to stay with the current implementation.

setup.py
```diff
@@ -2,204 +2,207 @@
...
-        default_runtime_dirs.extend(
-            [os.path.join(sys.prefix, 'Lib\\site-packages\\tables')])
-
+        default_runtime_dirs.append(
+            Path(sys.prefix) / "Lib" / "site-packages" / "tables"
+        )
 
+    # TODO: still an issue?
     # Gcc 4.0.1 on Mac OS X 10.4 does not seem to include the default
     # header and library paths.  See ticket #18.
```
avalentino, line 182: Mac OS X 10.4 has been released in 2005, and clang is the default compiler since 2011 if I'm not wrong.
Anyway I can't say if that hack is still needed for all reasonably modern versions of OS X/Xcode.
I can only test Catalina if you want.

setup.py
```diff
@@ -318,555 +340,566 @@ def find_directories(self, location, use_pkgconfig=False):
                     # then, the directory will be returned as '\stuff' (!!)
                     # F. Alted 2006-02-16
                     if idx == 0:
-                        directories[idx] = os.path.dirname(path[:path.rfind(name)])
+                        # TODO: tweak this to make the above comment obsolete
+                        directories[idx] = Path(
+                            str(path)[: str(path).rfind(name)]
+                        )
```
avalentino, line 346: +1
Looks safe

setup.py
```diff
@@ -318,555 +340,566 @@ def find_directories(self, location, use_pkgconfig=False):
...
-            if not exists(extcfile) or newer(extpfile, extcfile):
+            if not extcfile.exists() or newer(extpfile, extcfile):
                 # This is the only place where Cython is needed, but every
                 # developer should have it installed, so it should not be
                 # a hard requisite
                 from Cython.Build import cythonize
-                cythonize(extpfile, language_level = "2")
+
+                # TODO: language_level="3" ?
+                cythonize(str(extpfile), language_level="2")
```
avalentino, line 803: Recently someone told me that he has been able to successfully build PyTables with language_level="3".
Anyway I think that changing"language_level" should be done in a separate PR and carefully tested.
Also we probably need to bump the minimum Cython version required.

avalentino (commented):
General comment: we should be careful when we make changes to the setup.py.
It includes a lot of hacks do address several corner cases and specific behaviours  of different systems, ad, most of all, we do not have a reliable suite of tests for it.

eumiro:
> General comment: we should be careful when we make changes to the setup.py.
> It includes a lot of hacks do address several corner cases and specific behaviours of different systems, ad, most of all, we do not have a reliable suite of tests for it.

Thank you for your feedback and I'll turn this PR back to draft to make sure it does not get merged by accident, because it should not happen that way.

However I feel we should adopt some changes in order not to leave it as a piece of code that should never be touched. Because if we do not want to touch it now, it will never be easier in the future. I have never seen (and investigated) such a big `setup.py` file, so I want to understand it fully and split it into pieces if needed.

Although the usage of `pathlib` involves converting from/to `str` (as some libraries *still* do not support `pathlib` objects), its usage would clearly show the variables that mean some files on the system. As I am not sure about the Windows path system (my last time with that OS was ten years before I started with Python), I'd leave this to someone else. But that means, that we have a problem, because we have some piece of Windows-related code we do not fully understand.

If you agree, I'd open another PR with individual steps, starting with `black&isort`, going on with typo fixes and f-strings, all changes that i am pretty confident about, but will prepare the soil for further actions that we should discuss.

setup.py
eumiro: As Windows does not have one root dir and this looks like “go up up up and then to windows on this drive” although `PyTables` could be installed somewhere else, I have a bad feeling about this line. I do not want to change it because I do not understand it. But I want to understand it. :-)

setup.py
eumiro: Never used Mac, no idea. Just asking.

setup.py
eumiro: This is really a string hack, but it should be translatable to `Path` operations as well. I think I have understood the comments, but it should be clear without them as well.

setup.py
eumiro: I agree with separate PR and tests.

avalentino:
@eumiro to me this PR can be merged as is (non need to make separate PRs IMHO) because it mainly contains cosmetic changes.
My comment was mainly referred to more substantial changes that one could think about to improve the current situation.
I'm not saying we should not touch setup.py, I'm just saying we should do it carefully, and IMHO the process could take time.
Your approach to start splitting pieces ad re-factorize into something more consistent and robust is the correct approach IMHO.

avalentino (approved):

eumiro:
Are you sure about the approval, @avalentino? If both of us are not sure about the Mac/Win changes, we should mark them as `TODO` and leave the review to someone who can test it and maybe has some further improvement ideas. As I wrote, I am preparing another PR with individual steps (commits) and only with changes I am pretty confident about.

avalentino:
> Are you sure about the approval, @avalentino? If both of us are not sure about the Mac/Win changes, ...

In both cases the old behaviour is maintained if I understand correctly. Is t not the case?

eumiro:
I couldn't test it and CI was satisfied, but I'd like to have the changes separated. Wait for the new PR, please.

eumiro:
A more structured improvement is being done in #858 .
Closing this PR.

PR #858, opened 2021-01-07
First steps into modern `setup.py`

eumiro:
These are the first steps to #857:

1. fix two string typos and one `setup` parameter name `extras_require` that always raised an error until now
2. apply `black -l 79` to `setup.py`. No `isort` this time, because of the import problems of `Extension`. Will fix that later.
3. reformat strings and comments to fit into 79 chars
4. transform all strings to f-strings
5. use `pathlib.Path` to read/write project files. The `.glob` part for `inc_dirs` and `blosc_sources` comes later

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,8 +53,8 @@ if __name__ == "__main__":
 
         BuildExtensions does two things
         1) it makes sure numpy is available
-        2) it injects numpy's core/include directory in the include_dirs parameter
-        of all extensions
+        2) it injects numpy's core/include directory in the include_dirs
+           parameter of all extensions
         3) it runs the original build_ext command
         """
 
@@ -226,7 +226,8 @@ if __name__ == "__main__":
             returns None otherwise
             """
             # An explicit path can not be provided for runtime libraries.
-            # (The argument is accepted for compatibility with previous methods.)
+            # (The argument is accepted for compatibility with previous
+            # methods.)
 
             # dlopen() won't tell us where the file is, just whether
             # success occurred, so this returns True instead of a filename
@@ -365,7 +366,8 @@ if __name__ == "__main__":
 
         def find_runtime_path(self, locations=default_runtime_dirs):
             # An explicit path can not be provided for runtime libraries.
-            # (The argument is accepted for compatibility with previous methods.)
+            # (The argument is accepted for compatibility with previous
+            # methods.)
             return _find_file_path(
                 self.runtime_name,
                 default_runtime_dirs,
@@ -449,8 +451,8 @@ if __name__ == "__main__":
 
         if os.environ.get("HDF5_USE_PREFIX", None):
             # This is used on CI systems to link against HDF5 library
-            # The vendored `hdf5.dll` in a wheel is renamed to: `pytables_hdf5.dll`
-            # This should prevent DLL Hell.
+            # The vendored `hdf5.dll` in a wheel is renamed to:
+            # `pytables_hdf5.dll`  This should prevent DLL Hell.
             print(
                 "* HDF5_USE_PREFIX: Trying to build against pytables_hdf5.dll"
             )
@@ -639,7 +641,8 @@ if __name__ == "__main__":
         if not location and CONDA_PREFIX:
             location = CONDA_PREFIX
 
-        # looking for lzo/lzo1x.h but pkgconfig already returns '/usr/include/lzo'
+        # looking for lzo/lzo1x.h but pkgconfig already returns
+        # '/usr/include/lzo'
         use_pkgconfig = USE_PKGCONFIG if package.tag != "LZO2" else False
 
         (hdrdir, libdir, rundir) = package.find_directories(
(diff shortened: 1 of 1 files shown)


avalentino (approved):

eumiro:
@avalentino thank you for the review. I updated one or two tiny details in previous commits and added four more. We're now more or less on the state of #857 with some more stuff I learned under way.

6. Use `pathlib.Path` to collect files for `setup`
7. Consolidate versions check
8. Extract common arguments of `Extension`s
9. Simplify/normalize/beautify the code

setup.py
```diff
@@ -589,284 +596,300 @@ def get_blosc_version(headername):
...
 
         (hdrdir, libdir, rundir) = package.find_directories(
-            location, use_pkgconfig=use_pkgconfig)
+            location, use_pkgconfig=use_pkgconfig
+        )
 
         # check if HDF5 library uses old DLL naming scheme
-        if hdrdir and package.tag == 'HDF5':
-            hdf5_header = os.path.join(hdrdir, "H5public.h")
+        if hdrdir is not None and hdrdir.is_dir() and package.tag == "HDF5":
```
avalentino, line 627: Please check this (and the following ones).
If I remember correctly there is the possibility that ``hdrdir`` is ``True`` (see the implementation of [find_directories]( https://github.com/PyTables/PyTables/blob/cd26d6d88870a47e0231abb53105e08d292e3843/setup.py#L321)).

In this case
```
if hdrdir is not None and hdrdir.is_dir() and ...
```
is not equivalent to
```
if hdrdir and ...
``` 

Unfortunately I do not remember why this hack has been implemented.

setup.py
```diff
@@ -589,284 +596,300 @@ def get_blosc_version(headername):
...
-        if package.tag not in ['HDF5']:
+        if package.tag not in ["HDF5"]:
             # Keep record of the optional libraries found.
             optional_libs.append(package.tag)
-            def_macros.append(('HAVE_%s_LIB' % package.tag, 1))
+            def_macros.append((f"HAVE_{package.tag}_LIB", 1))
 
-        if hdrdir and package.tag == 'BLOSC':
-            blosc_header = os.path.join(hdrdir, "blosc.h")
+        if hdrdir is not None and hdrdir.is_dir() and package.tag == "BLOSC":
```
avalentino, line 702: See the comment above

setup.py
```diff
@@ -589,284 +596,300 @@ def get_blosc_version(headername):
...
 
         # check if the library is in the standard compiler paths
         if not libdir and package.target_function:
-            libdir = compiler.has_function(package.target_function,
-                                        libraries=(package.library_name,))
+            libdir = compiler.has_function(
+                package.target_function, libraries=(package.library_name,)
+            )
 
         if not (hdrdir and libdir):
```
avalentino, line 652: if ``hdrdir`` and ``libdirs`` are ``Path``s this is always ``True``, please check

setup.py
```diff
@@ -589,284 +596,300 @@ def get_blosc_version(headername):
...
 
             continue  # look for the next library
 
         if libdir in ("", True):
-            print("* Found %s headers at ``%s``, the library is located in the "
-                "standard system search dirs." % (package.name, hdrdir))
+            print(
+                f"* Found {package.name} headers at ``{hdrdir}``, the library "
+                f"is located in the standard system search dirs."
+            )
```
avalentino, line 680: OK ``True`` means that the compiler found the library somewhere in the system.

setup.py
```diff
@@ -589,284 +596,300 @@ def get_blosc_version(headername):
...
-            print("* Found %s headers at ``%s``, library at ``%s``."
-                % (package.name, hdrdir, libdir))
+            print(
+                f"* Found {package.name} headers at ``{hdrdir}``, "
+                f"library at ``{libdir}``."
+            )
 
         if hdrdir not in default_header_dirs:
             inc_dirs.append(hdrdir)  # save header directory if needed
         if libdir not in default_library_dirs and libdir not in ("", True):
```
avalentino, line 689: Here again ``True`` is explicitly checked

avalentino (changes requested):

setup.py
eumiro: So what do we do in the example above? `if hdrdir` with `True/str` is true if `True` or the `str` is not empty. But if it is only `True`, we cannot `os.path.join` it to some other string. What are the defaults? Is that `True` actually necessary? Can we do it a `Path` if found, `None` if not?

avalentino: Sorry I don't remember all details.
Yes I'm sure that there are better alternatives to the current implementation.
The point here is that.there are some cases in which we know the a library is available in the system but we do not know its path.
For this specific case ``None`` do not helps, we need some other kind of mechanism.

eumiro:
Okay, @avalentino, I will split the changes. Here again come all the changes from above except for applying `Path` to collect the files. The first five commits did not change, but the rest had to be rebased because of the postponed `Path/glob` commit.

When we merge it, we'll have a good base for those `Path/str` hacks. I have already prepared the commit, but will put it into a new PR, so we can discuss the diffs.

setup.py
eumiro: Postponed to a new PR.

setup.py
eumiro: postponed to a new PR

avalentino (approved):

PR #859, opened 2021-01-07
Use pathlib for dir/file operations

eumiro:
Replace all `os.path`/`open` calls with `pathlib.Path`. The changes are mostly local. A possibility of more global `Path` objects has to be investigated later.

### DIFF
--- a/bench/blosc.py
+++ b/bench/blosc.py
@@ -1,5 +1,5 @@
-import os
 import sys
+from pathlib import Path
 from time import perf_counter as clock
 import numpy as np
 import tables as tb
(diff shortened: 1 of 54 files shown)


eumiro:
Oops, it looks like `missing_ok` came to `Path.unlink` in Python 3.8 only. I'll have to add the check before it.

eumiro:
Rebased, repushed.

eumiro:
Rebased, repushed. Could we get this to `master` (or discuss it) before other bigger updates take place, such as #867?

avalentino (approved):

PR #860, opened 2021-01-10
Use pathlib to glob files in setup.py

eumiro:
This is a follow-up of #858 with most `os.path` and `glob` calls replaced by `pathlib.Path`, except for `hdrdir` and `libdir` manipulation that has to be investigated separately because it can optionally hold a boolean variable.

Since there are no tests, it has to be reviewed manually. `python setup.py sdist` runs through though.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,8 +8,6 @@ import ctypes
 import tempfile
 import textwrap
 import subprocess
-from os.path import exists, expanduser
-import glob
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


avalentino (approved):

avalentino:
sorry for taking so much time to review

eumiro:
> sorry for taking so much time to review

Thank you, @avalentino . This was also not a simple PR to review, but I'm happy we're one step further.

PR #862, opened 2021-01-24
Use lowercase "float64" as numpy dtype

keszybz:
Float64 is gone with numpy 1.20, which causes doctests to fail
(https://bugzilla.redhat.com/show_bug.cgi?id=1914335).

>>> numpy.__version__
'1.19.4'
>>> [k for k in numpy.sctypeDict.keys() if str(k).lower().startswith('float')]
['float16', 'Float16', 'float32', 'Float32', 'float64', 'Float64', 'float128', 'Float128', 'float_', 'float']

>>> numpy.__version__
'1.20.0rc2'
>>> [k for k in numpy.sctypeDict.keys() if str(k).lower().startswith('float')]
['float16', 'float32', 'float64', 'float128', 'float_', 'float']

### DIFF
--- a/bench/shelve-bench.py
+++ b/bench/shelve-bench.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 from tables import *
-import numpy as NA
+import numpy as np
 import struct
 import sys
 import shelve
@@ -44,8 +44,8 @@ class Medium(IsDescription):
 
 class Big(IsDescription):
     name = StringCol(itemsize=16)   # 16-character String
-    #float1 = Float64Col(shape=32, dflt=NA.arange(32))
-    #float2 = Float64Col(shape=32, dflt=NA.arange(32))
+    #float1 = Float64Col(shape=32, dflt=np.arange(32))
+    #float2 = Float64Col(shape=32, dflt=np.arange(32))
     float1 = Float64Col(shape=32, dflt=range(32))
     float2 = Float64Col(shape=32, dflt=[2.2] * 32)
     ADCcount = Int16Col()           # signed short integer
@@ -64,8 +64,8 @@ def createFile(filename, totalrows, recsize):
     # Get the record object associated with the new table
     if recsize == "big":
         d = Big()
-        arr = NA.array(NA.arange(32), type=NA.float64)
-        arr2 = NA.array(NA.arange(32), type=NA.float64)
+        arr = np.array(np.arange(32), type=np.float64)
+        arr2 = np.array(np.arange(32), type=np.float64)
     elif recsize == "medium":
         d = Medium()
     else:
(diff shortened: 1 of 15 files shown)


avalentino:
Thanks @keszybz.
The test on ubuntu seems to fail for a network problem, will try to restart the test tomorrow.

eumiro:
Thank you @keszybz
It looks like there are several occurrences of the upper-case version of the numpy types in the whole repo:

```
❯ rg "(numpy|np)\.(Int|Float)"
bench/pytables-search-bench.py
39:        arr_f8 = np.arange(i, stop, type=np.Float64)
40:        arr_i4 = np.arange(i, stop, type=np.Int32)
43:            arr_i4 = np.array(arr_f8, type=np.Int32)

bench/sqlite-search-bench.py
140:                var3 = np.arange(i, j, type=np.Float64)
143:            var2 = np.array(var3, type=np.Int32)

bench/recarray2-test.py
24:x3 = np.array(np.arange(reclen, reclen * 3, 2), np.Float64)

bench/bsddb-table-bench.py
85:        arr = np.array(np.arange(32), type=np.Float64)
86:        arr2 = np.array(np.arange(32), type=np.Float64)
89:        arr = np.array(np.arange(2), type=np.Float64)
109:                #d.float1 = np.array([i]*32, np.Float64)
110:                #d.float2 = np.array([i**2]*32, np.Float64)

bench/postgres-search-bench.py
17:    col_i = numpy.arange(start, stop, type=numpy.Int32)
21:        col_j = numpy.array(col_i, type=numpy.Float64)
```

Would you like to extend your PR? Maybe you'll find other deprecated names (and the way we import `numpy`, which tells me we should consolidate it) in the repository.

keszybz:
PTAL.

bench/shelve-bench.py
```diff
@@ -64,8 +64,8 @@ def createFile(filename, totalrows, recsize):
     # Get the record object associated with the new table
     if recsize == "big":
         d = Big()
-        arr = NA.array(NA.arange(32), type=NA.Float64)
-        arr2 = NA.array(NA.arange(32), type=NA.Float64)
+        arr = np.array(np.arange(32), type=np.float64)
```
eumiro, line 67: First I thought, that this could be written `np.arange(32, type=np.float64)` directly, but `type` is actually an invalid argument in both cases. It should read `dtype`.

@avalentino what's the state of the `bench` scripts? They are not covered by tests.

doc/source/cookbook/custom_data_types.rst
```diff
@@ -65,9 +65,8 @@ Submitted by Kevin R. Thornton.
 
 
     if __name__ == '__main__':
-        x = np.array(np.random.rand(100))
-        x=np.reshape(x,(50,2))
-        x.dtype=[('x',np.float),('y',np.float)]
+        x = np.array(np.random.rand(100)).reshape(50,2)
```
eumiro, line 68:
```suggestion
        x = np.random.rand(100).reshape(50,2)
```

keszybz: Right, that's even nicer. I'll update the commit with that.

keszybz:
I ammended with the suggested fixup and two new commits to do similar cleanups in other places.

eumiro (approved):
LGTM

eumiro:
@keszybz could you please rebase on `master` (I have fixed #863) now? Then we can see what the failed run says. Thank you.

bench/shelve-bench.py
avalentino: Personally I have not put too much attention in keeping them up to date.
Probably we could add a script that executes all of them in CI.

avalentino (approved):

eumiro:
Thank you, @keszybz!

PR #864, opened 2021-01-25
Use `import numpy as np` everywhere

eumiro:
In order to have a consistent import structure (and to discover deprecations like in #862 more easily) , replace all `import numpy` with `import numpy as np` and the `numpy.` with `np.` in the code.

Two usages of `np` as a variable have been replaced by `arr`.

PR #865, opened 2021-01-25
Consistent imports

eumiro:
This is the follow-up of #864 consolidating all `import` statements across the repo.

First, all unused imports are deleted, then the Python2 `__future__` imports and then the rest to have the same names in the whole codebase.

When this is approved, I'd squash all commits within this PR to a single one.

There's one big chapter missing: `import tables as tb` (it's my proposal to use the `tb` alias). But that would be a bigger separate PR.

### DIFF
--- a/tables/tests/test_attributes.py
+++ b/tables/tests/test_attributes.py
@@ -1,7 +1,7 @@
 """This test unit checks node attributes that are persistent (AttributeSet)."""
 
-import sys
 import datetime
+import sys
 import warnings
 from distutils.version import LooseVersion
 
--- a/tables/tests/test_enum.py
+++ b/tables/tests/test_enum.py
@@ -10,8 +10,8 @@
 
 """Test module for enumerated types under PyTables."""
 
-import operator
 import itertools
+import operator
 
 import tables
 from tables.tests import common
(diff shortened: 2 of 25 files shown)


tables/tests/test_attributes.py
```diff
@@ -1,7 +1,7 @@
 """This test unit checks node attributes that are persistent (AttributeSet)."""
 
+import datetime as dt
```
avalentino, line 3: It is my personal taste but I would prefer to keep the original name in this case

tables/tests/test_enum.py
```diff
@@ -11,7 +11,7 @@
 """Test module for enumerated types under PyTables."""
 
 import operator
-import itertools
+import itertools as it
```
avalentino, line 14: also in this case I would prefer to keep "itertools"

avalentino (changes requested):

tables/tests/test_attributes.py
eumiro: No problem, changed.

tables/tests/test_enum.py
eumiro: There are only a few lines of code, no problem.

avalentino (approved):

Issue #867, opened 2021-01-26
Should we automatically format the Python code?


eumiro:
Due to the organic nature of the project, many parts have been changed during the years by larger find&replace and the formatting diverged.

I'd propose to settle one standard automatic tool, process the whole codebase at once (with ugly diffs) and then set up an automatic check in CI, together with description in the DEVELOPMENT documentation. Or a `pre-commit` check.

My proposal would be `black` with its standard line length of 88 chars, but I am open to any other number, such as 79.

What do you think?

avalentino:
If I understand correctly the strategy would be to have a big reformat once and then a "stylecheck" test in CI that fails if the formatting is not correct. Isn't it?

IMHO the strategy is mostly OK. Of course we should carefully think about details.

There are a lot of good reasons to reformat the code in a clean and readable way but there are also drawbacks.

A big reformat (to be understood how big) means that it will be

* harder to merge/rebase existing branches
* harder to follow "real" code changes in the history

I do not have time now to look at details in this moment but I would like to investigate better the impact.
Do you think that there is some way to mitigate this drawbacks?

Also as a metter of personal taste I am not totally convinced that black is the right tool.
I think it is too strict and sometimes the code reformatted with black looks very strange to my eyes.
I would like to make some tests also with other tools to have a complete picture.

Finally, regarding the line length, personally I use 79 but also 88 it would ok to me.
IMHO we should evaluate the amount of changes is much higher wireframes 88, in that case probably would be better to stay with 79. But please consider this just as a comment to encourage the discussion.

albertosantagostino:
Jumping in just to quickly mention **[yapf](https://github.com/google/yapf)** as a really nice and [highly customizable](https://github.com/google/yapf#knobs) formatter compared to black

eumiro:
The best moment to reformat the whole repo was at the start of the project. The second best moment is now.

When doing larger refactorings (such as the consistent imports now, or f-strings, or pathlib, or unittest→pytest), there are hundreds or thousands of lines affected. And most of them are changed by a simple find&replace, so I even do not touch these lines manually, just read over them afterwards. Sometimes the lines get longer, so I should reformat them, but this can/should be automated. Or checked automatically if I use some my IDE's tools. But my IDE does probably a different job than yours and when I reformat some long expression that spans over three lines, your commit may reformat it back.

[...]

PR #868, opened 2021-01-26
Use `import tables as tb` and `from tables.tests import common` everywhere

eumiro:
Consolidate `tables` imports. Use `import tables as tb` everywhere. In tests use `from tables.tests import common`.

This cleans up the code and allows easier grep for usage of methods and attributes.

The formatting is not perfect now, but I'd postpone it to #867.


contrib/make_hdf.py
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env python
 
+import cPickle
```
avalentino, line 3: Probably this should be just pickle

avalentino (commented):
I would have preferred something like "tbcommon" instead of "common" but I see that "common" was already used a lot so I think it is batter to go with your solution.

contrib/make_hdf.py
eumiro: This `contrib/make_hdf.py` is a Python2 file with `print` statement, `.iter*()` dict methods and `raw_input()`. I can rewrite it to Python3, but I am wondering nobody has touched this file since 2012.

eumiro:
> I would have preferred something like "tbcommon" instead of "common" but I see that "common" was already used a lot so I think it is batter to go with your solution.

You're right, `tbcommon` would be more descriptive. But you've seen my `pytest` attempts and I feel most of `common.py` functionality will be put into `conftest.py` as fixtures, so it may disappear soon entirely.

contrib/make_hdf.py
avalentino: Yes please, keeping Python2 does not make sense.
In general bench and contrib have never received too much love :/

eumiro: That was not only Python 3 migration. That was several versions of PyTables worth updating. There are no tests, but the script `make_hdf.py` does something now. Please see the second commit.

avalentino (approved):

contrib/make_hdf.py
avalentino: thanks @eumiro

PR #866, opened 2021-01-26
Remove calls to numpy deprecated APIs

keszybz:
The tests pass, but I didn't do any benchmarking...

Issue #854, opened 2021-01-05
ImportError: Could not load any of ['hdf5.dll', 'hdf5dll.dll', 'pytables_hdf5.dll'], please ensure that it is installed in the package folder. Windows 10


Zachweiss20:
python3 -m tables.tests.test_all
Traceback (most recent call last):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.8_3.8.2032.0_x64__qbz5n2kfra8p0\lib\runpy.py", line 185, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.8_3.8.2032.0_x64__qbz5n2kfra8p0\lib\runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "C:\Users\name\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.8_qbz5n2kfra8p0\LocalCache\local-packages\Python38\site-packages\tables\__init__.py", line 82, in <module>  
    raise ImportError(
ImportError: Could not load any of ['hdf5.dll', 'hdf5dll.dll', 'pytables_hdf5.dll'], please ensure that it is installed in the package folder.
PS C:\Users\name\Documents\GitHub\name>

avalentino:
@Zachweiss20 please provide more information:
* windows version?
* python version ? (seems to be 3.8)
* PyTables version?
* Are you using conda?
* Are you using wheels?
* how have you installed hdf5 and pytabels?

Zachweiss20:
Hey @avalentino, to answer your questions:
1. Windows 10
2. Python 3.8
3. I have installed tables through pip install tables;  it's version 3.6.1
4. Not using conda
5. Not using wheels
6. I installed h5py 3.1.0, and again from what i've read pytables is ~ tables, I've installed tables ver 3.6.1

avalentino:
Thanks @Zachweiss20.
Can you please also tell us if during the installation process it has been downloaded ``tables-3.6.1-2-cp38-cp38-win_amd64.whl`` or ``tables-3.6.1.tar.gz``?

Zachweiss20:
tables-3.6.1-2-cp38-cp38-win_amd64.whl
Ah, I guess that's the problem: Since i'm not using wheels this version isn't working for me?

avalentino:
[...]

Issue #824, opened 2020-10-25
 Importing pytables (3.6.1) in python 3.7.8 fails for me in windows system


Sophia409:
Hi,

After installing and uninstalling pytables packages for many times, I still couldn't import tables. 

Hope for your help

```
(base) C:\Users\yuhong>conda list | grep pytables
pytables                  3.6.1            py37h14417ae_3    conda-forge  
(base) C:\Users\yuhong>python

Python 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 01:53:57) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import tables
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\yuhong\AppData\Roaming\Python\Python37\site-packages\tables\__init__.py", line 99, in <module>
    from .utilsextension import (
ImportError: DLL load failed: The specified procedure could not be found.
```

exrich:
+1.  I have the same issue on python 3.7.9.  This seems to be related to blosc and perhaps a module called snappy.  Downgrading to blosc=1.20.0 caused snappy to be reinstalled and fixed the issue for me.

avalentino:
@Sophia409, @exrich thanks for reporting.
As far as I can understand the issue seems to be mainly related to conda packaging of pytables.
Probably it is a good idea to open a corresponding issue on https://github.com/conda-forge/pytables-feedstock.

e-dism:
I have the same problem by trying to import tables in a python 3.7.9 environment. It seems, that the DLLs are not delivered correctly via pip. Installing the matching wheel file from https://www.lfd.uci.edu/~gohlke/pythonlibs/#pytables let me import tables without error messages.

avalentino:
@Sophia409 the issue seems to be very similar to #913.
Please try to re-install in a fresh conda environment.

@e-dism the issue with wheels has been solved in #926.

PR #872, opened 2021-01-30
Expand CI and wheel-building to more platforms using GHA

amotl:
Hi there,

this is a preliminary draft for resolving #823 and more. It builds upon @eumiro's #834 and @avalentino's b71ca54 (thanks!). I am still working on it on different details [1] and will also add some topics to discuss later.

With kind regards,
Andreas.

[1] Currently, there is still a gap on building 32-bit wheels for Linux, see https://github.com/joerick/cibuildwheel/issues/568.


amotl:
After working through one of the last hassles on behalf of https://github.com/joerick/cibuildwheel/issues/568 (thanks for all your suggestions, @YannickJadoul, @Czaki and @henryiii), the build has completed for the first time [1], yielding
```text
tables-3.6.2.dev0-cp36-cp36m-macosx_10_9_x86_64.whl
tables-3.6.2.dev0-cp36-cp36m-manylinux2010_i686.whl
tables-3.6.2.dev0-cp36-cp36m-manylinux2014_x86_64.whl
tables-3.6.2.dev0-cp36-cp36m-win32.whl
tables-3.6.2.dev0-cp36-cp36m-win_amd64.whl
tables-3.6.2.dev0-cp37-cp37m-macosx_10_9_x86_64.whl
tables-3.6.2.dev0-cp37-cp37m-manylinux2010_i686.whl
tables-3.6.2.dev0-cp37-cp37m-manylinux2014_x86_64.whl
tables-3.6.2.dev0-cp37-cp37m-win32.whl
tables-3.6.2.dev0-cp37-cp37m-win_amd64.whl
tables-3.6.2.dev0-cp38-cp38-macosx_10_9_x86_64.whl
tables-3.6.2.dev0-cp38-cp38-manylinux2010_i686.whl
tables-3.6.2.dev0-cp38-cp38-manylinux2014_x86_64.whl
tables-3.6.2.dev0-cp38-cp38-win32.whl
tables-3.6.2.dev0-cp38-cp38-win_amd64.whl
tables-3.6.2.dev0-cp39-cp39-macosx_10_9_x86_64.whl
tables-3.6.2.dev0-cp39-cp39-manylinux2010_i686.whl
tables-3.6.2.dev0-cp39-cp39-manylinux2014_x86_64.whl
tables-3.6.2.dev0.tar.gz
```

[1] https://github.com/daq-tools/PyTables/actions/runs/525956952

amotl:
As you can see from the list above - and as outlined at [1,2] - there is one thing missing yet. The build process croaks when trying to build wheels on Python 3.9/Windows with `fatal error LNK1181: cannot open input file 'hdf5.lib'`.

I can see the `setup.py` already does many sophisticated things in order to get things straight in this regard. However, on this specific case, it fails. Maybe one of the authors of these routines, eventually owning a Windows machine with a respective development environment can have a look? Maybe something like 29481179 (#540) by @tadeu or some adjustments to that is needed?

cc @FrancescAlted, @avalentino, @eumiro, @andreabedini, @tomkooij, @robbmcleod, @tadeu

[1] https://github.com/daq-tools/PyTables/blob/1acef23a79f54c404549c147cfc9e403f3f8e195/.github/workflows/wheels.yml#L134
[2] https://github.com/daq-tools/PyTables/blob/1acef23a79f54c404549c147cfc9e403f3f8e195/.github/workflows/wheels.yml#L147

avalentino:
Great news! Thank you very much @amotl.
As suggested in the other issue probably the better way to proceed at this point is to merge this PR and address the remaining windows/python3.9 part in a separate PR (maybe some windows user can step forward for this).

Is the PR ready for review or do you need to make some additional change before?

amotl:
Hi Antonio,

sure, let's go ahead and merge this. All further improvements can be done within a later iteration.

> Is the PR ready for review or do you need to make some additional change before?

I will just rebase on top of current master and remove the `testing.pypi.org` line.

With kind regards,
Andreas.

avalentino:
Dear @amotl I have just pushed a change to update minimum required version for Cython and numpy (it is necessary to properly support numpy 1.20, see #866).
Can you please rebase or make the corresponding updates in this PR?

Also please note that this PR is still marked as *draft*.

amotl:
Hi @avalentino and @eumiro,

the patch is ready for review. Please note that, in order to align to the convention of the _PyPI publish GitHub Action_ [1], `secrets.pypi_password` has been renamed to `secrets.PYPI_API_TOKEN`.

I will collect some more other items to discuss within a separate issue as I learned you are keen to merge this early.

With kind regards,
Andreas.

[1] https://github.com/pypa/gh-action-pypi-publish

henryiii:
> In order to align to the convention of the PyPI publish GitHub Action, `secrets.pypi_password` has been renamed to `secrets.PYPI_API_TOKEN`.

This was a fairly recent change, though it makes sense:

https://github.com/pypa/gh-action-pypi-publish/commit/4425980a3392e9b685c45af290ba37dc79870af9

eumiro (approved):
LGTM (and I am learning while reviewing this!)

amotl:
I just gave the macOS wheel a spin on my own machine still running _macOS High Sierra Version 10.13.6_.
```sh
virtualenv .venv --python=python3.8
source .venv/bin/activate
pip install ~/Downloads/artifact/tables-3.6.2.dev0-cp38-cp38-macosx_10_9_x86_64.whl

$ python
Python 3.8.6 (default, Dec  7 2020, 19:03:02)
[Clang 10.0.0 (clang-1000.10.44.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
```

and it croaked
```python
>>> import tables
  File "<stdin>", line 1, in <module>
  File "/path/to/testdrive/pytables/.venv/lib/python3.8/site-packages/tables/__init__.py", line 22, in <module>
    from .utilsextension import (
ImportError: dlopen(/path/to/testdrive/pytables/.venv/lib/python3.8/site-packages/tables/utilsextension.cpython-38-darwin.so, 2): Symbol not found: ____chkstk_darwin
  Referenced from: /path/to/testdrive/pytables/.venv/lib/python3.8/site-packages/tables/.dylibs/libsz.2.dylib (which was built for Mac OS X 10.15)
  Expected in: /usr/lib/libSystem.B.dylib
 in /path/to/testdrive/pytables/.venv/lib/python3.8/site-packages/tables/.dylibs/libsz.2.dylib
```

Bummer.

More information about supporting older releases of macOS on GHA can be found at https://github.com/actions/virtual-environments/issues/461 and https://github.com/actions/virtual-environments/issues/1256.

PyPI still has appropriate versions targeting older releases of macOS, as one can deduce from package names like `tables-3.6.1-cp37-cp37m-macosx_10_6_intel.whl` or `tables-3.6.1-cp38-cp38-macosx_10_9_x86_64.whl` [1].

On the other hand, I am wondering why the newly minted wheels also have names like `tables-3.6.2.dev0-cp38-cp38-macosx_10_9_x86_64.whl` despite being targeted for _macOS 10.15_? Maybe @joerick or @YannickJadoul know anything about it?

However, the recipe by @Safihre at https://github.com/actions/virtual-environments/issues/1256#issuecomment-770270252 might come to the rescue. I will try.

[1] https://pypi.org/project/tables/#files

.github/workflows/wheels.yml
```diff
@@ -14,33 +14,143 @@ on:
...
+        run: cibuildwheel --output-dir wheelhouse
+        env:
+          CIBW_BUILD: "cp36-manylinux_x86_64 cp37-manylinux_x86_64 cp38-manylinux_x86_64 cp39-manylinux_x86_64 cp36-macosx_x86_64 cp37-macosx_x86_64 cp38-macosx_x86_64 cp39-macosx_x86_64"
+
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+
+          CIBW_ENVIRONMENT: "PYTABLES_NO_EMBEDDED_LIBS=true DISABLE_AVX2=true"
+          CIBW_BEFORE_ALL_LINUX: "yum install -y blosc-devel bzip2-devel hdf5-devel lz4-devel lzo-devel snappy-devel libzstd-devel zlib-devel"
+          CIBW_BEFORE_ALL_MACOS: "brew install c-blosc bzip2 hdf5 lz4 lzo snappy zstd zlib"
+          CIBW_BEFORE_BUILD: "pip install --requirement=requirements.txt --requirement=requirements-dev.txt"
```
Czaki, line 82: please add this line to `env`
```
MACOSX_DEPLOYMENT_TARGET: "10.9"
```

henryiii:
`cibuildwheel` correctly downloads the official Python to build with. As long as MACOSX_DEPLOYMENT_TARGET is set, this should be absolutely fine. In fact, during the wheel build, it will be set (if unset) for you. But if you build anything else (libsz.2.dylib), you are going to have to make sure you also set `MACOSX_DEPLOYMENT_TARGET` when you build that - `cibuidlwheel` can't set that for you. :)

Czaki:
> On the other hand, I am wondering why the newly minted wheels also have names like `tables-3.6.2.dev0-cp38-cp38-macosx_10_9_x86_64.whl` despite being targeted for _macOS 10.15_? Maybe @joerick or @YannickJadoul know anything about it?

The source of the problem is that you install dependencies from brew instead of installing them from the source. brew could be used for developer tools but not for library dependencies.

henryiii:
As a general rule, you should always set `MACOSX_DEPLOYMENT_TARGET` when building anything portable on macOS. macOS has a fantastic ability to compile using the latest compilers but targeting older systems by careful design, but you do have to opt-in. There are some limitations - you can't target all of C++17 and set anything before 10.14 (though parts of C++17 can go back as far as 10.12).

henryiii:
> install dependencies from brew

Ah, good catch. Brew bottles per-OS, so anything from brew will be targeted at that OS.

henryiii:
You can use `brew install --build-from-source`, perhaps, with the env-var set?

Czaki:
@henryiii Maybe you could help me finish PR to delocate with check of `MACOSX_DEPLOYMENT_TARGET`?

YannickJadoul:
Argh, that tricky `MACOSX_DEPLOYMENT_TARGET` thing again? :-| I believed we had actually fixed it in `cibuildwheel`, but not for external dependencies :-(

Czaki:
> Argh, that tricky `MACOSX_DEPLOYMENT_TARGET` thing again? :-| I believed we had actually fixed it in `cibuildwheel`, but not for external dependencies :-(

The main problem is `brew` usage.

amotl:
Hi again.

Thanks for your valuable feedback, @Czaki and @henryiii!

> I will collect some more other items to discuss.

This might already be one of those things I wanted to mention. As you can see from the diff, I thought it wouldn't be a bad idea to use the "most native" way of installing build requirements through packages. Now that I am thinking again about this and supported by your statements like

> The source of the problem is that you install dependencies from brew instead of installing them from the source. brew could be used for developer tools but not for library dependencies.

I recognize that this actually might have been a bad idea. So, if I get this right, we have those options...

- `brew install --build-from-source` paired with `MACOSX_DEPLOYMENT_TARGET`?
- What about also using `conda` for installing build requirements as I did for the Windows builds?
- @Czaki's [`imagecodecs_build`](https://github.com/Czaki/imagecodecs_build) as discussed within https://github.com/joerick/cibuildwheel/issues/568 obviously won't be an option for running macOS builds. But do you believe the only sane option would be to follow the same path of installing the build requirements from their respective sources? I actually somehow expected that to happen.... ;]

With kind regards,
Andreas.

.github/workflows/wheels.yml
amotl: Thank. But as far as I understand now, this won't help because I used `brew` to install build-time prerequisites, right?

amotl:
At least, it looks like it works a modern variant of Linux ;].
```sh
docker run -it --rm --volume ~/Downloads/artifact:/media/wheels python:3.9.1 bash
pip install /media/wheels/tables-3.6.2.dev0-cp39-cp39-manylinux2014_x86_64.whl
python
>>> import tables
>>> tables.print_versions()
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
PyTables version:    3.6.2.dev0
HDF5 version:        1.8.12
NumPy version:       1.19.5
Numexpr version:     2.7.2 (not using Intel's VML/MKL)
Zlib version:        1.2.11 (in Python interpreter)
LZO version:         2.06 (Aug 12 2011)
BZIP2 version:       1.0.6 (6-Sept-2010)
Blosc version:       1.11.3 (2017-03-09)
Blosc compressors:   blosclz (1.0.5), lz4 (1.7.3), lz4hc (1.7.3), snappy (unknown), zlib (1.2.7), zstd (1.3.6)
Blosc filters:       shuffle, bitshuffle
Python version:      3.9.1 (default, Jan 12 2021, 16:45:25)
[GCC 8.3.0]
Platform:            Linux-4.19.76-linuxkit-x86_64-with-glibc2.28
Byte-ordering:       little
Detected cores:      4
Default encoding:    utf-8
Default FS encoding: utf-8
Default locale:      (en_US, UTF-8)
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
```

Czaki:
> @Czaki's [`imagecodecs_build`](https://github.com/Czaki/imagecodecs_build) as discussed within [joerick/cibuildwheel#568](https://github.com/joerick/cibuildwheel/issues/568) obviously won't be an option for running macOS builds. But do you believe the only sane option would be to follow the same path of installing the build requirements from their respective sources? I actually somehow expected that to happen.... ;]

In this repository, the same scripts are used to build a native library for macOS. I need to debug the current fail on macOS, but You could copy it and remove obsolete ones 

https://github.com/Czaki/imagecodecs_build/blob/24fcd092bc30b6c2af826ceab211818c81ede1dd/azure-pipelines.yaml#L187-L200

Czaki:
@amotl `otool -l`  is a tool available on macOS which allow to check minimum system requirements foior given binaries. 

macOS wheels are `tar.gz` compressed files So you could test two first approaches (from source and conda) to check libraries minimum system version.

amotl:
Thanks @Czaki. I will explore some options along the lines of what you are suggesting. Maybe giving Spack [1,2] a spin for installing development packages would also be worth an attempt. However, that would probably just improve the convenience for installing the libraries.

In order to be able to build compatible artifacts, it will all boil down to having the right SDK in a lower version available on the build image, right? As I can see from your build machinery, you are using Azure Pipelines, where - according to https://github.com/actions/virtual-environments/issues/461#issuecomment-590958811 ff. - older macOS releases might have been available. However, it looks like they also might have deprecated 10.13 already.

[1] https://spack.io/
[2] https://github.com/spack/spack

henryiii:
> In order to be able to build compatible artifacts, it will all boil down to having the right SDK in a lower version available on the build image, right? 

**NO**! Apple carefully designs this so that you _never need an older macOS_ to build, but you simply set `MACOSX_DEPLOYMENT_TARGET` to the value you want to support. Clang will error out if there is something required from the runtime stdlib that is not supported on that version of macOS; every new feature is gated by a check against the deployment target. That's why CI systems do not keep older macOS versions around as long as linux versions, etc. Everything build with `MACOSX_DEPLOYMENT_TARGET=10.9` will support 10.9+. The problem here is you are pulling homebrew builds and they are built for the DEPLOYMENT_TARGET. Homebrew doesn't keep building older binaries since they have to ship one set per version, so it's generally the last three versions (10.14, 10.15, 11, 11-AS, I would expect currently).

henryiii:
So you don't need an older SDK, just an environment variable. It's not linux.

amotl:
Hi Henry,

thanks again for clarifying. Now I got that `MACOSX_DEPLOYMENT_TARGET` is actually a Clang option. Nice! So,
```
CIBW_ENVIRONMENT_MACOS: "MACOSX_DEPLOYMENT_TARGET=10.9"
CIBW_BEFORE_ALL_MACOS: "brew install --build-from-source c-blosc bzip2 hdf5 lz4 lzo snappy zstd zlib"
```

would be all I need to make this going?

With kind regards,
Andreas.

henryiii:
Assuming the brew recipe is fine, and it doesn't take too much time, yes. You might be able to cache the built formula to save time for future runs. First needs to work. :)

Czaki:
The only problem could be that recipe in some point set own `MACOSX_DEPLOYMENT_TARGET`.  @YannickJadoul mentioned once that cmake silently set him higer `MACOSX_DEPLOYMENT_TARGET` because he use c++14.

amotl:
> Assuming the brew recipe is fine, and it doesn't take too much time, yes. You might be able to cache the built formula to save time for future runs.

Caching will be probably and dearly needed, yes. On the first run, it takes 18m 3s.

> First needs to work. :)

Sigh, yes ;].

So, the run with the amendments 2777a2d5 is [1]. Unfortunately, the outcome is the same. Probably the existing Homebrew installation provided by the GHA virtual environment [still] picks up [some of] the prerequisite libraries from its vanilla locations. As outlined within [2], we would probably have to a) use `brew reinstall` in order to be more safe than sorry and b) set another bunch of environment variables like
```sh
LDFLAGS="-L/usr/local/opt/bzip2/lib -L/usr/local/opt/zlib/lib"
CPPFLAGS="-I/usr/local/opt/bzip2/include -I/usr/local/opt/zlib/include"
```

[1] https://github.com/daq-tools/PyTables/runs/1810667103
[2] https://github.com/daq-tools/PyTables/runs/1810667103#step:7:44

amotl:
For the next build I am using
```sh
brew reinstall --build-from-source --force c-blosc bzip2 hdf5 lz4 lzo snappy zstd zlib
```

amotl:
Now [2] with an additional:
```sh
brew link --overwrite --force bzip2 zlib
```

[2] https://github.com/daq-tools/PyTables/actions/runs/529307162

.github/workflows/wheels.yml
amotl: Now I got it. The most recent cf3b745 uses
```yaml
env:
    MACOSX_DEPLOYMENT_TARGET: "10.9"
```

instead of
```yaml
env:
    CIBW_ENVIRONMENT_MACOS: "MACOSX_DEPLOYMENT_TARGET=10.9"
```

Thanks!

amotl:
Even when trying hard like
```sh
brew link --overwrite --force bzip2 zlib
```

it says:
```text
  Warning: Refusing to link macOS provided/shadowed software: bzip2
  If you need to have bzip2 first in your PATH run:
    echo 'export PATH="/usr/local/opt/bzip2/bin:$PATH"' >> /Users/runner/.bash_profile
  
  For compilers to find bzip2 you may need to set:
    export LDFLAGS="-L/usr/local/opt/bzip2/lib"
    export CPPFLAGS="-I/usr/local/opt/bzip2/include"
```

```text
  Warning: Refusing to link macOS provided/shadowed software: zlib
  For compilers to find zlib you may need to set:
    export LDFLAGS="-L/usr/local/opt/zlib/lib"
    export CPPFLAGS="-I/usr/local/opt/zlib/include"
  
  For pkg-config to find zlib you may need to set:
    export PKG_CONFIG_PATH="/usr/local/opt/zlib/lib/pkgconfig"
```

-- https://github.com/daq-tools/PyTables/runs/1811113206?check_suite_focus=true#step:7:159

amotl:
Until now, the error is the same, as can be confirmed using `otool -l` - thanks @Czaki.

While the directly built shared libraries are fine
```sh
otool -l /path/to/testdrive/pytables/.venv/lib/python3.8/site-packages/tables/hdf5extension.cpython-38-darwin.so | grep sdk -B1
  version 10.9
      sdk 10.15.6
```

the transitive are apparently not.
```sh
otool -l /Users/amo/dev/tmp/testdrive/pytables/.venv/lib/python3.8/site-packages/tables/.dylibs/libsz.2.dylib | grep platform -A2
  platform macos
       sdk 10.15
     minos 10.15
```

amotl:
Hm. While I tried to augment `LDFLAGS` and `CPPFLAGS` like
```yaml
# macOS specific configuration.
# Best compatibility, even with older releases of macOS.
MACOSX_DEPLOYMENT_TARGET: "10.9"
CIBW_ENVIRONMENT_MACOS: LDFLAGS+="-L/usr/local/opt/bzip2/lib -L/usr/local/opt/zlib/lib" CPPFLAGS+="-I/usr/local/opt/bzip2/include -I/usr/local/opt/zlib/include"
```

and `cibuildwheel` says [1]:
```text
environment: ParsedEnvironment(['LDFLAGS+="-L/usr/local/opt/bzip2/lib -L/usr/local/opt/zlib/lib"', 'CPPFLAGS+="-I/usr/local/opt/bzip2/include -I/usr/local/opt/zlib/include"'])
```

I am still getting the same error. Apparently, `site-packages/tables/.dylibs/libsz.2.dylib` is still not quite right, yet.

[1] https://github.com/daq-tools/PyTables/runs/1811410175?check_suite_focus=true#step:8:30

amotl:
I might have found the culprit. [1] is telling us
```
 ==> Downloading https://homebrew.bintray.com/bottles/szip-2.1.1_1.catalina.bottle.tar.gz
```
which is apparently another dependency of `libhdf5` itself
```
==> Installing dependencies for hdf5: automake and szip
```

One should probably also add the option `brew --no-binaries`?
> `--[no-]binaries`: Disable/enable linking of helper executables (default: enabled).
>
> -- https://docs.brew.sh/Manpage#reinstall-options-formulacask-

Now, `szip-2.1.1_1.catalina.bottle.tar.gz` [2] only contains binaries and no sources.

[1] https://github.com/daq-tools/PyTables/runs/1811611953?check_suite_focus=true#step:7:35
[2] https://homebrew.bintray.com/bottles/szip-2.1.1_1.catalina.bottle.tar.gz

amotl:
After adding `CIBW_BUILD_VERBOSITY: 3`, also warnings can be observed like [1]
```text
ld: warning: dylib (/usr/local/lib/libhdf5.dylib) was built for newer macOS version (10.15) than being linked (10.9)
ld: warning: dylib (/usr/local/lib/libblosc.dylib) was built for newer macOS version (10.15) than being linked (10.9)
```

Would this be sane?

[1] https://github.com/daq-tools/PyTables/runs/1811611953?check_suite_focus=true#step:8:465

amotl:
Now trying beforehand
```sh
brew install --build-from-source --build-bottle --force szip
```

avalentino:
>  ==> Downloading https://homebrew.bintray.com/bottles/szip-2.1.1_1.catalina.bottle.tar.gz

Please note that szip could have an incompatible license (if I remember correctly).
We should carefully check if we can bundle it.

avalentino:
A general comment: the CI and ubuntu workflows was originally designed to cover two different use cases
1. build and test with embedded blosc (en compressor) sources (CI)
2. build against system libraries (ubuntu)
Once this use cases are covered we can decide to use the more convenient strategy for building wheels.
From the above discussion I understand that reducing the number of system libraries could make things easier (there was a commend somewhere regarding the forced choice of the manylinux version in relation to blosc).
So please feel free to build blosc for bundled sources if you think it is better.
Please also not that in this case no share object is generated for blosc and compressors.
Apart for blosc there is only one library that is strictly necessary: HDF5 (bzip2, lzo etc are optional).

.github/workflows/ci.yml
```diff
@@ -4,41 +4,95 @@ on: [push, pull_request]
...
+
+        # Install packages.
+        conda create --yes python=${{ matrix.python-version }}
+        conda install --yes blosc bzip2 hdf5 lz4 lzo snappy zstd zlib
+
+        # Install build dependencies.
+        conda install --yes --file=requirements.txt --file=requirements-dev.txt
+
+    - name: Run build and install in development mode
+      run: python setup.py develop
```
avalentino, line 83: Is there a particular reason to build/install in developer mode?
I would prefer to stay with the old approach to ``pip install`` from the source tarball.
This allows to ensure the the sdist is properly working and using pip to install will be, let's say, required when we will migrate to a more modern setup environment (with pyproject.toml).

.github/workflows/wheels.yml
```diff
@@ -14,33 +15,162 @@ on:
...
 
-      - name: Build wheels
+      - name: Install prerequisites for macOS
+        if: contains(matrix.os, 'macos') # && steps.cache.outputs.cache-hit != 'true'
+        env:
+          # Best compatibility, even with older releases of macOS.
+          MACOSX_DEPLOYMENT_TARGET: "10.9"
         run: |
-          python -m cibuildwheel --output-dir wheelhouse
+          brew install --build-from-source --build-bottle --force szip
```
avalentino, line 80: Szip could be a problem, I need to check the license

.github/workflows/wheels.yml
```diff
@@ -49,49 +179,55 @@ jobs:
...
-          python -m pip install -r requirements.txt
-          python -m pip install cython
-          python -m pip install sphinx>=1.1 sphinx_rtd_theme numpydoc ipython
+          pip install --upgrade setuptools pip wheel
+          pip install --requirement=requirements.txt --requirement=requirements-dev.txt
+          pip install --requirement=requirements-docs.txt
 
       - name: Build sdist
-        run: make PYTHON=python dist
+        run: python setup.py sdist
```
avalentino, line 206: Please not that ``python setup.py sdist`` is not the same of ``make dist``.
The latter also includes in the sdist tarball a pre-built version of the sphinx documentation.

setup.py
```diff
@@ -827,10 +827,10 @@ def get_cython_extfiles(extnames):
...
 
     # Copy additional data for packages that need it.
-    setuptools_kwargs["package_data"] = {
-        "tables.tests": ["*.h5", "*.mat"],
-        "tables.nodes.tests": ["*.dat", "*.xbm", "*.h5"],
-    }
+    #setuptools_kwargs["package_data"] = {
+    #    "tables.tests": ["*.h5", "*.mat"],
+    #    "tables.nodes.tests": ["*.dat", "*.xbm", "*.h5"],
+    #}
```
avalentino, line 833: Is there a particular reason for this?

setup.py
```diff
@@ -1033,37 +1034,36 @@ def compiler_has_flags(compiler, flags):
...
-        long_description="""\
-PyTables is a package for managing hierarchical datasets and
-designed to efficiently cope with extremely large amounts of
-data. PyTables is built on top of the HDF5 library and the
-NumPy package and features an object-oriented interface
-that, combined with C-code generated from Cython sources,
-makes of it a fast, yet extremely easy to use tool for
-interactively save and retrieve large amounts of data.
-""",
+        long_description=README,
```
avalentino, line 1065: Have you checked that PyPI is able to render our README?

avalentino (changes requested):

Czaki:
> Please note that szip could have an incompatible license (if I remember correctly).
> We should carefully check if we can bundle it.

This remind me why I decide to buil and installl libraries in userspace. Same of libraries are present in CI host system, and compiled against current version of system. I this time I do not found painless way to overwrite it, so I know that I need to correct iclude path later.

amotl:
Hi again,

thanks for providing further suggestions and for the detailed review 💯. I will come back to your comments later.

With kind regards,
Andreas.

amotl:
Hi again,

apologies for not having been able to get back to this. I hope to be able to follow up on this soon.

With kind regards,
Andreas.

amotl:
> Maybe giving Spack [1,2] a spin for installing development packages would also be worth an attempt.
>
> [1] https://spack.io/
> [2] https://github.com/spack/spack

> Spack looks really interesting.

@Czaki: Please note there is also vcpkg [3], which is also very interesting in this regard. I haven't used it yet, though.

[3] https://github.com/microsoft/vcpkg

amotl:
# Wheels for Windows

## Introduction
Another excellent piece of software is [_Windows Docker Machine_](https://github.com/StefanScherer/windows-docker-machine) by @StefanScherer and contributors. It will tremendously help us to debug Windows packages on non-Windows machines more conveniently, attaching to #824, https://github.com/PyTables/PyTables/issues/823#issuecomment-770460668 and #854. I have been able to give it a quick shot and compiled a small recipe how I used it on my macOS machine.

## Walkthrough

### Prepare environment
```sh
# Install VirtualBox and Vagrant.
brew install --cask virtualbox
brew install vagrant

# Install Windows VM with Docker environment.
git clone https://github.com/StefanScherer/windows-docker-machine
cd windows-docker-machine
vagrant up --provider virtualbox 2019-box
```

### Acquire wheels from CI
Download build artifact(s)
1. Navigate to https://github.com/daq-tools/PyTables/actions/runs/530575821, scroll down to "Artifacts".
2. Download ZIP archive.
3. Extract ZIP archive somewhere; here: `~/Downloads/artifact`


### Test Windows wheels
```sh
docker --context=2019-box run -it --rm --volume C:/Users/amo/Downloads/artifact:C:/media/wheels python:3.8-windowsservercore-1809 powershell
```
Remark: The left hand side of the bind mount feels weird, but that's how it's supposed to be. Truth is stranger than fiction.

```powershell
PS C:\> pip install .\media\wheels\tables-3.6.2.dev0-cp38-cp38-win_amd64.whl
```

```python
PS C:\> python

Python 3.8.7 (tags/v3.8.7:6503f05, Dec 21 2020, 17:59:51) [MSC v.1928 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import tables
  File "<stdin>", line 1, in <module>
  File "C:\Python\lib\site-packages\tables\__init__.py", line 12, in <module>
    from .utilsextension import (
ImportError: DLL load failed while importing utilsextension: The specified module could not be found.
>>> exit()
```

### Exit and shutdown environment
```powershell
PS C:\> exit
```

```sh
vagrant halt
```

Czaki:
You may want to try delvewheel to fix windows wheels.
https://github.com/joerick/cibuildwheel/issues/459#issuecomment-760969447

.github/workflows/ci.yml
```diff
@@ -4,41 +4,95 @@ on: [push, pull_request]
...
-        python-version: ${{ matrix.python }}
-    - name: Install APT packages
-      if: contains(${{ matrix.os }}, 'ubuntu')
-      run: |
-        sudo apt install libbz2-dev libhdf5-serial-dev liblzo2-dev
-    - name: Install dependencies
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install Debian packages
+      if: contains(matrix.os, 'ubuntu')
```
Czaki, line 36: use rather 
```
if: runner.os == 'Linux'
```
than checking system in the matrix.

amotl: Thanks for this suggestion!

Czaki:
But see here: https://github.com/andrix/python-snappy/pull/103

This repair tool need separate job for windows 32 and 64 bits.

RalfG:
@amotl, @avalentino,

Thanks for the great work so far! Where did this PR strand? Is there anything I can do to help? I have some experience with cibuildwheel and GitHub Actions. I also have a Windows machine to test on, if this would be required.

avalentino:
@RalfG I think that this PR has been superseded by #927, #924, #923, #917 and #918.
Some of them are based on this PR.
Now there are still w PR related to wheels still open: #929 and #930.
If you want you could help with the latter ones.
Of course any improvement to what is currently in master is welcome.

@amotl I'm going to close this PR. Please feel free to re-open if you thisn that there is nome feature that has not been implemented in the above mentioned PRs.

Issue #874, opened 2021-01-31
Doctests fail


avalentino:
```
$ env PYTHONPATH=. python3 -m pytest --doctest-only tables
============================================================= test session starts =============================================================
platform linux -- Python 3.8.6, pytest-4.6.11, py-1.9.0, pluggy-0.13.0
rootdir: /home/antonio/projects/PyTables
plugins: remotedata-0.3.2, hypothesis-5.29.0, openfiles-0.5.0, flake8-1.0.6, arraydiff-0.3, filter-subpackage-0.1.1, astropy-header-0.1.2, cov-2.8.1, doctestplus-0.7.0, cython-0.1.0
collected 37 items                                                                                                                            

tables/atom.py .........                                                                                                                [ 24%]
tables/attributeset.py F                                                                                                                [ 27%]
tables/expression.py F                                                                                                                  [ 29%]
tables/file.py F                                                                                                                        [ 32%]
tables/filters.py ..                                                                                                                    [ 37%]
tables/group.py FF                                                                                                                      [ 43%]
tables/link.py FFFFF                                                                                                                    [ 56%]
tables/path.py ....                                                                                                                     [ 67%]
tables/table.py F                                                                                                                       [ 70%]
tables/utils.py F                                                                                                                       [ 72%]
tables/misc/enum.py ..........                                                                                                          [100%]

================================================================== FAILURES ===================================================================
_________________________________________________ [doctest] tables.attributeset.AttributeSet __________________________________________________
112         >>> import os, tempfile
113         >>> import tables as tb
114         >>>
115         >>> class MyClass(object):
116         ...   foo = 'bar'
117         ...
118         >>> myObject = MyClass()  # save object of custom class in HDF5 attr
119         >>> h5fname = tempfile.mktemp(suffix='.h5')
120         >>> h5f = tb.open_file(h5fname, 'w')
121         >>> h5f.root._v_attrs.obj = myObject  # store the object
UNEXPECTED EXCEPTION: PicklingError("Can't pickle <class 'tables.attributeset.MyClass'>: attribute lookup MyClass on tables.attributeset failed")
Traceback (most recent call last):

  File "/usr/lib/python3.8/doctest.py", line 1336, in __run
    exec(compile(example.source, filename, "single",

  File "<doctest tables.attributeset.AttributeSet[6]>", line 1, in <module>

  File "/home/antonio/projects/PyTables/tables/attributeset.py", line 488, in __setattr__
    self._g__setattr(name, value)

[...]

PR #877, opened 2021-01-31
Fix doctests

eumiro:
This should fix all failed doctests in #874, but does only a few of them. Most are still failing:

1. `tables.attributeset.AttributeSet`
2. `tables.file.File.__str__`
3. `tables.group.Group.__repr__`
4. `tables.group.Group.__str__`
5. `tables.link.ExternalLink.__call__`
6. `tables.link.ExternalLink.__str__`
7. `tables.link.SoftLink.__call__`
8. `tables.link.SoftLink.__str__`
9. `tables.table.Table.where`

`1` raises `PicklingError`. `2–8` require an existing file that does not exist. `9` is a short piece of code that expects an existing structure beforehand.

### DIFF
--- a/tables/expression.py
+++ b/tables/expression.py
@@ -60,13 +60,15 @@ class Expr:
     --------
     The following shows an example of using Expr.
 
-        >>> a = f.create_array('/', 'a', np.array([1,2,3]))
-        >>> b = f.create_array('/', 'b', np.array([3,4,5]))
+        >>> import tables as tb, numpy as np
+        >>> h5f = tb.open_file('/tmp/expr.h5', 'w')
+        >>> a = h5f.create_array('/', 'a', np.array([1,2,3]))
+        >>> b = h5f.create_array('/', 'b', np.array([3,4,5]))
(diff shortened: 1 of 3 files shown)


eumiro:
Some of these doctests read the file `data/test.h5`. But when I do `git log --all --full-history -- "*test.h5*"`, such file is not in the history of the repository. @avalentino  any idea? Is it possible, that these doctests actually never worked and served as an illustration only?

avalentino:
>  @avalentino any idea? Is it possible, that these doctests actually never worked and served as an illustration only?

@eumiro yes it is possible

eumiro:
Do we then want to have fully testable doctests?

eumiro:
In the example the files were created in 2004. `git` was born one year later. :man_shrugging:

avalentino:
IMHO we should make testable what we can without too much stress 
The other things can stay as examples as they are now

avalentino:
Fixed in cf9518757.

avalentino:
See also #874.

PR #890, opened 2021-03-10
Proposed fix for infinite loop at lrucacheextension.pyx

zequihg50:
Hi,

Under non specific conditions I'm sometimes running into an infinite loop at [lrucacheextension.pyx:372](https://github.com/PyTables/PyTables/blob/10985dada67284ffecac3c0aa196d4afee254116/tables/lrucacheextension.pyx#L372). When the code enters the `while` loop, it tries to remove a `node` from the cache through a call to [removeslot_](https://github.com/PyTables/PyTables/blob/10985dada67284ffecac3c0aa196d4afee254116/tables/lrucacheextension.pyx#L377) but if the node is `None` it does not free cache space (see [this](https://github.com/PyTables/PyTables/blob/10985dada67284ffecac3c0aa196d4afee254116/tables/lrucacheextension.pyx#L352)).

I have added the following line before [this one](https://github.com/PyTables/PyTables/blob/10985dada67284ffecac3c0aa196d4afee254116/tables/lrucacheextension.pyx#L374) to prevent selecting `None` nodes. Although this has worked for the concrete example I was testing I'm not sure if this is an adequate modification.

```python
while size + self.cachesize > self.maxcachesize:
    not_nones = [x is not None for x in self.__list]
    largidx = self.sizes[not_nones].argsort()[-10:]
    ...
```

Would someone be able to review this modification?

### DIFF
--- a/tables/lrucacheextension.pyx
+++ b/tables/lrucacheextension.pyx
@@ -371,7 +371,8 @@ cdef class ObjectCache(BaseCache):
     # Protection against too large data cache size
     while size + self.cachesize > self.maxcachesize:
       # Remove the LRU node among the 10 largest ones
-      largidx = self.sizes.argsort()[-10:]
+      not_nones = [x is not None for x in self.__list]
+      largidx = self.sizes[not_nones].argsort()[-10:]
       nslot1 = self.atimes[largidx].argmin()
       nslot2 = largidx[nslot1]
       self.removeslot_(nslot2)


avalentino:
@zequihg50 I'm not sure hap you can index the array of sizes with ``not_nones``.
I would like to make a deeper analysis.
Are you able to provide a self-contained test code to reproduce the issue?

Another question: have you had the chance to run the ``--heavy`` test suite with your fix?

zequihg50:
I wasn't able to run the tests with the `--heavy` flag, I was running `python setup.py test`.

The following PyTables file is an example of the infinite loop:

[infloop.zip](https://github.com/PyTables/PyTables/files/6191168/infloop.zip)

This is the code I use to read data:

```python
import tables

f = tables.open_file('infloop.hdf', 'r') 
t = f.get_node('/files')

step = 100000
s = set()
for i in range(len(t)//step + 1):
    s.update( t.colindexes['_eva_ensemble_aggregation'].read_sorted(i*step, i*step+step) )

for i,agg in enumerate(sorted(s)):
    if len(agg) == 0:
        continue

    dataset = t.where('''_eva_ensemble_aggregation == {}'''.format(agg))
    print('i={}, agg={}'.format(i,agg))

f.close()
```
I see the following output:

```
i=1, agg=b'CMIP6.AerChemMIP.AS-RCEC.TaiESM1.histSST.AERmon.gn.v20201223@esgf-data1.llnl.gov'
i=2, agg=b'CMIP6.AerChemMIP.AS-RCEC.TaiESM1.histSST.AERmon.gn.v20201223@esgf.rcec.sinica.edu.tw'
i=3, agg=b'CMIP6.AerChemMIP.AS-RCEC.TaiESM1.histSST.Amon.gm.v20200309@esgf.rcec.sinica.edu.tw'
...
i=101, agg=b'CMIP6.AerChemMIP.BCC.BCC-ESM1.hist-piNTCF.SIday.gn.v20190620@cmip.bcc.cma.cn'
```

It should print additional elements after 101 but it freezes in the infinite loop.

I'm running the current [master version](https://github.com/PyTables/PyTables/tree/0eed850b9031fb540edd2c1ff5c81b91efeba9d6).

Forget about the fix I proposed above, it does not work.

avalentino:
Fixed in #1321, thanks to @alpae

Issue #821, opened 2020-10-02
Column attributes (e.g., units)?


jondoesntgit:
I see that there is an API for assigning attributes to groups and tables, but no API for assigning attributes to groups. However, PyTables adds table attributes for column metadata such as FIELD_8_NAME, FIELD_8_FILL, etc... Is there some sort of API for possibly assigning metadata? For example

```python
import tables as pt
import numpy as np

class Description(pt.IsDescription):
    voltage = pt.Float32Col(dflt=np.nan, attrs={'unit': 'V'})
    temperature = pt.Float32Col(dflt=np.nan, attrs={'unit': 'degC'})
```

would create table metadata

```python
FIELD_1_NAME = 'voltage'
FIELD_1_FILL = np.nan
FIELD_1_unit = 'V'
FIELD_2_NAME = 'temperature'
FIELD_2_FILL = np.nan
FIELD_2_unit = 'degC'
```

I know there is a package called [pint](https://github.com/hgrecco/pint) which provides unit-awareness for numpy, a package called [pint-xarray](https://github.com/xarray-contrib/pint-xarray/) that provides unit-awareness for xarray, and [pint-pandas](https://github.com/hgrecco/pint-pandas) which provides unit-awareness for pandas. Is there anything that can be done for pytables?

jondoesntgit:
Would the maintainers be open to a pull request that adds an API for column-level attributes? Something like

```
h5f.root.table.cols.voltage.unit = 'V'
```

avalentino:
Dear @jondoesntgit thanks for your interest in pytables.
IMHO the one you are proposing is a very nice addition to the pytables features, a PR would be very welcome.

eumiro:
Should this be only the `unit` attribute (as a string or some enum to facilitate identical columns in similar tables?) or general metadata for a column (as a dict of anything serializable)?

jondoesntgit:
What I had in mind was using this regular expression `FIELD_\d+_([^\s]*)` to match attributes that come after FIELD_1_{attribute key goes here}. But maybe something like `FIELD_1_ATTRS` as a dict would be better.

jondoesntgit:
Comparing the two options:

## Serializing a dictionary

This keeps all of the column attributes in one serialized dictionary at `FIELD_1_ATTRS`, but any modifying requires re-generating the dictionary and rewriting it to the file. 

## Matching a regular expression

[...]

PR #893, opened 2021-03-19
Added basic API for Column-level attributes (issue #821)

jondoesntgit:
The column-level attributes are only accessible once the table and the columns are instantiated. There's currently no support for adding attributes on creation (e.g., something like `pt.Float32Col(dflt=0.0, shape=(3,2), attrs=dict(foo='bar', description='some string')`)

### DIFF
--- a/tables/attributeset.py
+++ b/tables/attributeset.py
@@ -680,6 +680,139 @@ be ready to see PyTables asking for *lots* of memory and possibly slow I/O"""
             return str(self)
 
 
+class ColumnAttributeSet:
+
+    def __init__(self, column):
+
+        self._v_tableattrs = column.table.attrs
+        self.__dict__['_v_fieldindex'] = column._v_pos
+        self.__dict__['_v_column'] = column # Does this need to be a weak reference?
+
+    def issystemcolumnname(self, key):
+        """Checks whether a key is a reserved attribute name, or should be passed through."""
+        return key in ['_v_tableattrs', '_v_fieldindex', '_v_column']
+
+    def _prefix(self, string):
+        """Prefixes a key with a special pattern for storing with table attributes"""
+        field_index = self.__dict__['_v_fieldindex']
+        return 'FIELD_%i_ATTR_%s' % (field_index, string)
+
+    def __getattr__(self, key):
+        """Retrieves a PyTables attribute for this column"""        
+        if not self.issystemcolumnname(key):
+            return getattr(self._v_tableattrs, self._prefix(key))
+        else:
+            return super().__getattr__(key)
+
+    def __setattr__(self, key, val):
+        """Sets a PyTables attribute for this column"""        
+        if not self.issystemcolumnname(key):
+            setattr(self._v_tableattrs, self._prefix(key), val)
+        else:
+            return super().__setattr__(key, val)
+
+    def __getitem__(self, key):
+        """A dictionary-like interface for __getattr__"""
+        if not self.issystemcolumnname(key):
+            return self._v_tableattrs[self._prefix(key)]
+        else:
+            return self[key]
+
+    def __setitem__(self, key, value):
+        """A dictionary-like interface for __setattr__"""
+        if not self.issystemcolumnname(key):
+            self._v_tableattrs[self._prefix(key)] = value
+        else:
+            self[key] = value
+
+    def __delattr__(self, key):
+        """Deletes the attribute for this column"""
+        if self.issystemcolumnname(key):
+            # This is a bad idea, but if the user really wants it, do it
+            super().__delattr__(key)
+        else:
+            delattr(self._v_tableattrs, self._prefix(key))
+
+    def __delitem__(self, key):
+        """A dictionary-like interface for __delattr__"""
+        if self.issystemcolumnname(key):
+            # This is a bad idea, but if the user really wants it, do it
+            super().__delitem__(key)
+        else:
+            del self._v_tableattrs[self._prefix(key)]
+
+    def _f_rename(self, oldattrname, newattrname):
+        """Rename an attribute from oldattrname to newattrname."""
+
+        if oldattrname == newattrname:
+            # Do nothing
+            return
+
+        # First, fetch the value of the oldattrname
+        attrvalue = getattr(self, oldattrname)
+
+        # Now, create the new attribute
+        setattr(self, newattrname, attrvalue)
+
+        # Finally, remove the old attribute
+        delattr(self, oldattrname)        
+
+    def _f_copy(self, where):
+        """Copy attributes to another column"""
+
+        # Is there a better way to do this?
+        from .table import Column
+        if not isinstance(where, Column):
+            raise TypeError(f"destination object is not a column: {where!r}")
+
+        for key in self.keys():
+            where.attrs[key] = self[key]
+
+    def keys(self):
+        """Returns the list of attributes for this column"""
+        col_prefix = self._prefix('')
+        length = len(col_prefix)
+        return [key[length:] for key in self._v_tableattrs._v_attrnames if key.startswith(col_prefix)]
+
+    def contains(self, key):
+        """Returns whether a key is in the attribute set"""
+        return key in self.keys()
+
+    def __str__(self):
+        """A string representation of this object"""
+
+    def __repr__(self):
+        """A detailed string representation for this object"""
+        print('TODO')
+
+
+    def __str__(self):
+        """The string representation for this object."""
+
+
+        # The pathname
+        pathname = self._v_tableattrs._v__nodepath
+        # Get this class name
+        classname = self._v_column.__class__.__name__ #self._v_tableattrs._v_node.__class__.__name__
+        # The attribute names
+        attrnumber = sum(1 for _ in self.keys())
+        # The column name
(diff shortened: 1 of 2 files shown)


tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
...
+        """Returns the list of attributes for this column"""
+        col_prefix = self._prefix('')
+        length = len(col_prefix)
+        return [key[length:] for key in self._v_tableattrs._v_attrnames if key.startswith(col_prefix)]
+
+    def contains(self, key):
+        """Returns whether a key is in the attribute set"""
+        return key in self.keys()
+
+    def __str__(self):
```
avalentino, line 781: Please remove, it is (re-)defined below

tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
...
+        """A string representation of this object"""
+
+    def __repr__(self):
```
avalentino, line 784: Please remove, it is (re-)defined below

tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
...
+        """A detailed string representation for this object"""
+        print('TODO')
+
+
+    def __str__(self):
+        """The string representation for this object."""
+
+
```
avalentino, line 792: Please remove empty lines, not PEP8 compliant

tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
             return str(self)
 
 
+class ColumnAttributeSet:
+
+    def __init__(self, column):
+
+        self._v_tableattrs = column.table.attrs
+        self.__dict__['_v_fieldindex'] = column._v_pos
+        self.__dict__['_v_column'] = column # Does this need to be a weak reference?
```
avalentino, line 689: Probably yes, better to use a weakref

tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
             return str(self)
 
 
+class ColumnAttributeSet:
+
+    def __init__(self, column):
+
+        self._v_tableattrs = column.table.attrs
```
avalentino, line 687: Any particular reason why you don't use access through ``__dict__`` also for this one?

tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
...
+        if not self.issystemcolumnname(key):
+            self._v_tableattrs[self._prefix(key)] = value
+        else:
+            self[key] = value
+
+    def __delattr__(self, key):
+        """Deletes the attribute for this column"""
+        if self.issystemcolumnname(key):
+            # This is a bad idea, but if the user really wants it, do it
+            super().__delattr__(key)
```
avalentino, line 732: To me this should not be allowed.
If the user really wants there is always a way to do it ...

tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
...
+        else:
+            delattr(self._v_tableattrs, self._prefix(key))
+
+    def __delitem__(self, key):
+        """A dictionary-like interface for __delattr__"""
+        if self.issystemcolumnname(key):
+            # This is a bad idea, but if the user really wants it, do it
+            super().__delitem__(key)
```
avalentino, line 740: Same as above

tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
...
+        else:
+            del self._v_tableattrs[self._prefix(key)]
+
+    def _f_rename(self, oldattrname, newattrname):
+        """Rename an attribute from oldattrname to newattrname."""
+
+        if oldattrname == newattrname:
+            # Do nothing
+            return
+
```
avalentino, line 750: Please also check the ``oldattrname`` with ``issystemcolumnname``

tables/table.py
```diff
@@ -3674,3 +3675,15 @@ def __repr__(self):
...
+    @lazyattr
+    def _v_pos(self):
+        return self.descr._v_colobjects[self.name]._v_pos
+
+    @lazyattr
+    def _v_attrs(self):
+        return ColumnAttributeSet(self)
+
+    @lazyattr
+    def attrs(self):
```
avalentino, line 3688: Maybe this doesn't need to be lazy

tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
...
+
+    def __str__(self):
+        """The string representation for this object."""
+
+
+        # The pathname
+        pathname = self._v_tableattrs._v__nodepath
+        # Get this class name
+        classname = self._v_column.__class__.__name__ #self._v_tableattrs._v_node.__class__.__name__
```
avalentino, line 796: Please remove the comment

tables/attributeset.py
```diff
@@ -680,6 +680,139 @@ def __repr__(self):
...
+
+        # Finally, remove the old attribute
+        delattr(self, oldattrname)        
+
+    def _f_copy(self, where):
+        """Copy attributes to another column"""
+
+        # Is there a better way to do this?
+        from .table import Column
+        if not isinstance(where, Column):
```
avalentino, line 765: Probably it is better to move this class to ``table.py``

avalentino (changes requested):
In general the PR looks good but there are few things to address.
Also there are two important things missing:
* unit tests
* documentation update

tables/attributeset.py
jondoesntgit: Can you tell me more about this? Are we trying to prevent users from renaming system column names?

avalentino: Yes

mennthor:
Hi :)
What is missing / needs to be done to make this merged?

I'm creating IsDescription classes by wrapper functions to set column lengths for 1 Bit serialised bool arrays, so they take less space with `np.packbits` and `tables.StrCol(int(np.ceil(NBITS / 8.))`.
It would be great to store the number of Bits directly in the method where I'm creating the table IsDescription subclass, because I rely on that information to be present in the table attrs to properly truncate the array when deserialising after reading the table back in.

avalentino (changes requested):
The implementation looks good to me.
Please add
* at least basic unit tests
* basic documentation
* a changelog entry

tables/table.py
```diff
@@ -3674,3 +3675,135 @@ def __repr__(self):
...
+        """A dictionary-like interface for __setattr__"""
+        if not self.issystemcolumnname(key):
+            self._v_tableattrs[self._prefix(key)] = value
+        else:
+            self[key] = value
+
+    def __delattr__(self, key):
+        """Deletes the attribute for this column"""
+        if self.issystemcolumnname(key):
+            raise Exception('Deleting system attributes is prohibited')
```
avalentino, line 3740: Maybe a ``TypeError`` would be more clear in this case.

tables/table.py
```diff
@@ -3674,3 +3675,135 @@ def __repr__(self):
...
+        else:
+            delattr(self._v_tableattrs, self._prefix(key))
+
+    def __delitem__(self, key):
+        """A dictionary-like interface for __delattr__"""
+        if self.issystemcolumnname(key):
+            raise Exception('Deleting system attributes is prohibited')
```
avalentino, line 3747: See the comment above about the exception type

avalentino:
Sorry @jondoesntgit for the long delay to provide a feedback.
The PS seems to be almost ready for merge.

jondoesntgit:
No worries on the delay. I'd like to give some time this December to this,
but there's a lot of other things competing for my time (new family member,
wrapping up PhD, etc...)

Zybulon:
Hi,
This feature is a great idea, I am very interesting in it to store units and description (like you did I guess).
Is there still a lot a work to do ? The requested changes still need to be done ?

mennthor:
@Zybulon I adressed the original leftover tasks quite a while ago, but I'm not sure where this currently stands

avalentino:
Could you please push new changes in this PR?
I hope that @jondoesntgit doesn't mind it.

Zybulon:
I am not very familiar with Github so I might misunderstand some things. @avalentino does not seems very active, so I guess we can't push code in this PR, am I wrong ? Maybe @mennthor  can create a new PR to push its modifications ?

jondoesntgit:
Hello,

Sorry for the late reply. Things got a little crazy over the holidays.

I think when I left off, I just needed to write some unit tests and documentation. I’d love if someone took over this PR.

Zybulon:
@jondoesntgit I think you can accept menthor's commit in your fork, can you do it ? It should be immediate since there is not conflict.

jondoesntgit:
Is `git rebase` the right move at this point?

avalentino:
> Is git rebase the right move at this point?

you can re-base or merge master into your branch

avalentino:
Dear @jondoesntgit, thanks a lot for you effort.
Could you please just remove the `appveyor.yml` file?
I think that ir was removed in master after your initial PR.

jondoesntgit:
@avalentino Done

avalentino (approved):

avalentino:
Thanks again @jondoesntgit

PR #917, opened 2021-12-11
Run CI against python 3.10

xmatthias:
Considering that python 3.10 is now released since quite some time, i think pytables should be running CI against this.

This seems to be possible without major issues - however it will ensure that compatibility with 3.10 is there (and remains there) by running the test-suite against python3.10 as well.


avalentino (approved):

avalentino:
Thanks @xmatthias

PR #919, opened 2021-12-13
CI build failure on windows for Python 3.9 and 3.10

avalentino:


xmatthias:
@avalentino don't waste more time on this for the moment. I've JUST NOW figured out (hopefully) how to fix this.

Should i be unsuccessful / encounter problems with it, I'll let you know latest tomorrow morning (9am Italian time) - otherwise you can expect a PR attempting to fix this issue.

avalentino:
> @avalentino don't waste more time on this for the moment. I've JUST NOW figured out (hopefully) how to fix this - but will need some more time to investigate (it's working on my dev machine ... will need to test it in CI).

Thanks @xmatthias. This would help a lot.
This is indeed something hard to track down for me because I do not have a windows machine available.

avalentino:
Fixed in #923

PR #924, opened 2021-12-14
Readme should no longer show badge from travis

xmatthias:
Readme should no longer show badge from travis, since travis is no longer used.


avalentino (approved):

avalentino:
Thanks @xmatthias

PR #926, opened 2021-12-15
Add windows wheels and independent test to GHA

xmatthias:
Add windows wheels and independent test to GHA

For the most part (especially the windows part), this is based on top of the great work from @amotl in #872.

Built wheels for ubuntu x64, windows (win32 and amd64) are tested in a second CI step.
This ensures that no hdf5 Build dependencies are installed, as these are fresh, new machines.
The build step also does not do the checkout step, so the original source is also not available (intentionally).

python for linux on x86 devices seems to be unavailable based on [this run](https://github.com/xmatthias/PyTables/runs/4526501674?check_suite_focus=true) - which shows that linux binaries are not available through the python setup action (i also don't think 32bit for linux is all too relevant).

**EDIT** You can find the last run which ran the affected CI [here](https://github.com/xmatthias/PyTables/actions/runs/1580025842) - which ran against the latest commit in this branch.

----

Please note the following:
* macOS ci is still missing and will be added in a separate PR
* linux/unix wheels are using a different, older version of hdf5 for now (does not change in this PR). I think the best approach will be to compile that from source to get the latest version (or at least an updated one). 

I have both of the above working (for the most part) - but would like to add things step by step, to keep PR's minimal, and simplify reviews.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -6,6 +6,7 @@ on:
     branches:
       - master
       - 'releases/**'
+      - 'ci/**'
     tags:
       - v*
   release:
@@ -19,8 +20,8 @@ jobs:
     strategy:
       matrix:
         # os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ ubuntu-latest ]
         arch: [x86_64, aarch64]
-        os: [ubuntu-latest]
 
     steps:
       - uses: actions/checkout@v2
@@ -28,10 +29,10 @@ jobs:
       - uses: actions/setup-python@v2
         name: Install Python
         with:
-          python-version: '3.7'
+          python-version: '3.9'
 
       - uses: docker/setup-qemu-action@v1
-        if:  ${{matrix.arch}} == 'aarch64'
+        if:  startsWith(matrix.os, 'ubuntu')
         name: Set up QEMU
 
       - name: Install cibuildwheel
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels.yml
```diff
@@ -19,19 +20,19 @@ jobs:
...
 
       - uses: actions/setup-python@v2
         name: Install Python
         with:
-          python-version: '3.7'
+          python-version: '3.9'
 
       - uses: docker/setup-qemu-action@v1
-        if:  ${{matrix.arch}} == 'aarch64'
+        if:  startsWith(matrix.os, 'ubuntu')
```
avalentino, line 35: Probably the two conditions should go in "and", but it seems not to hurt so no problem.

avalentino (approved):

avalentino:
A huge thank you @xmatthias.
Do you think that now we can drop appveyor builds?

xmatthias:
> Do you think that now we can drop appveyor builds?

i guess we can (intentionally didn't want to do it in this PR though as a short period of redundancy won't hurt our cause).

to add to the above: 
I've manually (and in CI) verified that the windows builds work for a regular python installation (from python.org / python via windows store).
The ONLY case where they do not work on my is  when using miniconda python in combination with the wheels.
```
conda create --yes -n test_env39 python=3.9
conda activate test_env39
pip install tables-3.6.2.dev0-cp39-cp39-win_amd64.whl
```

output:

```
(test_env39) PS C:\Users\User\Downloads> pip install tables-3.6.2.dev0-cp39-cp39-win_amd64.whl
Processing c:\users\user\downloads\tables-3.6.2.dev0-cp39-cp39-win_amd64.whl
Collecting numpy>=1.19.0
  Using cached numpy-1.21.4-cp39-cp39-win_amd64.whl (14.0 MB)
Collecting numexpr>=2.6.2
  Using cached numexpr-2.8.1-cp39-cp39-win_amd64.whl (88 kB)
Collecting packaging
  Using cached packaging-21.3-py3-none-any.whl (40 kB)
Collecting pyparsing!=3.0.5,>=2.0.2
  Using cached pyparsing-3.0.6-py3-none-any.whl (97 kB)
Installing collected packages: pyparsing, packaging, numpy, numexpr, tables
Successfully installed numexpr-2.8.1 numpy-1.21.4 packaging-21.3 pyparsing-3.0.6 tables-3.6.2.dev0
(test_env39) PS C:\Users\User\Downloads> python -m tables.tests.test_all
  File "C:\Users\User\miniconda3\envs\test_env39\lib\runpy.py", line 188, in _run_module_as_main
  File "C:\Users\User\miniconda3\envs\test_env39\lib\runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "C:\Users\User\miniconda3\envs\test_env39\lib\site-packages\tables\__init__.py", line 46, in <module>
    from .utilsextension import (
ImportError: DLL load failed while importing utilsextension: The specified module could not be found.
```

The same problem happens with the wheels built in appveyor (same behavior, same correct functioning in a "plain windows environment", which makes me suspect that it's a conda issue more than an issue with the wheel itself.

I'm no expert with conda, so i have no idea how to fix this - but I'm not sure it's relevant, as i guess the correct way to install pytables in a conda environment might be using `conda install tables`.

avalentino:
> i guess we can (intentionally didn't want to do it in this PR though as a short period of redundancy won't hurt our cause).

OK, I will drop appveyor in a week or two if nothing strange happens.

> I'm no expert with conda, so i have no idea how to fix this - but I'm not sure it's relevant, as i guess the correct way to install pytables in a conda environment might be using conda install tables.

@xmatthias I agree with you.
The use case that you spotted is really a "corner case". I think that we can live with it.
If some user really needs to use the wheel package in that way than probably he will be also able to provide a fix.

aufdenkampe:
As long as Conda users can `conda install tables` inside their Conda environments, then I agree that doing `pip install tables` inside a Conda environment would indeed be an unusual case that does not need to be supported.

PR #927, opened 2021-12-15
Add CI wheel build for MacOS

xmatthias:
Adding support for MacOS builds, for both x64 and arm64 builds.
Only the x64 are being tested - and I'd be grateful if someone can step up and try to install the arm64 binary on a M1 system.

A recent CI run for this can be found [here](https://github.com/xmatthias/PyTables/runs/4535307675?check_suite_focus=true).

This is based mostly on the great work from @amotl in #872

---

This is currently using the hdf5 (and blosc) library from brew (as well as some other libraries necessary for building).
While this might for sure be seen as controversial, i don't see this as worse than using libraries from conda (which pytables does and did for Windows CI already).

I'm also working on a "build hdf5 from source" PR - which will build hdf5 from source on linux/macos Systems - but the brew dependency will probably remain there for some "basic" libraries we don't really want to build.

I'm not sure if it would be better to use the builtin / vendored blosc (I'm not sure on reasons for vendoring that library - but discovered that it'll be built as part of the wheel building process if the library is not there).

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ 'ubuntu-latest', 'macos-latest', 'macos-10.15' ]
+        os: [ 'ubuntu-latest', 'macos-10.15' ]
         # arch: [x86_64, aarch64]
         arch: [ x86_64 ]
 
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -19,9 +19,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ 'ubuntu-latest', 'macos-10.15' ]
-        # arch: [x86_64, aarch64]
-        arch: [ x86_64 ]
+        os: [ 'ubuntu-latest', 'macos-latest' ]
+        arch: [x86_64, aarch64]
 
     steps:
       - uses: actions/checkout@v2
@@ -53,7 +52,8 @@ jobs:
         run: |
           python -m cibuildwheel --output-dir wheelhouse
         env:
-          CIBW_ARCHS_LINUX: ${{matrix.arch}}
+          CIBW_ARCHS_LINUX: ${{ matrix.arch }}
+          CIBW_ARCHS_MACOS: ${{ matrix.arch == 'aarch64' && 'arm64' || 'x86_64'}}
           CIBW_BUILD: "cp36-* cp37-* cp38-* cp39-* cp310-*"
           CIBW_BEFORE_ALL_LINUX: "yum -y update && yum install -y epel-release && yum install -y hdf5-devel zlib-devel bzip2-devel lzo-devel blosc-devel"
           CIBW_BEFORE_BUILD: "pip install -r requirements.txt cython>=0.29.21"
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels.yml
```diff
@@ -35,19 +34,32 @@ jobs:
...
+      - name: Install prerequisites for macOS
+        if: contains(matrix.os, 'macos')
+        env:
+          # Best compatibility, even with older releases of macOS.
+          MACOSX_DEPLOYMENT_TARGET: "10.9"
+        run: |
+          brew install --build-from-source --no-binaries --force szip
+          brew reinstall --build-from-source --no-binaries --force c-blosc bzip2 hdf5 lz4 lzo snappy zstd zlib
+          brew link --overwrite --force bzip2 zlib
+
```
avalentino, line 46: Personally I would feel more comfortable with a `conda` based solution, but OK, I will merge this.

.github/workflows/wheels.yml
```diff
@@ -103,12 +115,14 @@ jobs:
...
       matrix:
-        os: [ 'ubuntu-latest', 'windows-latest' ]
+        os: [ 'ubuntu-latest', 'windows-latest', 'macos-latest' ]
         python-version: ['3.7', '3.8', '3.9','3.10']
         arch: ['x64', 'x86']
         exclude:
         - os: 'ubuntu-latest'
           arch: 'x86'
+        - os: 'macos-latest'
+          arch: 'x86'
```
avalentino, line 126: Just realised that we are not testing aarch64 (linux), this is not related to this specific PR anyways.

xmatthias: we are not and we really cannot realistically.

github is not offering ARM based runners at the moment, and running tests with emulation (qemu + docker) is pretty slow (30+ minutes to run the "light" test-suite when i last ran that).

.github/workflows/wheels.yml
xmatthias: i hope there is no misunderstanding here - you should not feel forced to merge PR's you're not comfortable with.

If there is a reason you think using brew is worse than using conda let me know, it's a PR for a reason (allowing discussions, ...) :)

avalentino:
> I'm also working on a "build hdf5 from source" PR - which will build hdf5 from source on linux/macos Systems - but the brew dependency will probably remain there for some "basic" libraries we don't really want to build.
>
> I'm not sure if it would be better to use the builtin / vendored blosc (I'm not sure on reasons for vendoring that library - but discovered that it'll be built as part of the wheel building process if the library is not there).

Blosc was vendored mostly because it was not available via standard packages not even on linux distributions some times ago.

Today it would be probably not necessary to vendor it but I like your proposal to have a zero-dependency procedure the builds everything for the scratch. Having a static linking would be even better, probably, from a wheel user perspective.

.github/workflows/wheels.yml
avalentino: No it is just a personal feeling.
I know conda better than brew so I would prefer to go with conda but this is irrelevant for this PR IMHO.

.github/workflows/wheels.yml
avalentino: > running tests with emulation (qemu + docker) is pretty slow (30+ minutes to run the "light" test-suite when i last ran that)

OK you convinced me
This is something that can be implemented in the future when proper runners will be available.
I will file in issue for that as a reminder.

PR #942, opened 2022-03-05
Bump HDF5 version to 1.13.1.

matham:
Bumps HDF5 version to 1.13.1.

1.13.0 introduced universal binaries for macos. I'm hoping if the upgrade works that we can also add universal wheel support.


avalentino:
Thanks @matham

PR #959, opened 2022-08-17
Direct chunking blosc2

FrancescAlted:
Preliminary version of direct chunking I/O with Blosc2.  Only tables are supported for now, but blosc2 should be usable from all containers too (earray, carray, vlarray), although only from the integrated blosc2 filter.


FrancescAlted:
Will merge and continue development in the `direct-chunking-blosc2` branch.

PR #962, opened 2022-10-06
Build wheels for 3.11

xmatthias:
As Python 3.11 is comming out by the end of the month - i think pytables should have wheels ready for this.

This PR will add wheel building for 3.11 for all supported platforms.
Tests for the wheels currently run on 3.11-dev (which is 3.11.0rc2 at this moment) - but it's supposed to be near identical to the final version, so i think it'll suffice to test automatically built wheels for now.

It would obviously be great if we could have a new minor release for this (on pypi and other platforms)- as many other projects a already did (Release which simply adds wheels for 3.11).

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -75,6 +75,9 @@ jobs:
            - arch: aarch64
              cibw_python: cp310-*
              os: ubuntu-latest
+           - arch: aarch64
+             cibw_python: cp311-*
+             os: ubuntu-latest
            - arch: arm64
              cibw_python: "cp*"
              os: macos-latest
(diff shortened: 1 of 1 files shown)


FrancescAlted:
This looks good, thanks!  Yes, we definitely want to do a release (including Python 3.11 wheels), by the end of the month, so this is very timely indeed.

FWIW, we would also like to drop support for Python 3.7 (having wheels for 3.8, 3.9, 3.10 and 3.11 is already quite a lot), but that can come later one.

avalentino:
Thanks @xmatthias

heyman:
A new release with Python 3.11 wheels on PyPI would be awesome ❤️

PR #966, opened 2022-10-25
Ci/drop 37 support

xmatthias:
Drop python 37 wheels 
as mentioned in https://github.com/PyTables/PyTables/pull/962#issuecomment-1269962300 - there's only really a point in having wheels for 3.8, 3.9, 3.10 and 3.11 (supported versions of python).

Based on this, this PR will aim to
* Remove CI for 3.6 and 3.7
* Remove wheel building for 3.6 and 3.7
* Update the pypi classifier to match the supported versions
* Update CI actions to avoid action deprecation warnings

About this last point, only [conda-incubator/setup-miniconda ](https://github.com/conda-incubator/setup-miniconda/issues/248) is still left - due to the issue linked above (they didn't release an updated new version so far).

The resulting artifact.zip file will contain the following files (notice that 3.6 and 3.7 are now gone).

screenshot

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -49,11 +49,13 @@ jobs:
           path: dist/*
 
   build_wheels:
-    name: Build ${{ matrix.os }} wheels for ${{ matrix.arch }}
+    name: Build ${{ matrix.os }} ${{ matrix.cibw_pythonn }} wheels for ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     env:
       HDF5_VERSION: 1.12.1
       MACOSX_DEPLOYMENT_TARGET: "10.9"
+      # Skip 3.6 and 3.7 wheels
+      CIBW_SKIP: "cp36-* cp37-*"
     strategy:
       matrix:
         os: [ 'ubuntu-latest', 'macos-latest' ]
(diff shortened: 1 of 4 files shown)


.github/workflows/wheels.yml
```diff
@@ -196,7 +192,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ 'ubuntu-latest', 'windows-latest', 'macos-latest' ]
-        python-version: ['3.7', '3.8', '3.9','3.10', '3.11-dev']
+        python-version: ['3.8', '3.9','3.10', '3.11-dev']
```
FrancescAlted, line 195: Any idea on when we could use plain '3.11' instead of '3.11-dev' ?

FrancescAlted (approved):
This looks good to me.  I just have a minor comment.

.github/workflows/wheels.yml
xmatthias: not really - the way i understood this, the "final" version of python needs to be in https://github.com/actions/python-versions/blob/main/versions-manifest.json.
considering 3.11 stable has only been released yesterday, i assume it'll still take a few days - but then changing to "3.11" is absolutely the right approach!

avalentino (approved):
LGTM

PR #969, opened 2022-10-26
Add Blosc2 filter

oscargm98:
This PR implements Blosc2 filter support. For more details see: [RELEASE_NOTES.rst](https://github.com/PyTables/PyTables/blob/direct-chunking-blosc2/RELEASE_NOTES.rst)


FrancescAlted:
We have been trying to use Python 3.11 for the CI GHA workflow, but this does not seem to work yet (we tried using both `3.11` and `3.11-dev` tags).  @xmatthias, any hint on that?

xmatthias:
the `ci.yml` pipeline is not using "regular" python (through actions/setup-python),  but conda through the action `conda-incubator/setup-miniconda`.

3.11-dev is here definitely wrong as version, but 3.11 would've been correct here (that'll be the prior failure).

Looking at a run with 3.11 on master - it looks like it has different problems with version conflicts - 
https://github.com/xmatthias/PyTables/actions/runs/3329444954/jobs/5506671463


At first glance, it looks like some dependency problems  with 3.11 - most likely because some dependencies pin the supported python version to < 3.11.
At first glance, it seems to be `numexpr` - but this might be a dependency that can be "outwaited" - since they released a new version a few hours ago, so it's probably a question of conda catching up to contain the latest version.

FrancescAlted:
Thanks @xmatthias.  So I say let's activate the '3.11' label here and when conda-forge would release numexpr 2.8.4, the CI should magically turn green (hopefully :-).

FrancescAlted:
Ok, after making numexpr to be installed from wheels instead than conda-forge, all Python 3.11 is green now :-)

xmatthias:
> Ok, after making numexpr to be installed from wheels instead than conda-forge, all Python 3.11 is green now :-)

That's quite a neat workaorund! :+1:

t20100:
Hi,

Just to let you know that we started integrating this `blosc2` filter in `hdf5plugin`: https://github.com/silx-kit/hdf5plugin/pull/201
It's still work in progress.

Our plan is to first make a release without it and then make an alpha release with it so people can test it more easily.

t20100:
Some early feedbacks from starting integration of the blosc2 filter in `hdf5plugin` (https://github.com/silx-kit/hdf5plugin/pull/201) and to answer https://github.com/silx-kit/hdf5plugin/pull/201#discussion_r1030169076:

- For the default number of threads, I would use 1 and allow the users to change it if needed (through an env var?). The issue I see with setting it by default to something else than 1 is that it's not a good default when using multiprocessing.
- For the filter `cd_values`, I've just checked and it seems possible to pass an arbitrary number of integers. What about leveraging this and allow to set a pipeline of filters rather than using the same parameters as for blosc1? And for example passing (0, 0, 0, 0, compression, compression level, filter1, filter1_param, filter2, filter2_param, varying number...). It probably does not cover all cases, but would already enable using delta or trunc_prec + bitshuffle. Else this is only available through direct chunk write

FrancescAlted:
> Some early feedbacks from starting integration of the blosc2 filter in `hdf5plugin` ([silx-kit/hdf5plugin#201](https://github.com/silx-kit/hdf5plugin/pull/201)) and to answer [silx-kit/hdf5plugin#201 (comment)](https://github.com/silx-kit/hdf5plugin/pull/201#discussion_r1030169076):
> 
> * For the default number of threads, I would use 1 and allow the users to change it if needed (through an env var?). The issue I see with setting it by default to something else than 1 is that it's not a good default when using multiprocessing.

That's a good suggestion.  We have just implemented support for BLOSC_* env vars (in particular BLOSC_NTHREADS) for the super-chunk API in C-Blosc2 2.5.0, and are working towards releasing Python-Blosc2 including C-Blosc2 2.5.0.  Once this would be there, we can proceed to make the default number of threads to 1 and control that via BLOSC_NTHREADS.

> * For the filter `cd_values`, I've just checked and it seems possible to pass an arbitrary number of integers. What about leveraging this and allow to set a pipeline of filters rather than using the same parameters as for blosc1? And for example passing (0, 0, 0, 0, compression, compression level, filter1, filter1_param, filter2, filter2_param, varying number...). It probably does not cover all cases, but would already enable using delta or trunc_prec + bitshuffle. Else this is only available through direct chunk write.

Good to know and agreed.  Let's start with something simple first and let's see how we can accommodate more parameters in the future.

FrancescAlted:
We think this is mostly ready to be merged.  If nobody opposes, we will merge this and start the release process during the next week.

doc/source/usersguide/installation.rst
```diff
@@ -50,24 +50,27 @@ Prerequisites
...
-* HDF5_ >= 1.8.4 (>=1.8.15 is strongly recommended)
+* HDF5_ >= 1.10.5
 * NumPy_ >= 1.19.0
 * Numexpr_ >= 2.6.2
 * Cython_ >= 0.29.21
-* c-blosc_ >= 1.4.1 (sources are bundled with PyTables sources but the user can
+* c-blosc_ >= 1.11.1 (sources are bundled with PyTables sources but the user can
   use an external version of sources using the :envvar:`BLOSC_DIR` environment
   variable or the `--blosc` flag of the :file:`setup.py`)
+* blosc2_ >= 0.3.1 This is the Python wheel containing *both* the C-Blosc2 libs
```
avalentino, line 60: Is this correct?
In other places I see blosc2>=0.6.1

avalentino (approved):
It is mostly OK for me.
Just a minor comment.

doc/source/usersguide/installation.rst
FrancescAlted: Oops.  Fixed.

requirements.txt
```diff
@@ -2,4 +2,8 @@
 # doc/source/usersguide/installation.rst
 numpy>=1.19.0
 numexpr>=2.6.2
+# blosc2 wheel is actually only needed when compiling on conda envs.
+# Otherwise, lib comes bundled in PyTables wheels (but it doesn't hurt either).
+blosc2>=0.6.1
 packaging
+py-cpuinfo>=9.0.0
```
avalentino, line 9: `cpuinfo` should be also added to the `pyproject.toml` if we no longer vendor it.

PR #972, opened 2022-11-02
Bump HDF5 to v1.12.2

matham:
- Bumps HDF5 to v1.12.2 and includes the required patch files in the repo instead of downloading it.
- It also removes some unneeded patches and env variables.
- Also spruced up the python versions used in the CI when testing or the baseline python.


matham:
Closing for https://github.com/PyTables/PyTables/pull/973 because I didn't name the branch correctly so the wheels CI did not run.

PR #973, opened 2022-11-02
Bump HDF5 to v1.12.2

matham:
- Bumps HDF5 to v1.12.2 and includes the required patch files in the repo instead of downloading it.
- It also removes some unneeded patches and env variables.
- Also spruced up the python versions used in the CI when testing or the baseline python.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python: ['3.8', '3.9', '3.10']
+        python: ['3.8', '3.9', '3.10', '3.11']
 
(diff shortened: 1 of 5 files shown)


matham:
There was an issue with the [mac wheels](https://github.com/matham/PyTables/actions/workflows/wheels.yml) because github recently switched their mac-latest image from mac-12 to mac-13 and the last snappy release doesn't build with the gcc tools in mac-13. So I removed Snappy as it doesn't seem to be used by hdf5 as far as I can tell. I'm not sure why I originally thought it was needed.

FrancescAlted (approved):
This looks good to me.  I am not sure why we need such a recent HDF5 version (1.12.2), but I am in favor of that anyway.

avalentino (approved):
LGTM

Issue #940, opened 2022-03-01
HDF5 version problems (Invalid version)


fumoffu947:
Hi,

I am trying to install and use pytables, but I encounter a small problem with HDF5.
I have followed the installation recommendation:
```sudo apt-get install libhdf5-serial-dev```
Which is version 1.10.0-patch1.

Then install tables:
```python3 -m pip install tables```
```
Collecting tables
  Using cached ...
Collecting numexpr>=2.6.2 (from tables)
  Downloading ...
    100% |████████████████████████████████| 102kB 709kB/s
Collecting numpy>=1.19.0 (from tables)
  Downloading ...
    100% |████████████████████████████████| 11.4MB 218kB/s
Collecting packaging (from tables)
  Downloading ...
    100% |████████████████████████████████| 40kB 943kB/s
Collecting pyparsing!=3.0.5,>=2.0.2 (from packaging->tables)
  Downloading ...
    100% |████████████████████████████████| 102kB 1.1MB/s
Building wheels for collected packages: tables, numexpr, numpy
  Running setup.py bdist_wheel for tables ... done
  Stored in directory: /home/.../.cache/pip/wheels/1c/6f/2b/...
  Running setup.py bdist_wheel for numexpr ... done
  Stored in directory: /home/.../.cache/pip/wheels/d8/14/d3/...
  Running setup.py bdist_wheel for numpy ... done
  Stored in directory: /home/.../.cache/pip/wheels/7f/2f/9a/...
Successfully built tables numexpr numpy
Installing collected packages: numpy, pyparsing, packaging, numexpr, tables
Successfully installed numexpr-2.8.1 numpy-1.22.2 packaging-21.3 pyparsing-3.0.7 tables-3.7.0
```

Then wen running the tests it fails with:
```
$ python3.8 -m tables.test.test_all
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/lib/python3.8/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "/home/.../.local/lib/python3.8/site-packages/tables/__init__.py", line 53, in <module>
    from .tests import print_versions, test
  File "/home/.../.local/lib/python3.8/site-packages/tables/tests/__init__.py", line 9, in <module>
    from tables.tests.common import print_versions
[...]

PR #974, opened 2022-11-12
Fix for HDF5 version problems (Invalid version) (#940)

heyman:
PR for #940 as suggested here: https://github.com/PyTables/PyTables/issues/940#issuecomment-1062793552

Decided to use [packaging.version.parse()](https://packaging.pypa.io/en/stable/version.html#packaging.version.parse) since it does what was suggested in the comment.

I had the same exception seen in #940, and this change fixed it for me.

### DIFF
--- a/tables/tests/common.py
+++ b/tables/tests/common.py
@@ -8,7 +8,7 @@ import platform
 import tempfile
 from pathlib import Path
 from time import perf_counter as clock
-from packaging.version import Version
+from packaging import version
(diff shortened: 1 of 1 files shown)


heyman:
I added the missing import that should hopefully fix the broken tests.

avalentino (approved):
LGTM

avalentino:
Thanks @heyman

PR #975, opened 2022-11-17
Feature/simplify sdist

avalentino:
Simplify the source distribution according to what agreed in https://github.com/PyTables/PyTables/issues/968#issuecomment-1311355237:

* drop cython generated files
* drop pre-built HTML docs


FrancescAlted (approved):
Looks good!

avalentino:
Thanks @FrancescAlted

PR #979, opened 2022-12-12
Tables 3.8.0

FrancescAlted:
After fixing several issues related with a bad location of the libblosc2 library in different platforms, I have started this PR for releasing 3.8.0.

Right now, we are able to produce wheels (see [current artifacts](https://github.com/PyTables/PyTables/actions/runs/3674358095)) that are able to install and run the test suite in both Linux and Mac OSX.  However, it looks like the wheels for Windows are not able to load the extensions (I suspect that I am not including libblosc2 properly here); as my access to Win machines is quite restricted, I'd appreciate help from somebody more knowledgeable.

Also, I have [disabled the generation of wheels for arm64 and macosx](https://github.com/PyTables/PyTables/commit/df035fb0c15088ad597fdfc716cf04562e274c12) because it was [not compiling correctly](https://github.com/PyTables/PyTables/actions/runs/3665864556/jobs/6197221894).  I think this is due to the fact that the current build mechanism in setup.py is not aware of Mac on arm64, and I don't have cycles enough to address this right now.  As Mac OSX can emulate x86 perfectly on arm64 platforms, I consider this a minor issue, and propose to go ahead without that.  But it would be great if anybody wants to dig more into this.

I am currently pointing at having a release during this week, provided that we can fix the issue with the Windows platform.

@matham @avalentino @tomkooij @eumiro


avalentino:
> BTW, I have also disabled the generation of Win32 wheels in favor of Win64 only. For a rational, see this [discussion](https://discuss.scientific-python.org/t/releasing-or-not-32-bit-windows-wheels/282/8). I fully agree with the reasons argued there, so I am in favor of discontinuing Win32 wheels. Sounds that good?

OK for me

avalentino:
Regarding the support for arm64 on OSX we had a lot of requests for it.
Unfortunately I do not have the HW to test.

xmatthias:
It seems like the last comleted ci run on master (from 3 hours ago) did have (successful) wheels built for arm64 on macos  (this run: https://github.com/PyTables/PyTables/actions/runs/3696416831/jobs/6260085086) - so i wonder if that's really the case.

Was this fixed in the meantime on master?

FrancescAlted:
> > Also, I have [disabled the generation of wheels for arm64 and macosx](https://github.com/PyTables/PyTables/commit/df035fb0c15088ad597fdfc716cf04562e274c12) because it was [not compiling correctly](https://github.com/PyTables/PyTables/actions/runs/3665864556/jobs/6197221894).
> 
> It seems like the last comleted ci run on master (from 3 hours ago) did have (successful) wheels built for arm64 on macos (this run: https://github.com/PyTables/PyTables/actions/runs/3696416831/jobs/6260085086) - so i wonder if that's really the case.
> 
> Was this fixed in the meantime on master?

Yes. I think upgrading to latest Blosc sources did the trick. We have wheels for the major platforms now, so this is getting in good shape.

FrancescAlted:
> > > > Also, I have [disabled the generation of wheels for arm64 and macosx](https://github.com/PyTables/PyTables/commit/df035fb0c15088ad597fdfc716cf04562e274c12) because it was [not compiling correctly](https://github.com/PyTables/PyTables/actions/runs/3665864556/jobs/6197221894).
> > > 
> > > 
> > > It seems like the last comleted ci run on master (from 3 hours ago) did have (successful) wheels built for arm64 on macos (this run: https://github.com/PyTables/PyTables/actions/runs/3696416831/jobs/6260085086) - so i wonder if that's really the case.
> > > Was this fixed in the meantime on master?
> > 
> > 
> > Yes. I think upgrading to latest Blosc sources did the trick. We have wheels for the major platforms now, so this is getting in good shape.
> 
> I was able to install tables from the latest artifacts. But it is not working on my M1 machine. I think compiled blosc2 doesn't have arm64 binaries. Error is basically: `libblosc2.dylib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))`
> 
> ```
> OSError: dlopen(/Temp/.env/lib/python3.10/site-packages/tables/libblosc2.dylib, 0x0006): tried: '/Temp/.env/lib/python3.10/site-packages/tables/libblosc2.dylib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64')), '/System/Volumes/Preboot/Cryptexes/OS/Temp/.env/lib/python3.10/site-packages/tables/libblosc2.dylib' (no such file), '/Temp/.env/lib/python3.10/site-packages/tables/libblosc2.dylib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))
> ```

I have been able to fix the issues with the Python-Blosc2 wheels, and they work fine for M1 Mac now.  However, when I applied the same configuration to PyTables, I keep getting errors:

```
bash-3.2$ python -m tables.tests.test_all.py 
  File "/Users/faltet/miniconda3-arm64/lib/python3.9/runpy.py", line 188, in _run_module_as_main
  File "/Users/faltet/miniconda3-arm64/lib/python3.9/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/Users/faltet/miniconda3-arm64/lib/python3.9/site-packages/tables/__init__.py", line 32, in <module>
ImportError: dlopen(/Users/faltet/miniconda3-arm64/lib/python3.9/site-packages/tables/utilsextension.cpython-39-darwin.so, 0x0002): symbol not found in flat namespace '_H5E_CALLBACK_g'
```

You many want to test the [latest wheel artifacts](https://github.com/PyTables/PyTables/actions/runs/3712304400) and see how it goes, but I don't think arm64 wheels work on any M1 Mac. 

However, I have tested that installing from sources (via e.g. `python -m pip install tables-3.7.1.dev0.tar.gz`) works well on my M1 Mac.   Frankly, I already used way more time than expected here, so unless somebody would suggest a fix, I'd go ahead a release PyTables without support for the MacOSX arm64 binaries.

Opinions?

aemr3:
> 
> I have been able to fix the issues with the Python-Blosc2 wheels, and they work fine for M1 Mac now. However, when I applied the same configuration to PyTables, I keep getting errors:
> 
> ```
> bash-3.2$ python -m tables.tests.test_all.py 
> Traceback (most recent call last):
>   File "/Users/faltet/miniconda3-arm64/lib/python3.9/runpy.py", line 188, in _run_module_as_main
>     mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
>   File "/Users/faltet/miniconda3-arm64/lib/python3.9/runpy.py", line 111, in _get_module_details
>     __import__(pkg_name)
>   File "/Users/faltet/miniconda3-arm64/lib/python3.9/site-packages/tables/__init__.py", line 32, in <module>
>     from .utilsextension import get_hdf5_version as _get_hdf5_version
> ImportError: dlopen(/Users/faltet/miniconda3-arm64/lib/python3.9/site-packages/tables/utilsextension.cpython-39-darwin.so, 0x0002): symbol not found in flat namespace '_H5E_CALLBACK_g'
> ```

I was able to pass this error with setting HDF5_DIR=/opt/homebrew/Cellar/hdf5/1.12.2_2 environment variable. I compiled from sources after that. And tests were running. 

> 
> You many want to test the [latest wheel artifacts](https://github.com/PyTables/PyTables/actions/runs/3712304400) and see how it goes, but I don't think arm64 wheels work on any M1 Mac.
> 
> However, I have tested that installing from sources (via e.g. `python -m pip install tables-3.7.1.dev0.tar.gz`) works well on my M1 Mac. Frankly, I already used way more time than expected here, so unless somebody would suggest a fix, I'd go ahead a release PyTables without support for the MacOSX arm64 binaries.
> 
> Opinions?

Tried latest artifacts again. Same error there. Tables don't have arm64 libraries. But they are available in blosc2 wheel as you said.

```
(.env) ➜  artifact file ./.env/lib/libblosc2.dylib
./.env/lib/libblosc2.dylib: Mach-O 64-bit dynamically linked shared library arm64
(.env) ➜  artifact file ./.env/lib/python3.10/site-packages/tables/libblosc2.dylib
./.env/lib/python3.10/site-packages/tables/libblosc2.dylib: Mach-O 64-bit dynamically linked shared library x86_64
```

So maybe if we include arm64 library in the tables wheel and pass HDF5_DIR environment to the installation we can solve this problem.

xmatthias:
> So maybe if we include arm64 library in the tables wheel and pass HDF5_DIR environment to the installation we can solve this problem.

the idea of a wheel is that there is no requirement to have other system dependencies installed.
requiring a native installation of hdf5 (which is where you point `HDF5_DIR` to) therefore defeats the purpose of having a wheel (which suggests otherwise).

aemr3:
> the idea of a wheel is that there is no requirement to have other system dependencies installed. requiring a native installation of hdf5 (which is where you point `HDF5_DIR` to) therefore defeats the purpose of having a wheel (which suggests otherwise).

I meant it to be done in CI system which looks like it’s already there.

FrancescAlted:
> > the idea of a wheel is that there is no requirement to have other system dependencies installed. requiring a native installation of hdf5 (which is where you point `HDF5_DIR` to) therefore defeats the purpose of having a wheel (which suggests otherwise).
> 
> I meant it to be done in CI system which looks like it’s already there.

Exactly.  The thing is that the wheels for MacOSX arm64 are done exactly in the [same way for both x86_64 and arm64](https://github.com/PyTables/PyTables/blob/master/.github/workflows/wheels.yml#L120).  The fact that they work on Mac Intel, but not on Mac M1 makes me think that something is not right in the CIBuildWheel library (which, BTW, works great for python-blosc2).

At any rate, in case we cannot fix that, and as said before, it is still fortunate that PyTables wheels can be built flawlessly in an Mac arm64, so this should be not a show stopper.

xmatthias:
@aemr3 would you mind testing the latest wheels? https://github.com/PyTables/PyTables/actions/runs/3712304400

blosc sources were updated to the fixed version - so i suspect / hope this does fix it for PyTables, too.

aemr3:
> @aemr3 would you mind testing the latest wheels? https://github.com/PyTables/PyTables/actions/runs/3712304400
> 
> blosc sources were updated to the fixed version - so i suspect / hope this does fix it for PyTables, too.

Unfortunately it didn’t work. I have tested them 3 days ago. I have tried to fix the issues on a fork but it didn’t work too. Github ci is too slow to work actively on it. Currently I don’t have an intel mac to replicate ci build process but working on it. Once I get access to an intel mac I will try my best to fix the issues. 

In the meantime I think @FrancescAlted can release the version, I don’t want to block the release. Maybe we could also get help from the community in the coming days.

matham:
Sorry for not getting back for a bit, I have been sick. I'm still trying to understand the exact changes and what the issue is. But I looked around a little and I see that the python `blosc2` package packages the required dll/.dylib/.so and that is included in our wheels.

I wonder, since `blosc2` is a direct requirement, why do we need to repackage it instead of letting hdf5 get the extension binary from them? The exact installed path where they install the binaries should be easily accessible from python.

Secondarily. without fully understanding the issue on mac yet, could we not get the arm64 and x86_64 binaries from the blosc2 package, merge them into a fat binary and then distribute that? Then the OS should correctly pick the right arch.

matham:
On second thought, my PR alone won't fix the issue. That's because when pip installs blosc2, from which we get the binary, pip installs it for the host arch, which is x86_64 and not arm64. So that's the binary that we repackage as well.

Instead we need to do `pip install blosc2>=0.6.3 --platform macosx_11_0_arm64 --only-binary=:all:` on arm64. Or, long term I do think it's better if we don't repackage the binary and instead let blsoc2 install it and we can get it from there.

FrancescAlted:
Ok, I double checked that your latest build includes x86 binaries for blosc2, not arm64.  So I wonder if I am doing something incorrect by using this hack: https://github.com/PyTables/PyTables/blob/master/setup.py#L839-L848
In principle, when this runs on arm64, it should copy the correct arm64 library from the blosc2 wheel, but apparently it is not.

FrancescAlted:
A curious thing that I noticed is that, when building wheels for MacOSX arm64, all the required wheels (and not only the blosc2 ones) choose the x86_64 arch instead (see section "Running before_build..." in https://github.com/PyTables/PyTables/actions/runs/3740325873/jobs/6349927213).  No idea why.

matham:
> In principle, when this runs on arm64, it should copy the correct arm64 library from the blosc2 wheel, but apparently it is not.


> when building wheels for MacOSX arm64, all the required wheels (and not only the blosc2 ones) choose the x86_64 arch instead

Right, exactly. Pip will always install the wheels for the host arch. The host arch on github CI is x86_64 because github does not offer arm64 machines. Therefore, pip downloads and installs x86_64 wheels, from which we get the blosc2 binary and repackage into our arm64 wheel that we're trying to build.


We basically have two options:
1. Don't repackage the blosc2 binary at all, instead get it directly at runtime from the python blosc2 package. To do this we'd need to tie to an exact major/minor blosc2 version so it matches whatever we compiled against.
2. Use the manual `pip install blosc2>=0.6.3 --platform macosx_11_0_arm64 --only-binary=:all: --target xxx` command to download and install the correct target wheel and binary (arm64) and package that.

I feel like 1 is a more appropriate approach because we can just let `blosc2` handle the building and distributing of the binaries as they currently already do, unless there are reasons we need to package. But option 2 will also work, just needs a little messing about in the code.

matham:
Here's a test of how option 1 would work. I installed tables and uninstalled blosc2, and things work:

```sh
(venv) me263-2:Downloads matte$ python3
Python 3.10.9 (v3.10.9:1dd9be6584, Dec  6 2022, 14:37:36) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import tables
>>> tables.which_lib_version("blosc2")
(('2.6.1.dev', '$Date:: 2022-12-08 #$'), '2.6.1.dev', '$Date:: 2022-12-08 #$')
```

Then I deleted the tables bundled `libosc2.dylib` and tables fails:
```sh
(venv) me263-2:Downloads matte$ python3
Python 3.10.9 (v3.10.9:1dd9be6584, Dec  6 2022, 14:37:36) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import tables
  File "/Users/matte/Downloads/venv/lib/python3.10/site-packages/tables/__init__.py", line 26, in <module>
    cdll.LoadLibrary(blosc2_lib)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ctypes/__init__.py", line 452, in LoadLibrary
    return self._dlltype(name)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(libblosc2.dylib, 6): image not found

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/Users/matte/Downloads/venv/lib/python3.10/site-packages/tables/__init__.py", line 28, in <module>
    cdll.LoadLibrary(os.path.join(current_dir, blosc2_lib))
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ctypes/__init__.py", line 452, in LoadLibrary
    return self._dlltype(name)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(/Users/matte/Downloads/venv/lib/python3.10/site-packages/tables/libblosc2.dylib, 6): image not found
```

So I installed blosc2 and made sure its dylib was on the path (manually) and it works again:
```sh
(venv) me263-2:Downloads matte$ pip install blosc2
Collecting blosc2
  Using cached blosc2-0.6.6-cp310-cp310-macosx_10_9_x86_64.whl (3.9 MB)
Requirement already satisfied: msgpack in ./venv/lib/python3.10/site-packages (from blosc2) (1.0.4)
Installing collected packages: blosc2
Successfully installed blosc2-0.6.6
(venv) me263-2:Downloads matte$ export DYLD_FALLBACK_LIBRARY_PATH=venv/lib
(venv) me263-2:Downloads matte$ python3
Python 3.10.9 (v3.10.9:1dd9be6584, Dec  6 2022, 14:37:36) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import blosc2
>>> import tables
>>> tables.which_lib_version("blosc2")
(('2.6.1.dev', '$Date:: 2022-12-08 #$'), '2.6.1.dev', '$Date:: 2022-12-08 #$')
>>> 
```

We could set the path appropriately automatically when blosc2 is initialized if we were happy with relying on blosc2.

Otherwise, I could look into option 2.

FrancescAlted:
> > In principle, when this runs on arm64, it should copy the correct arm64 library from the blosc2 wheel, but apparently it is not.
> 
> > when building wheels for MacOSX arm64, all the required wheels (and not only the blosc2 ones) choose the x86_64 arch instead
> 
> Right, exactly. Pip will always install the wheels for the host arch. The host arch on github CI is x86_64 because github does not offer arm64 machines. Therefore, pip downloads and installs x86_64 wheels, from which we get the blosc2 binary and repackage into our arm64 wheel that we're trying to build.
> 
> We basically have two options:
> 
> 1. Don't repackage the blosc2 binary at all, instead get it directly at runtime from the python blosc2 package. To do this we'd need to tie to an exact major/minor blosc2 version so it matches whatever we compiled against.
> 2. Use the manual `pip install blosc2>=0.6.3 --platform macosx_11_0_arm64 --only-binary=:all: --target xxx` command to download and install the correct target wheel and binary (arm64) and package that.
> 
> I feel like 1 is a more appropriate approach because we can just let `blosc2` handle the building and distributing of the binaries as they currently already do, unless there are reasons we need to package. But option 2 will also work, just needs a little messing about in the code.

Ok.  But we need the libblosc2.dylib* distributed in the blosc2 package to compile PyTables, so unless I am missing something, we probably need to go in the 2 direction.  Perhaps something like:

```
  - name: Install an arm64 version of the blosc2 library for macosx
    if: ${{ matrix.os == "macosx-latest" && matrix.arch == "arm64" }}
    run: |
      pip install blosc2>=0.6.6 --platform macosx_11_0_arm64 --only-binary=:all: --target /tmp/hdf5 --update`
```

should work?
[Although even if that works, it looks like a bit fragile, as there is a reference to the macosx version (11) there.  Ugly...]

FrancescAlted:
> Here's a test of how option 1 would work. I installed tables and uninstalled blosc2, and things work:
> 
> ```shell
> (venv) me263-2:Downloads matte$ python3
> Python 3.10.9 (v3.10.9:1dd9be6584, Dec  6 2022, 14:37:36) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import tables
> >>> tables.which_lib_version("blosc2")
> (('2.6.1.dev', '$Date:: 2022-12-08 #$'), '2.6.1.dev', '$Date:: 2022-12-08 #$')
> ```
> 
> Then I deleted the tables bundled `libosc2.dylib` and tables fails:
> 
> ```shell
> (venv) me263-2:Downloads matte$ python3
> Python 3.10.9 (v3.10.9:1dd9be6584, Dec  6 2022, 14:37:36) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import tables
> Traceback (most recent call last):
>   File "/Users/matte/Downloads/venv/lib/python3.10/site-packages/tables/__init__.py", line 26, in <module>
>     cdll.LoadLibrary(blosc2_lib)
>   File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ctypes/__init__.py", line 452, in LoadLibrary
>     return self._dlltype(name)
>   File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ctypes/__init__.py", line 374, in __init__
>     self._handle = _dlopen(self._name, mode)
> OSError: dlopen(libblosc2.dylib, 6): image not found
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/Users/matte/Downloads/venv/lib/python3.10/site-packages/tables/__init__.py", line 28, in <module>
>     cdll.LoadLibrary(os.path.join(current_dir, blosc2_lib))
>   File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ctypes/__init__.py", line 452, in LoadLibrary
>     return self._dlltype(name)
>   File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ctypes/__init__.py", line 374, in __init__
>     self._handle = _dlopen(self._name, mode)
> OSError: dlopen(/Users/matte/Downloads/venv/lib/python3.10/site-packages/tables/libblosc2.dylib, 6): image not found
> ```
> 
> So I installed blosc2 and made sure its dylib was on the path (manually) and it works again:
> 
> ```shell
> (venv) me263-2:Downloads matte$ pip install blosc2
> Collecting blosc2
>   Using cached blosc2-0.6.6-cp310-cp310-macosx_10_9_x86_64.whl (3.9 MB)
> Requirement already satisfied: msgpack in ./venv/lib/python3.10/site-packages (from blosc2) (1.0.4)
> Installing collected packages: blosc2
> Successfully installed blosc2-0.6.6
> (venv) me263-2:Downloads matte$ export DYLD_FALLBACK_LIBRARY_PATH=venv/lib
> (venv) me263-2:Downloads matte$ python3
> Python 3.10.9 (v3.10.9:1dd9be6584, Dec  6 2022, 14:37:36) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import blosc2
> >>> import tables
> >>> tables.which_lib_version("blosc2")
> (('2.6.1.dev', '$Date:: 2022-12-08 #$'), '2.6.1.dev', '$Date:: 2022-12-08 #$')
> >>> 
> ```
> 
> We could set the path appropriately automatically when blosc2 is initialized if we were happy with relying on blosc2.

Hey, this looks pretty cool actually.  So, as far as I understand this, tables would come with its own version of Blosc2 C library, and even if you put manually another version in the dylib path first, this continue to work?  I'd buy it indeed.  Please send a PR.

> 
> Otherwise, I could look into option 2.

What you achieved for option 1 looks fine to me already.  Which advantage would have option 2 then?

matham:
>So, as far as I understand this, tables would come with its own version of Blosc2 C library, and even if you put manually another version in the dylib path first, this continue to work?

It's not quite that. Tables at build time will have installed the python blosc2 package. This is what it builds against. However, we won't copy the blosc2 binary from that package and re-package it in our wheel. Instead, we only have to make sure blosc2 is installed at runtime and that we can use the same binary we built against (or a compatible with same major/minor version) to run against. This allows us to completely offload the blosc2 binary to the python blosc2 package. They correctly packaged already for arm/x86 so it wouldn't require any work from our end to maintain blosc2.

Please see https://github.com/PyTables/PyTables/pull/983.

FrancescAlted:
Ok.  This looks good.  Let's the party begin.

PR #982, opened 2022-12-20
Run osx arm64 and x86 builds in different containers to prevent binary clash

matham:
See https://github.com/PyTables/PyTables/pull/979.


matham:
https://github.com/matham/PyTables/actions/runs/3736441629 should generate the wheels to test.

FrancescAlted (approved):

PR #983, opened 2022-12-20
Use the blosc2 python package for the blosc2 binary

matham:
The goal here is for pytables to not package at all the blosc2.dll/so/dylib in our wheel, because we end up repackaging the wrong arch. Instead, we will rely on blosc2's wheel to provide the binary and we made blosc2 a hard dependency. Then, at runtime, using the package metadata we find the blosc2.dll/so/dylib and load it as before. 

Changes:

* It uses `importlib.metadata` to locate the `blosc2` binaries and header files in setup.py so we don't have to manually search for them in the RECORD.
* Similarly it uses it to locate the binary when tables is imported and loads it into memory. This way, we don't package it all and instead rely on `blosc2` providing the binary.
* Changed the version specifier from `blosc2>=0.6.6` to `blosc2~=0.6.6`. This ensures that users install the same major/minor version of `blosc2` that tables was built against to prevent forward incompatibility.
* Instead of manually listing all the deps in the CI for each install, or using requirements.txt, I've added the deps as `required_deps`. This will ensure users get the required blosc2 and other libraries when installing normally with pip.

  This would potentially only be an issue on conda, where we want to get the deps from conda, not pip. But already, we manually install deps in conda before installing tables, so this should not be an issue.

  In fact, we can remove the explicit deps list in the dev guide and instead tell people, get the deps listed in pyproject.toml/setup.cfg so that we don't have to list them there as well.

  The only reason for having a requirement.txt at all in the current scheme is when installing using the makefile, which doesn't use pip.
* To prevent the wheel repair from choking on the blosc2 dlls and to make sure they are not included in our wheel (so that we don't get the wrong arch in our wheel), I explicitly excluded them. For windows and linux there's option in the auditwheel/delvewheel for this. There isn't currently an option for mac so I had to rely on `--ignore-missing-dependencies` for now. Eventually we should use https://github.com/matthew-brett/delocate/pull/106 instead once merged.

Here's the wheels being generatedl: https://github.com/matham/PyTables/actions/runs/3744274349.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,6 @@ requires = [
     "packaging",
     "py-cpuinfo",
     "Cython >=0.29.21",
-    "blosc2 >=0.6.6"
+    "blosc2 ~=0.6.6"
 ]
 build-backend = "setuptools.build_meta"
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,10 +45,12 @@ project_urls =
 [options]
 python_requires = >=3.8
 zip_safe = False
-# install_requires =
-#     numpy>=1.19.0
-#     numexpr>=2.6.2
-#     packaging
+install_requires =
+    numpy>=1.19.0
+    numexpr>=2.6.2
+    blosc2~=0.6.6
+    packaging
+    py-cpuinfo
 packages = find:
 include_package_data = True
 
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ import re
 from setuptools import setup, Extension
 from setuptools.command.build_ext import build_ext
 import pkg_resources
+from importlib.metadata import files as import_metadata_files
 from packaging.version import Version
 
 
@@ -116,22 +117,25 @@ def get_blosc2_directories():
         raise EnvironmentError(
             "Cannot find neither the c-blosc2 package nor the "
             "python-blosc2 wheel")
-    version = blosc2.__version__
-    basepath = Path(os.path.dirname(blosc2.__file__))
-    recinfo = basepath.parent / f'blosc2-{version}.dist-info' / 'RECORD'
-    library_path = None
-    for line in open(recinfo):
-        if 'libblosc2' in line:
-            library_path = Path(os.path.abspath(basepath.parent /
-                                                Path(line[:line.find('libblosc2')])))
-            break
-    if not library_path:
-        raise NotADirectoryError("Library directory not found for blosc2!")
 
-    include_path = library_path.parent / 'include'
-    if not os.path.isfile(include_path / 'blosc2.h'):
-        install_path = os.path.abspath(library_path.parent)
-        raise NotADirectoryError("Install directory for blosc2 not found in %s" % install_path)
+    files = import_metadata_files('blosc2')
+    library_path = None
+    include_path = None
+
+    for f in files:
+        if f.name in {"libblosc2.dll", "libblosc2.so", "libblosc2.dylib"}:
+            library_path = str(f.locate().resolve().parent)
+            break
+
+    for f in files:
+        if f.name == "blosc2.h":
+            include_path = str(f.locate().resolve().parent)
+            break
+
+    if library_path is None:
+        raise NotADirectoryError("Library directory not found for blosc2")
+    if include_path is None:
+        raise NotADirectoryError("Include directory not found for blosc2")
 
     return include_path, library_path
 
@@ -821,40 +825,7 @@ if __name__ == "__main__":
                 "nt": "any of the directories in %%PATH%%",
             }[os.name]
 
-            if package.name == "blosc2":
-                # We will copy this into the tables directory
-                print("  * Copying blosc2 runtime library to 'tables' dir"
-                      " because it was not found in standard locations")
-                platform_system = platform.system()
-                if platform_system == "Linux":
-                    copy_libs += ['libblosc2.so']
-                    dll_dir = '/tmp/hdf5/lib'
-                    # Copy dlls when producing the wheels in CI
-                    if "bdist_wheel" in sys.argv and os.path.exists(dll_dir):
-                        shared_libs = glob.glob(str(libdir) + '/libblosc2.so*')
-                        for lib in shared_libs:
-                            shutil.copy(lib, dll_dir)
-                    else:
-                        shutil.copy(libdir / 'libblosc2.so', 'tables')
-                elif platform_system == "Darwin":
-                    copy_libs += ['libblosc2.dylib']
-                    dll_dir = '/tmp/hdf5/lib'
-                    # Copy dlls when producing the wheels in CI
-                    if "bdist_wheel" in sys.argv and os.path.exists(dll_dir):
-                        shared_libs = glob.glob(str(libdir) + '/libblosc2*.dylib')
-                        for lib in shared_libs:
-                            shutil.copy(lib, dll_dir)
-                    else:
-                        shutil.copy(libdir / 'libblosc2.dylib', 'tables')
-                else:
-                    copy_libs += ['libblosc2.dll']
-                    dll_dir = 'C:\\Miniconda\\envs\\build\\Library\\bin'
-                    # Copy dlls when producing the wheels in CI
-                    if "bdist_wheel" in sys.argv and os.path.exists(dll_dir):
-                        shutil.copy(libdir.parent / 'bin' / 'libblosc2.dll', dll_dir)
-                    else:
-                        shutil.copy(libdir.parent / 'bin' / 'libblosc2.dll', 'tables')
-            else:
+            if package.name != "blosc2":
                 if "bdist_wheel" in sys.argv and os.name == "nt":
                     exit_with_error(
                         f"Could not find the {package.name} runtime.",
@@ -868,7 +839,6 @@ if __name__ == "__main__":
                     f"in {loc}. In case of runtime problems, please "
                     f"remember to install it.",
                 )
-
 
         if os.name == "nt":
             # LZO DLLs cannot be copied to the binary package for license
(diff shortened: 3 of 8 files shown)


matham:
I'm not setup for conda, but it would be nice if someone can manually test the generated wheels on conda (assuming it works).

tables/__init__.py
```diff
@@ -8,25 +8,19 @@
...
 
+try:
+    # if blosc2 is not installed, move on
+    import blosc2
+    for f in import_metadata_files('blosc2'):
+        if f.name in {"libblosc2.dll", "libblosc2.so", "libblosc2.dylib"}:
+            cdll.LoadLibrary(str(f.locate().resolve()))
+            break
+    else:
+        raise FileNotFoundError('Unable to locate the blosc2 binary')
```
avalentino, line 21: I think that this is not OK for the cases (like linux distributions) where libblosc2 is typically provided in a standalone package and both python-blosc2 and pytables use the system libraries.
Is there a way to workaround?

setup.cfg
```diff
@@ -45,10 +45,12 @@ project_urls =
...
 python_requires = >=3.8
 zip_safe = False
-# install_requires =
-#     numpy>=1.19.0
-#     numexpr>=2.6.2
-#     packaging
+install_requires =
+    numpy>=1.19.0
+    numexpr>=2.6.2
+    blosc2~=0.6.6
```
avalentino, line 51: Strictly speaking blosc2 (intended as pytthon-blosc2) is not a requirement (cblosc2 is).
In principle cblosc2 could be provided as system library.

This is not a big issue IMHO bur maybe it could make sense to add a small note in the documentation or at least a comment to explain the situation.

setup.py
```diff
@@ -116,22 +117,25 @@ def get_blosc2_directories():
         raise EnvironmentError(
             "Cannot find neither the c-blosc2 package nor the "
             "python-blosc2 wheel")
-    version = blosc2.__version__
-    basepath = Path(os.path.dirname(blosc2.__file__))
-    recinfo = basepath.parent / f'blosc2-{version}.dist-info' / 'RECORD'
+
+    files = import_metadata_files('blosc2')
```
avalentino, line 121: This will make the life very hard for packagers.
Please make the setup script compatible with a scenario in which cblosc2 is provided as system library.

Conda could be used in CI to test this scenario.

pyproject.toml
```diff
@@ -6,6 +6,6 @@ requires = [
     "packaging",
     "py-cpuinfo",
     "Cython >=0.29.21",
-    "blosc2 >=0.6.6"
+    "blosc2 ~=0.6.6"
```
avalentino, line 9: See comment below.
Strictly speaking python-blosc2 is not a build requirement, cblosc2 is.

avalentino (changes requested):

tables/__init__.py
matham: The issue only arises if blosc2 is installed but it uses the system `libblosc2` because then the import will work but `LoadLibrary` will fail. Could I just remove the `FileNotFoundError` exception and not raise an exception at all?

What would be the downside of that? Eventually an error will be raised by python when it tries to load the blosc2 and can't find it, right? In that cause there's no real reason to raise an error here!?

setup.py
matham: But wasn't it the same case before? Previously it directly looked at the record file. All this change is doing is using `importlib.metadata` to inspect the record instead of us doing it manually. Unless I'm missing something!?

matham: Do you mean for a case in which `cblosc2` is installed but not `blosc2`? I think it is not currently supported even before my changes!?

setup.cfg
matham: Ok, I can add a note to the dev guide and to file!

At the moment, even before this PR, it already seems like our wheels have a hard dependency on everything in the https://github.com/PyTables/PyTables/blob/master/requirements.txt file because it's all listed in the record as a dependency. I'm not sure where that is coming from. Does pip automatically look at the requirements file as a list of dependencies?

FrancescAlted: Some context here. By adopting dependency on Python-Blosc2 wheels I wanted to avoid the different issues in supporting plain C-Blosc2 builds. Python-Blosc2 comes with high quality C-Blosc2 binary libs, and we (the Blosc team) plan doing so forever. So, for the sake of simplicity, I'd like to continue using C-Blosc2 *only* via Python-Blosc2.

Regarding the dependency of requirements.txt, I think it was mainly my fault.  I just wanted to be more consistent in all the building yaml files, and decided to go this route. If that is creating issues, we can go back here.

tables/__init__.py
avalentino: Yes, to me it is fine.

setup.cfg
avalentino: As I said having python-blosc2 listed as a requirement is not a big issue for me.
After all it is a small package and the current solution simplifies a lot of stuffs.

I suggested to document it somewhere mostly to help a little bit people that creates packages that are not based on wheels.

setup.py
avalentino: Ah sorry, I'm commenting on what I see in the diff.
There could be some pre-existing issue but please consider that all the blosc2 part is quite new.
In any case we should try to fix.

FrancescAlted:
I have tried the new wheels on my M1 Mac, but no success:

```
(base) Francescs-MacBook-Air:build-arm francesc$ arch
arm64
(base) Francescs-MacBook-Air:build-arm francesc$ python -m pip install ~/Downloads/artifact-7/tables-3.7.1.dev0-cp39-cp39-macosx_11_0_arm64.whl --force-reinstall
Processing /Users/faltet/Downloads/artifact-7/tables-3.7.1.dev0-cp39-cp39-macosx_11_0_arm64.whl
Collecting cython>=0.29.21
  Using cached Cython-0.29.32-py2.py3-none-any.whl (986 kB)
Collecting py-cpuinfo
  Using cached py_cpuinfo-9.0.0-py3-none-any.whl (22 kB)
Collecting numpy>=1.19.0
  Using cached numpy-1.24.0-cp39-cp39-macosx_11_0_arm64.whl (13.9 MB)
Collecting packaging
  Using cached packaging-22.0-py3-none-any.whl (42 kB)
Collecting numexpr>=2.6.2
  Using cached numexpr-2.8.4-cp39-cp39-macosx_11_0_arm64.whl (89 kB)
Collecting blosc2~=0.6.6
  Using cached blosc2-0.6.6-cp39-cp39-macosx_11_0_arm64.whl (3.1 MB)
Collecting msgpack
  Using cached msgpack-1.0.4-cp39-cp39-macosx_11_0_arm64.whl (69 kB)
Installing collected packages: py-cpuinfo, msgpack, packaging, numpy, cython, blosc2, numexpr, tables
  Attempting uninstall: py-cpuinfo
    Found existing installation: py-cpuinfo 9.0.0
    Uninstalling py-cpuinfo-9.0.0:
      Successfully uninstalled py-cpuinfo-9.0.0
  Attempting uninstall: msgpack
    Found existing installation: msgpack 1.0.4
    Uninstalling msgpack-1.0.4:
      Successfully uninstalled msgpack-1.0.4
  Attempting uninstall: packaging
    Found existing installation: packaging 22.0
    Uninstalling packaging-22.0:
      Successfully uninstalled packaging-22.0
  Attempting uninstall: numpy
    Found existing installation: numpy 1.24.0
    Uninstalling numpy-1.24.0:
      Successfully uninstalled numpy-1.24.0
  Attempting uninstall: cython
    Found existing installation: Cython 0.29.32
    Uninstalling Cython-0.29.32:
      Successfully uninstalled Cython-0.29.32
  Attempting uninstall: blosc2
    Found existing installation: blosc2 0.6.6
    Uninstalling blosc2-0.6.6:
      Successfully uninstalled blosc2-0.6.6
  Attempting uninstall: numexpr
    Found existing installation: numexpr 2.8.4
    Uninstalling numexpr-2.8.4:
      Successfully uninstalled numexpr-2.8.4
  Attempting uninstall: tables
    Found existing installation: tables 3.7.1.dev0
    Uninstalling tables-3.7.1.dev0:
      Successfully uninstalled tables-3.7.1.dev0
Successfully installed blosc2-0.6.6 cython-0.29.32 msgpack-1.0.4 numexpr-2.8.4 numpy-1.24.0 packaging-22.0 py-cpuinfo-9.0.0 tables-3.7.1.dev0
(base) Francescs-MacBook-Air:build-arm francesc$ python -m tables.tests.test_all
  File "/Users/faltet/miniconda3-arm64/lib/python3.9/runpy.py", line 188, in _run_module_as_main
  File "/Users/faltet/miniconda3-arm64/lib/python3.9/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/Users/faltet/miniconda3-arm64/lib/python3.9/site-packages/tables/__init__.py", line 26, in <module>
ImportError: dlopen(/Users/faltet/miniconda3-arm64/lib/python3.9/site-packages/tables/utilsextension.cpython-39-darwin.so, 0x0002): symbol not found in flat namespace '_blosc2_cbuffer_sizes'
```

As the modifications here are substantial, and we are not getting results on arm64 Macs, I don't think it is wise to merge this.  So I stick with my proposal of not generating binary wheels for arm64 Mac.  Having said this, I'll go through the suggestions found here and come with some better docs on how the new blosc2 dependency should be handled (probably at the _Installation_ chapter of UG).

tables/__init__.py
bnavigator: Note that linux distributions have a libblosc2.so.2 but not necessarlily a libblosc2.so when there are no developments packages installed.

avalentino:
@matham this PR now has conflicts
Do you plan to rebase it and address Francesc's comments?
Otherwise maybe we should close it.

matham:
I'm not sure I'll have time in the near future to fix stuff, so I'll close and look at opening a new one when I have time.

PR #987, opened 2022-12-29
Allow to use of system blosc2 to build pytables

### LINKED ISSUE #984: Using compiler.has_function() without includes specified breaks clang-16
The upcoming clang-16 release treats implicit function declarations as fatal errors (see https://discourse.llvm.org/t/configure-script-breakage-with-the-new-werror-implicit-function-declaration/65213). As a result, using `compiler.has_function()` from distutils without explicitly specifying necessary includes results in checks failing incorrectly, e.g.:

```
python3.10 setup.py build -j 1 -j1
/tmp/portage/dev-python/pytables-3.8.0/temp/blosc2_list_compressors2avhs2_r.c:2:5: error: call to undeclared function 'blosc2_list_compressors'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    blosc2_list_compressors();
    ^
1 error generated.
[...]
* Could not find blosc2 headers and library; disabling support for it.
```

(this is with blosc2 library being available)

### LINKED ISSUE #986: Please make it possible to use system c-blosc2 installation (without blosc2 Python package)
Currently, `setup.py` forces `blosc2` Python package to be installed and forces blosc2 paths from it. Could you please make this optional and us use system-installed blosc2 like other compression libraries?

To confirm that it works, I have used the following hack locally:

```diff
diff --git a/setup.py b/setup.py
index 966be028..2708deb9 100755
--- a/setup.py
+++ b/setup.py
@@ -262,11 +262,11 @@ if __name__ == "__main__":
 
     debug = "--debug" in sys.argv
 
-    blosc2_inc, blosc2_lib = get_blosc2_directories()
+    blosc2_inc, blosc2_lib = "", ""
 
     # Global variables
     lib_dirs = [blosc2_lib]
-    inc_dirs = [Path("hdf5-blosc/src"), Path("hdf5-blosc2/src"), blosc2_inc]
+    inc_dirs = [Path("hdf5-blosc/src"), Path("hdf5-blosc2/src")]
     optional_libs = []
     copy_libs = []
 
```

avalentino:
Closes #986.
Closes #984.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -262,17 +262,24 @@ if __name__ == "__main__":
 
     debug = "--debug" in sys.argv
 
-    blosc2_inc, blosc2_lib = get_blosc2_directories()
-
     # Global variables
-    lib_dirs = [blosc2_lib]
-    inc_dirs = [Path("hdf5-blosc/src"), Path("hdf5-blosc2/src"), blosc2_inc]
+    lib_dirs = []
+    inc_dirs = [Path("hdf5-blosc/src"), Path("hdf5-blosc2/src")]
     optional_libs = []
     copy_libs = []
 
     default_header_dirs = None
     default_library_dirs = None
     default_runtime_dirs = None
+
+    blosc2_inc = blosc2_lib = None
+    if os.environ.get("PYTABLES_NO_BLOASC2_WHEEL", None) is None:
+        try:
+            blosc2_inc, blosc2_lib = get_blosc2_directories()
+            lib_dirs.append(blosc2_lib)
+            inc_dirs.append(blosc2_inc)
+        except EnvironmentError:
+            pass
 
     if os.name == "posix":
         prefixes = ("/usr/local", "/sw", "/opt", "/opt/local", "/usr", "/")
@@ -691,7 +698,7 @@ if __name__ == "__main__":
     # Try to locate the compulsory and optional libraries.
     lzo2_enabled = False
     compiler = new_compiler()
-    if not BLOSC2_DIR:
+    if not BLOSC2_DIR and blosc2_inc is not None:
         # Inject the blosc2 directory as detected
         BLOSC2_DIR = os.path.dirname(blosc2_inc)
     for (package, location) in [
--- a/setup.py
+++ b/setup.py
@@ -273,7 +273,7 @@ if __name__ == "__main__":
     default_runtime_dirs = None
 
     blosc2_inc = blosc2_lib = None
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -262,18 +262,25 @@ def run(self):
...
+    inc_dirs = [Path("hdf5-blosc/src"), Path("hdf5-blosc2/src")]
     optional_libs = []
     copy_libs = []
 
     default_header_dirs = None
     default_library_dirs = None
     default_runtime_dirs = None
 
+    blosc2_inc = blosc2_lib = None
+    if os.environ.get("PYTABLES_NO_BLOASC2_WHEEL", None) is None:
```
mgorny, line 276: I think you meant:

```suggestion
    if os.environ.get("PYTABLES_NO_BLOSC2_WHEEL", None) is None:
```

mgorny (approved):
Thanks. It seems to work fine for us. LGTM modulo the typo (but I don't know about Conda).

avalentino:
Thanks @mgorny

FrancescAlted:
Looks good to me.  In the future, I'd like to strictly require python-blosc2 for allowing the registration of [user-defined codecs and filters](https://github.com/Blosc/python-blosc2/blob/main/examples/ucodecs-ufilters.ipynb) straight from Python, and hence, accessible from PyTables.  But that will be next year, the soonest ;-)

FrancescAlted (approved):

avalentino:
Thanks @FrancescAlted.
Probably switching to a build system based on cmake or meson would help a lot in our case.
This is something that I had in mind singe very long time, probably it is time to start thinking to it more seriously.

Do you have experience with [scikit-build](https://scikit-build.readthedocs.io) or [meson](https://mesonbuild.com)? ... or any other alternative?

mgorny:
My experience with scikit-build is one of pure horror and revulsion, please don't do that. meson on the other hand can be quite good, especially with [meson-python](https://github.com/mesonbuild/meson-python) PEP517 backend. There are some rough edges, particularly regarding installing pkg-config files and other data that doesn't fit cleanly inside a venv but I don't think any of that applies to PyTables.

avalentino:
@mgorny do you know other alternatives it that could worth to consider?

mgorny:
I could joke autotools but you probably don't want to do that ;-). No, I think meson is the state of art right now, as far as build systems are concerned, and it's getting first-class support from a lot of people right now, so if you hit any problems there's a good chance someone will provide a good solution.

PR #993, opened 2023-01-09
Use python warnings instead of messages to sys.stderr, fixes #992

### LINKED ISSUE #992: Use python's warning system instead of printing to stderr
pytables is a library, as such it is very annoying that warnings are printed to stderr and not using the python standard warnings system.

It makes it much harder to e.g. make tests fail for resource leaks like `Closing remaining open files` which would be trivial if tables would instead of printing to stderr raise a `UnclosedFileWarning` and doing `-Werror::tables.UnclosedFileWarning`

maxnoe:
fixes #992

### DIFF
--- a/tables/attributeset.py
+++ b/tables/attributeset.py
@@ -1,7 +1,6 @@
 """Here is defined the AttributeSet class."""
 
 import re
-import sys
 import warnings
 import pickle
 import numpy as np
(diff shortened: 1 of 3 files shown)


avalentino (approved):

avalentino:
Thanks @maxnoe

Issue #991, opened 2023-01-05
Use H5Dchunk_iter rather than H5Dget_chunk_info for Blosc2


mkitti:
HDF5 1.14 introduces `H5Dchunk_iter` which provides a more efficient way to iterate over all chunks rather than having $O(N^2)$ scaling where $N$ is the number of chunks. HDF5 1.14 was recently released. The function is being backported and is expected in 1.12.3 and 1.10.10.

Docs: https://docs.hdfgroup.org/hdf5/develop/group___h5_d.html#title6
Original pull request: https://github.com/HDFGroup/hdf5/pull/6

cc: @oscargm98 @FrancescAlted

FrancescAlted:
Thanks @mkitti, that's interesting.  I think the advantage in speed should be more evident when chunks are small, which is not the usual case for Blosc2 chunks, but worth considering for next releases anyway.

FrancescAlted:
For the record, I was curious on the kind of overhead HDF5 is adding to the e.g. [inkernel case](https://www.blosc.org/posts/blosc2-pytables-perf/), and it is really low (< 1%) when using Blosc2:

![image](https://user-images.githubusercontent.com/314521/210977268-a01b5990-7b37-46eb-9327-e34c3676acc3.png)

Note that the time is used mainly on reading from the OS filesystem cache and multithread synching (38% above), blosc2 (27%) and finally numexpr (14%).

Also, even using old plain Blosc filter (where the chunksizes are 72x smaller, and queries perfom ~2x slower too), the HDF5 layer is not adding too much overhead either:
￼
![image](https://user-images.githubusercontent.com/314521/210977511-19ef74f6-fd9c-410d-9371-39a09c9a15e8.png)

￼So, IMO we should not be too pressed in optimizing the HDF5 calls.  Having a much larger chunksize (as Blosc2 allows) is a way better path for getting good perfomance (at least for usual PyTables use cases).

mkitti:
How many chunks are in your typical use case with large chunks? This only became an issue for me when I started to scale up to large multi-terabyte datasets.

FrancescAlted:
In the case of the inkernel queries above for Blosc2, the number is < 100 (3.1 GB / 36 MB).  The point is that this means typically 2 orders of magnitude less than using the typical HDF5 single-partition.

[...]

PR #997, opened 2023-01-31
Optimization for iter chunk

FrancescAlted:
Here there is an optimization for using `H5Dchunk_iter`.  This was originally suggested in https://github.com/PyTables/PyTables/issues/991 by @mkitti.

This should be only useful when there is a large number of chunks on a dataset, and one wants to walk all or a large part of it.  Preliminary benchmarks are suggesting that we can get up to 10% of speedup for datasets of 270 GB (7731 chunks when they are 36 MB in size), when using the optimized path for Blosc2.  Probably we can get more speedup for larger datasets (to be tested).

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -52,7 +52,7 @@ jobs:
     name: Build ${{ matrix.os }} wheels for ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     env:
-      HDF5_VERSION: 1.12.2
+      HDF5_VERSION: 1.14.0  # needed for H5Dchunk_iter
       MACOSX_DEPLOYMENT_TARGET: "10.9"
       # Skip 3.6 and 3.7 wheels
       CIBW_SKIP: "*-musllinux_* cp36-* cp37-*"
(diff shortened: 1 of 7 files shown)


avalentino (approved):
LGTM

avalentino:
Thanks @FrancescAlted

oscargm98 (approved):

FrancescAlted:
It is in.

mkitti:
I did some benchmarks for HDF5.jl. The chunk info retrieval time goes from 0.6 seconds with 4K chunks to 13 seconds with 16K chunks when using indexing. With iteration, the retrieval time stays under 0.1 second. This should improve scaling for large datasets.

https://github.com/JuliaIO/HDF5.jl/pull/1031#issuecomment-1407749686

FrancescAlted:
Ah, I finally figured out the main situation where this is useful and filed a new PR: https://github.com/PyTables/PyTables/issues/991#issuecomment-1416243215.  Although I needed to use > 1 million chunks before seeing significant speed-ups, it has been a fun exercise anyway :-)

PR #999, opened 2023-02-03
Optimize for H5Dchunk_iter for random reads

FrancescAlted:
Here it is yet another optimization for doing random reads in combination with the new `H5Dchunk_iter` introduced in HDF5 1.14 series.

The new `bench/read-table.py` script generates a table with 20 *billions* of rows, accounting for more than 2 million of chunks on my machine (as the chunk size is determined automatically based on the L3 size, this figure is machine dependent).  With that, here it is the new random speed when using HDF5 1.14.0:

```
Reading 100.0 Krows...
Time to read: 1.459s
```

which makes for 15 us per read; pretty good.

Using a build with HDF5 1.12.2 (i.e. not having the `H5Dchunk_iter` machinery), I'm getting:

```
Reading 100.0 Krows...
Time to read: 7.302s
```

so, 5x times slower.

As this depends on the square of the number of chunks, the improvement can be larger for even a larger number of chunks --in the suposition that more than 2 million of chunks (that is, more than 20 billions of rows in our case) is a realistic assumption ;-)

### DIFF
--- a/bench/read-table.py
+++ b/bench/read-table.py
@@ -1,0 +1,50 @@
+import tables as tb
+import numpy as np
+from time import time
+import sys
+
+
+N = 200
+NBUNCH = 100_000_000
+NREADS = 100_000
+filename = "read-table.h5"
+
+
+class Particle(tb.IsDescription):
+    lati = tb.Int32Col()
+    longi = tb.Int32Col()
+    pressure = tb.Float32Col()
+    temperature = tb.Float32Col()
+
+if len(sys.argv) > 1 and sys.argv[1] == "w":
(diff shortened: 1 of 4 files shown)


avalentino (approved):
LGTM

FrancescAlted:
This completes the acceleration path for Blosc2/H5Dchunk_iter in both the sequential reads (by up to a 10% or more; implemented in PR #997) and random small reads scenarios --the acceleration here can be way larger (above a 5x speedup has been demonstrated), as the bottleneck is locating the address where the data is; this is where the team Blosc2/H5Dchunk_iter does an excellent job.

PR #1007, opened 2023-03-14
Set github workflow permissions to least privileges

### LINKED ISSUE #1006: Set github workflow permissions to least privileges
I would like to suggest setting the permissions to the github workflows as read only on the [top level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) and any write permission be given at the [run level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions).

This is necessary due to a behavior of github workflow to grant to GITHUB_TOKEN write permissions to all types of permissions, regardless of they being used or not. In case of the workflow getting compromised, an attacker can exploit this permissions.

This can be seen in the Action run step "Set up job" such as https://github.com/PyTables/PyTables/actions/runs/4407964071/jobs/7722340374:
 
image

Thus, it is both a recommendation from [OpenSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and the [Github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) to always use credentials that are minimally scoped.

Let me know if a PR is welcome and I'll submit it as soon as possible. 

Thanks!

Disclosure: I'm from Google, working with the OpenSSF to help many open source projects to increase their supply-chain security.

joycebrum:
### Changes

Closes #1006 

- ubuntu.yml - success run https://github.com/joycebrum/PyTables/actions/runs/4418480233
- ci.yml - success run https://github.com/joycebrum/PyTables/actions/runs/4418480235
- wheels.yml - it is with the same errors https://github.com/joycebrum/PyTables/actions/runs/4418480249. I believe the permissions granted (`contents: read`) will be enough to this workflow, looking at the jobs.

If you have any problem with wheels.yml permission after fixing it feel free to ping me and I'll be happy to help. I also tried to figure out why it is not working but also not sure about it 😥 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: CI
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 jobs:
 
   sdist:
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -2,6 +2,9 @@ name: ubuntu
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-22.04
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -13,6 +13,9 @@ on:
     types:
       - published
 
+permissions:
+  contents: read
+
 jobs:
   build_sdist:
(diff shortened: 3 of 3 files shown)


avalentino:
Thanks @joycebrum

Issue #1003, opened 2023-03-03
Build failures with Cython 3.0.0b1


mgorny:
Cython 3.0.0b1 has been released recently, and pytables fails to build with it:

```pytb
$ python -m build -w
* Creating virtualenv isolated environment...
* Installing packages in isolated environment... (Cython >=3.0.0b1, blosc2 ~=2.0.0, oldest-supported-numpy, packaging, py-cpuinfo, setuptools >=61.0.0, wheel)
* Getting build dependencies for wheel...
* Using Python 3.11.2 (main, Feb 15 2023, 08:19:40) [GCC 12.2.1 20230121]
* Found cython 3.0.0b1
* USE_PKGCONFIG: True
* Found HDF5 headers at ``/usr/include``, library at ``/usr/lib64``.
* Found LZO 2 headers at ``/usr/include``, library at ``/usr/lib64``.
.. WARNING:: Could not find the LZO 2 runtime.
   The LZO 2 shared library was *not* found in the default library
   paths. In case of runtime problems, please remember to install it.
* Skipping detection of LZO 1 since LZO 2 has already been found.
* Found bzip2 headers at ``/usr/include``, library at ``/usr/lib64``.
.. WARNING:: Could not find the bzip2 runtime.
   The bzip2 shared library was *not* found in the default library
   paths. In case of runtime problems, please remember to install it.
* Found blosc headers at ``/usr/include``, library at ``/usr/lib64``.
* Found blosc2 headers at ``/tmp/build-env-c08sfzmk/include``, library at ``/tmp/build-env-c08sfzmk/lib64``.
Compiling tables/utilsextension.pyx because it changed.
[1/1] Cythonizing tables/utilsextension.pyx

Error compiling Cython file:
------------------------------------------------------------
...


def _dump_h5_backtrace():
    cdef object bt = []

    if H5Ewalk(H5E_DEFAULT, H5E_WALK_DOWNWARD, e_walk_cb, <void*>bt) < 0:
                                               ^
------------------------------------------------------------

tables/utilsextension.pyx:384:47: Cannot assign type 'herr_t (unsigned int, const H5E_error_t *, void *) except? -1 nogil' to 'H5E_walk_t'
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
    main()
[...]

PR #1008, opened 2023-03-15
Fix build errors when compiled using cython 3.0.0b1.

matusvalo:
Two issues were fixed:
* cython 3 changes default exception handling: https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#exception-values-and-noexcept
* Cython supports namespace packages: https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#namespace-packages

Fixes: #1003

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ requires = [
     "oldest-supported-numpy",
     "packaging",
     "py-cpuinfo",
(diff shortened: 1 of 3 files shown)


avalentino:
Thanks @matusvalo

PR #1009, opened 2023-03-19
Fix wheels workflow

avalentino:
Results of the workflows/wheels.yml can be checked at https://github.com/avalentino/PyTables/actions

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -43,6 +43,7 @@ jobs:
           python -m pip install --upgrade setuptools pip wheel build
           python -m pip install -r requirements.txt
           python -m pip install sphinx>=1.1 sphinx_rtd_theme numpydoc ipython
+          python -m pip install blosc2
 
       - name: Build sdist
         run: make PYTHON=python dist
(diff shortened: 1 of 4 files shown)


.github/workflows/wheels.yml
```diff
@@ -214,8 +221,7 @@ jobs:
 
       - name: Install tables on ${{ matrix.python-version }}
         run: |
-          python -m pip install -r wheelhouse/artifact/requirements.txt
-          python -m pip install --no-index --find-links wheelhouse/artifact/ tables
+          python -m pip install --pre --find-links wheelhouse/artifact/ tables
```
FrancescAlted, line 224: Good point.  Just curious, which is the package that we need a pre-release?

FrancescAlted (approved):

.github/workflows/wheels.yml
avalentino: `tables`, for all the runs except tags.
The problem here was that `numexpr` is no longer listed in `requirements.txt` that now is more for build dependencies (according to the comment in the first line). Probably we should use a more clear naming.

PR #1010, opened 2023-03-21
Update leaf.py

wony-zheng:
avoid RuntimeWarning: overflow encountered in long_scalars

### DIFF
--- a/tables/leaf.py
+++ b/tables/leaf.py
@@ -316,7 +316,7 @@ class Leaf(Node):
 
         # Compute the chunksize
         MB = 1024 * 1024
-        expected_mb = (expectedrows * rowsize) // MB
+        expected_mb = np.int32((np.int64(expectedrows) * rowsize) // MB)
         chunksize = calc_chunksize(expected_mb)
         complib = self.filters.complib
         if (complib is not None and


tables/leaf.py
```diff
@@ -316,7 +316,7 @@ def _calc_chunkshape(self, expectedrows, rowsize, itemsize):
 
         # Compute the chunksize
         MB = 1024 * 1024
-        expected_mb = (expectedrows * rowsize) // MB
+        expected_mb = np.int32((np.int64(expectedrows) * rowsize) // MB)
```
avalentino, line 319: Sorry, I have not analysed the problem in depth but the need of a double casting looks very strange to me.
What is the data type of `expectedrows` normally?
In any case Python `int` should never be subject to overflow. Is it not enough ro have a single cast to `int`?

avalentino: @wony-zheng could you please confirm that the following path works for you?

```patch
-        expected_mb = (expectedrows * rowsize) // MB
+        expected_mb = (int(expectedrows) * rowsize) // MB)
```

avalentino (changes requested):

ivilata:
This should be fixed in a more generic way (by sanitizing `expectedrows` user input) by commit 7d08ab1. Thanks anyway @wony-zheng by the report!

PR #1013, opened 2023-04-03
Handle py-cpuinfo not being installed

thesamesam:
Fallback gracefully if py-cpuinfo isn't installed. We already handle this in setup.py but we need to avoid calling it in leaf.py too.

### DIFF
--- a/tables/leaf.py
+++ b/tables/leaf.py
@@ -4,7 +4,11 @@ import warnings
 import math
 
 import numpy as np
-import cpuinfo
+try:
+    import cpuinfo
+    cpu_info = cpuinfo.get_cpu_info()
+except ImportError:
+    cpu_info = {}
 
 from .flavor import (check_flavor, internal_flavor, toarray,
                      alias_map as flavor_alias_map)
@@ -336,7 +340,6 @@ class Leaf(Node):
             # Use a decent default value for chunksize
             chunksize *= 16
             # Now, go explore the L3 size and try to find a smarter chunksize
-            cpu_info = cpuinfo.get_cpu_info()
             if 'l3_cache_size' in cpu_info:
                 # In general, is a good idea to set the chunksize equal to L3
                 l3_cache_size = cpu_info['l3_cache_size']


FrancescAlted:
Thanks @thesamesam for this.  However, provided that py-cpuinfo is an official requirement, I'd rather go the other way around: just assume that it is installed in setup.py instead of trying to handle its absence.  @avalentino ?

avalentino:
I agree with Francesc.
Cpuinfo is currently tracked as a dependency both a build time and runtime, so , in principe, no need to have fallback paths in the code.
@thesamesam is the installation of cpuinfo problematic on your platform?

thesamesam:
Unfortunately, it is on some of our platforms, like alpha/hppa/ia64/s390 (not x)/older arm. I plan to look at porting cpuinfo to these platforms (it looks like it should be straightforward with enough time) though.

The issue here for us is that things like Pandas get pulled in as test dependencies for some important software (even for non-scientific uses, I mean), and then that itself then needs PyTables for testing.

My humble request would be if you folks would consider allowing this fallback path - no need to support it officially or encourage folks to not have py-cpuinfo, but just allow it provided that we look after the fallback path? I completely understand if that's not okay though. Alternatively, we could continue to just patch it downstream.

Thank you for your time!

tables/leaf.py
```diff
@@ -4,7 +4,11 @@
 import math
 
 import numpy as np
-import cpuinfo
+try:
+    import cpuinfo
+    cpuinfo_available = True
+except ImportError:
+    cpuinfo_available = False
```
avalentino, line 11: If you don't mind, I would suggest to reduce the patch footprint as follows:

```diff
- import cpuinfo
+ try:
+     import cpuinfo
+     cpu_info = cpuinfo.get_cpu_info()
+ except ImportError:
+     cpu_info = {}
```

then the other changes can just be dropped.

avalentino (changes requested):

avalentino:
@FrancescAlted since the patch is really minimal (after the requested change) I would suggest to merge it to be able to support less common platforms where cpuinfo is not available.

Not sure, anyway, how to handle the the build dependencies in `pyproject.toml`.
@thesamesam how do you plan to manage it?

FrancescAlted:
> @FrancescAlted since the patch is really minimal (after the requested change) I would suggest to merge it to be able to support less common platforms where cpuinfo is not available.

:+1:

tables/leaf.py
thesamesam: Much nicer, thank you!

tables/leaf.py
```diff
@@ -4,7 +4,10 @@
 import math
 
 import numpy as np
-import cpuinfo
+try:
+    import cpuinfo
```
avalentino, line 8: Sorry there is something missing here:
```diff
+    cpu_info = cpuinfo.get_cpu_info()
```

avalentino (changes requested):

tables/leaf.py
thesamesam: Doh.

thesamesam:
Could you kick CI?

avalentino:
> We have three choices, I think:
>
>    1. Just unconditionally keep it in with a comment above it (this is fine for us, fwiw)
>    2. Make it optional with a feature name, but I don't really love this given it requires people to do e.g. pip install PyTables[cpuinfo] or pip install PyTables[full] AFAIK
>    3. Don't include it in requirements at all and automagically use it (use if installed)

Maybe we could try something like the following:

```toml
requires = [
    "setuptools >=61.0.0",
    ...
    "py-cpuinfo; platform_system != 'alpha' and ...",
    ...
]
```

Could you please provide the exact platform specifier (or whatever) for the platforms you are interested in?

avalentino:
Thanks @thesamesam

PR #1021, opened 2023-06-15
DOC: fix get_node docstring formatting

StefRe:
Currently the documentation of [`get_node`](https://www.pytables.org/usersguide/libref/file_class.html#tables.File.get_node) is a bit garbled as Sphinx tries to format the last paragraph as parameters:
image

Fixed by formatting this paragraph as a note.


avalentino:
Thanks @StefRe

Issue #1044, opened 2023-08-14
tables.test() Fails ERROR: None (tables.tests.test_queries.ScalarTableUsageTestCase.None)


thatSAMazing:
PyTables version:    3.8.1.dev0
HDF5 version:        1.14.1-2
NumPy version:       1.24.4
Numexpr version:     2.8.5 (not using Intel's VML/MKL)
Zlib version:        1.2.11 (in Python interpreter)
Blosc version:       1.21.4 (2023-05-16)
Blosc compressors:   blosclz (2.5.1), lz4 (1.9.4), lz4hc (1.9.4), zlib (1.2.13), zstd (1.5.5)
Blosc filters:       shuffle, bitshuffle
Blosc2 version:      2.10.1 (2023-08-01)
Blosc2 compressors:  blosclz (2.5.3), lz4 (1.9.4), lz4hc (1.9.4), zlib (1.2.11.zlib-ng), zstd (1.5.5)
Blosc2 filters:      shuffle, bitshuffle
Python version:      3.11.4 (v3.11.4:d2340ef257, Jun  6 2023, 19:15:51) [Clang 13.0.0 (clang-1300.0.29.30)]
Platform:            macOS-14.0-arm64-arm-64bit
Byte-ordering:       little
Detected cores:      10
Default encoding:    utf-8
Default FS encoding: utf-8
Default locale:      (en_US, UTF-8)

ERROR: None (tables.tests.test_queries.ScalarTableUsageTestCase.None)
ValueError: Expression [] has forbidden control characters.
ValueError: Expression c_bool[0] has forbidden control characters.

Hope this is enough - thanks!

PR #1045, opened 2023-08-17
Improve checks for blosc2 availability

avalentino:
See discussion in #1000.


avalentino:
@FrancescAlted this PR should improve a little bit the check on the availability of blosc2 at import time.
The original code is from a commit of yours.
I'm not really sure of the purpose of that check, I assume to raise an early error, but probably on linux it is not so useful. Is it more relevant on windows maybe?

@keszybz please feel free to comment/review if you want

avalentino:
The test failure seems to be unrelated, see #1044.

FrancescAlted (approved):
Yes, that seems a little bit better than existing code, so +1

PR #1047, opened 2023-08-17
Ci/fix build wheel

avalentino:


FrancescAlted:
Provided that Python 3.12rc1 will be ABI compatible with 3.12 final, I'd start building for it soon after NumPy is providing wheels for it (which should happen soon).  With that, and to avoid building too many versions, I'd deprecate building wheels for 3.8.  Opinions?

Issue #832, opened 2020-12-05
Support for Apple's arm64 architecture (osx-arm64)


kartzke:
Hello, I get the following error when trying to install tables on a MacBook Pro with the M1 processor.  Part of the error shows PyTables supports ARM, but it failed to detect the CPU, so I'm guessing it's related to `osx-arm64`

`cpuinfo failed, assuming no CPU features: py-cpuinfo currently only works on X86 and some PPC and ARM CPUs.`

Here's the full error:

```
Collecting tables
  Using cached tables-3.6.1.tar.gz (4.6 MB)
    ERROR: Command errored out with exit status 1:
     command: /Users/ryan/miniforge3/envs/ft/bin/python3.8 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/6g/73g1_dms191f5pj1mc5n8byh0000gn/T/pip-install-6weqp8s9/tables_6efa87f1030c473590e79b8388430fc1/setup.py'"'"'; __file__='"'"'/private/var/folders/6g/73g1_dms191f5pj1mc5n8byh0000gn/T/pip-install-6weqp8s9/tables_6efa87f1030c473590e79b8388430fc1/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/6g/73g1_dms191f5pj1mc5n8byh0000gn/T/pip-pip-egg-info-zbtiyiw2
         cwd: /private/var/folders/6g/73g1_dms191f5pj1mc5n8byh0000gn/T/pip-install-6weqp8s9/tables_6efa87f1030c473590e79b8388430fc1/
    Complete output (12 lines):
    /var/folders/6g/73g1_dms191f5pj1mc5n8byh0000gn/T/H5closeb4hby9yj.c:2:5: error: implicit declaration of function 'H5close' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        H5close();
        ^
    1 error generated.
    cpuinfo failed, assuming no CPU features: py-cpuinfo currently only works on X86 and some PPC and ARM CPUs.
    * Using Python 3.8.6 | packaged by conda-forge | (default, Nov 27 2020, 18:57:00)
    * USE_PKGCONFIG: False
[...]

PR #1049, opened 2023-09-05
Try enabling macOS ARM64 wheels

cbrnr:
As discussed in #832, let's see if this works now.


cbrnr:
Once the workflows are approved, we'll see how it goes...

avalentino:
If I remember correctly you should name the branch `ci/macos-arm64-wheels` to enable wheels creation.

PR #1056, opened 2023-09-19
Support for Blosc2 ND arrays with direct chunking (optimized)

ivilata:
PyTables already leverages HDF5 direct chunking for `Table` (unidimensional) datasets, where data chunks are offloaded to a filter for accelerated Blosc2 compression and decompression during I/O.

This PR brings this feature to multidimensional chunked datasets too (`CArray` and `EArray` objects). The HDF5 Blosc2 filter is extended to accept a target Blosc2 blocksize and chunk dimensions as filter parameters on array creation (from `H5ARRAYOmake` in `src/H5ARRAY-opt.c`), and store them in the HDF5 chunk's data as a Blosc2 b2nd metalayer (with an opaque NumPy dtype), for a superchunk that contains a single chunk, i.e. each HDF5 chunk is stored as a b2nd-tagged superchunk with one chunk; all chunks have the same shape (even fringe ones with smaller data). To select a blocksize for Blosc2 compression, the filter queries Blosc2 for a target blocksize (as in C-Blosc2's `examples/get_blocksize.c`), then computes a blockshape that fits into that target blocksize (with the algorithm in python-blosc2's `blosc2/core.py`). As mentioned above, array creation already computes a target blocksize to spare the filter from doing it for each chunk (such computation has also been ported to `src/H5TB-opt.c`). The environment variable `PT_DEFAULT_B2_BLOCKSIZE` may be used to force a target blocksize (in bytes) regardless of Blosc2 recommendation, but it should only be used for internal testing purposes (it is also supported for tables).

All chunk read and write operations go through the HDF5 filter pipeline, however slice reading operations with matching endianness and step=1 (i.e. contiguous) avoid the pipeline and follow an optimized path using HDF5 direct chunking, implemented in `src/H5ARRAY-opt.c`. The Python extension (Cython code) decides whether the array is fit for optimized slicing, and the slice read operation `H5ARRAYOreadSlice` in `src/H5*-opt.c` checks that the slice is supported by optimized slicing code; otherwise it reverts to the normal filter pipeline (it does also revert when the optimized read fails for some reason, as the data is not modified).

The actual optimized slicing is done by `get_blosc2_slice` in `src/H5ARRAY-opt.c`, which reads the parts of the requested slice in the different chunks (with `read_chunk_slice_b2nd`) and assembles the final read slice.

Error handling in `src/H5ARRAY-opt.c` has been updated to use macros; this also affects other functions in that file not necessarily related with optimized slicing, which haven't changed otherwise (the macros may also be applied to other `src/H5*.c` source but this hasn't been done). Some refactoring and simplification is still possible between `H5ARRAYmake` and `H5ARRAYOmake`, and `H5ARRAYreadSlice` and `H5ARRAYOreadSlice`.

A few unit tests for `CArray` and `EArray` have been added to check for the same operations in both unoptimized (filter) and optimized (direct) paths, especially for cross-chunk slices.

A benchmark is included (based on python-blosc2's `bench/ndarray/compare_getslice.py`) whose results are shown with some graphics in a new passage of the /Optimization tips/ section in the User Manual. The passage provides some recommendations on choosing appropriate chunksizes depending on the desired compression parameters (and associated blocksize).


FrancescAlted:
FWIW, here it is a link for the results of the benchmark where it is shown the degree of optimizations done here: https://github.com/PyTables/PyTables/pull/1056/commits/33966b06a306b5cbdc9ecdcf031dba888aac8432

pyproject.toml
```diff
@@ -3,7 +3,7 @@
 requires = [
     "setuptools >=61.0.0",
     "wheel",
-    "oldest-supported-numpy",
+    "numpy", #"oldest-supported-numpy",
```
avalentino, line 6: Just for curiosity, is there any specific issue with "oldest-supported-numpy"?

avalentino (approved):

avalentino:
> FWIW, here it is a link for the results of the benchmark where it is shown the degree of optimizations done here: [33966b0](https://github.com/PyTables/PyTables/commit/33966b06a306b5cbdc9ecdcf031dba888aac8432)

@FrancescAlted I'm not sure to understand why the non optimised version in PyTables is so slower w.r.t. the h5py version.
In my understanding both relay on the HDF5 filter pipeline, but probably I'm missing something ...

martaiborra (approved):
LGTM!

pyproject.toml
ivilata: Hi @avalentino, commit fba55dd45 message indicates "This should avoid requiring very old versions of Blosc2 for older Python 3.8 and 3.9.", but @FrancescAlted's commit e6086fa06 indicates "Fallback to use oldest-supported-numpy (blosc2 should support this now)". I see that Blosc2 depends on oldest-supported-numpy from v0.6.3.

I recall having discussed this with @FrancescAlted, he can probably offer a better explanation for the change (that I should definitely provide in a comment or in the commit message :sweat_smile: ).

Thanks for having a look!

FrancescAlted:
> > FWIW, here it is a link for the results of the benchmark where it is shown the degree of optimizations done here: [33966b0](https://github.com/PyTables/PyTables/commit/33966b06a306b5cbdc9ecdcf031dba888aac8432)
> 
> @FrancescAlted I'm not sure to understand why the non optimised version in PyTables is so slower w.r.t. the h5py version. In my understanding both relay on the HDF5 filter pipeline, but probably I'm missing something ...

Yeah, that's a good question indeed.  As fas as I remember, h5py has traditionally had a better handling of multidimensional hyperslicing, and I remember myself back porting part of their job to PyTables many years ago.  My guess is that they have advanced more in this area since my backport (they are avoiding memory copies somewhere?), so it would be nice to revisit their code in the future and see exactly why they are having better speed with our non-optimized path.

pyproject.toml
FrancescAlted: Yeah, so sorry, but I don't remember the details, but IIRC during our recent work, we detected that  `oldest-supported-numpy` gave problems, so I decided to give regular "numpy" a try, and that fixed the issue. 

Anyway, by looking at https://pypi.org/project/oldest-supported-numpy/2023.8.3, building with that makes sense to me.  So I'd suggest to enforce `oldest-supported-numpy` again and retry :-)

avalentino:
Thanks @FrancescAlted 
I opened #1058 to not forget this point

pyproject.toml
avalentino: I agree

ivilata: While working on the branch for the v3.9.0 release, commit 626de54d did update Blosc2 dependencies, and this resulted in CI errors like <https://github.com/PyTables/PyTables/actions/runs/6248015264/job/16961771859#step:6:100>, which may clarify the issue:

```
ERROR: Cannot install blosc2==2.2.6, blosc2==2.2.7 and oldest-supported-numpy==0.1 because these package versions have conflicting dependencies.

  The conflict is caused by:
      oldest-supported-numpy 0.1 depends on numpy==1.17.3; python_version >= "3.8" and platform_system != "AIX"
      blosc2 2.2.7 depends on numpy>=1.20.3
      oldest-supported-numpy 0.1 depends on numpy==1.17.3; python_version >= "3.8" and platform_system != "AIX"
      blosc2 2.2.6 depends on numpy>=1.20.3
```

I then cherry-picked fba55dd for that branch, added commit a203326 with a better commit message, and CI seems to be happy with that. So I may revert the change in this PR's branch, but make the change in the release branch. Or I can undo the changes in the latter too if they don't look like the way to go yet. :wink:

avalentino: Thanks @ivilata for taking care of this.
In this case, maybe we should update the GHS for building wheel to use `numpy==1.20.3`.

ivilata: You're welcome @avalentino! After commit a203326, in the wheels CI action the "Build source/Install dependencies" decides to install NumPy 1.26.0 and everything seems to work fine: <https://github.com/PyTables/PyTables/actions/runs/6261949618>. Do you think we really should set a fixed version, or is it ok to just let `pip` et al. decide on their own?

PR #1057, opened 2023-09-20
Fix and update release timeline in release notes

ivilata:
The latest releases where missing; added them with the dates when their respective Git tags where created.

Also fix a typo in the PyTables 0.9 line.


avalentino (approved):

avalentino:
Thanks @ivilata

Issue #1062, opened 2023-10-06
regression: ResolutionImpossible with tables 3.9.0 when installing on Ubuntu 20.04 Python 3.8


astraw:
Hi,

With the release of tables 3.9.0 on Python 3.8 in Ubuntu 20.04, a [scheduled CI build of `flydra_core` is failing with `ResolutionImpossible`](https://github.com/strawlab/flydra/actions/runs/6425542868/job/17448289578). The ResolutionImpossible is that oldest-supported-numpy 0.1 depends on numpy==1.17.3 and blosc2 2.2.9 depends on numpy>=1.20.3. I am also seeing this on in-house CI builds, so it's not a single isolated incident. The fundamental issue may not be in tables but the proximal change introducing breakage does seem to be the new tables release. I have not dug further into the issue.

```
2023-10-05T23:29:06.5690421Z ##[group]Run python -m pip install ./flydra_core
2023-10-05T23:29:06.5690926Z [36;1mpython -m pip install ./flydra_core[0m
2023-10-05T23:29:06.5749984Z shell: /usr/bin/bash -e {0}
2023-10-05T23:29:06.5750303Z env:
2023-10-05T23:29:06.5750693Z   pythonLocation: /opt/hostedtoolcache/Python/3.8.18/x64
2023-10-05T23:29:06.5751176Z   PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.8.18/x64/lib/pkgconfig
2023-10-05T23:29:06.5751619Z   Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.8.18/x64
2023-10-05T23:29:06.5751992Z   Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.8.18/x64
2023-10-05T23:29:06.5752430Z   Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.8.18/x64
2023-10-05T23:29:06.5752884Z   LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.18/x64/lib
2023-10-05T23:29:06.5866861Z ##[endgroup]
2023-10-05T23:29:07.0893369Z Processing ./flydra_core
2023-10-05T23:29:07.0937038Z   Installing build dependencies: started
2023-10-05T23:29:12.9899257Z   Installing build dependencies: finished with status 'done'
2023-10-05T23:29:12.9912107Z   Getting requirements to build wheel: started
[...]

Issue #1061, opened 2023-10-05
Version 3.9.0 depends on blosc2>=2.2.8, which doesn't support Python 3.8


natkam:
It looks like the new version of PyTables released today cannot be installed with Python 3.8, because its [dependency blosc2](https://github.com/PyTables/PyTables/blob/v3.9.0/pyproject.toml#L76) in version 2.2.8 [requires Python>=3.9](https://github.com/Blosc/python-blosc2/blob/v2.2.8/pyproject.toml#L33).

I'm running Python 3.8.10 on Ubuntu 20.04, and installing the newest version of `tables` produces this error:
```
✗ pip install tables
Collecting tables
  Downloading tables-3.9.0.tar.gz (4.7 MB)
     |████████████████████████████████| 4.7 MB 20.8 MB/s 
  Installing build dependencies ... error
  ERROR: Command errored out with exit status 1:
   command: /home/natkam/Projects/foo/venv-pytables/bin/python3 /home/natkam/Projects/foo/venv-pytables/lib/python3.8/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-1c_tbu2l/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- 'setuptools >=61.0.0' wheel oldest-supported-numpy packaging py-cpuinfo 'Cython >=0.29.32' 'blosc2 >=2.2.8'
       cwd: None
  Complete output (14 lines):
  Collecting setuptools>=61.0.0
    Downloading setuptools-68.2.2-py3-none-any.whl (807 kB)
  Collecting wheel
    Downloading wheel-0.41.2-py3-none-any.whl (64 kB)
  Collecting oldest-supported-numpy
    Downloading oldest_supported_numpy-2023.8.3-py3-none-any.whl (4.8 kB)
  Collecting packaging
    Downloading packaging-23.2-py3-none-any.whl (53 kB)
  Collecting py-cpuinfo
    Downloading py_cpuinfo-9.0.0-py3-none-any.whl (22 kB)
  Collecting Cython>=0.29.32
    Downloading Cython-3.0.2-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (3.6 MB)
[...]

PR #1064, opened 2023-10-06
Preliminaries for 3.9.1 release

FrancescAlted:
This is a proposal for a hot release for deprecating the support of Python 3.8.  For details, see: https://github.com/PyTables/PyTables/issues/1062 and https://github.com/PyTables/PyTables/issues/1061.  The hot release will be just a new set of wheels and sources, without regenerating docs or anything else.  The idea is that people get a quick error when they try to use PyTables 3.9.1 when using Python 3.8

### DIFF
--- a/ANNOUNCE.txt.in
+++ b/ANNOUNCE.txt.in
@@ -8,7 +8,8 @@ We are happy to announce PyTables @VERSION@.
 What's new
 ==========
 
-XXX version-specific blurb XXX
+This is a hot fix release to deprecate support for Python 3.8.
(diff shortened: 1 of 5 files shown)


FrancescAlted:
I think I'll go ahead and try to do the hot fix release.  In case I fuck this, let's have another try next week.

avalentino:
Thanks @FrancescAlted

ivilata:
Tanks for fixing this, @FrancescAlted! :vulcan_salute:

PR #1068, opened 2023-10-10
support numpy 2 nightlies

graingert:
numpy v2 is scheduled to be released in January 2024 (about 12 weeks from now) https://hackmd.io/@seberg/Bk5P9wJuj so it would be good to support and test the numpy v2 nightly wheels with PyTables.

This is also blocking dask where we test with PyTables and numpy v2.dev0 in our CI


tables/atom.py
```diff
@@ -306,20 +306,26 @@ def from_sctype(cls, sctype, shape=(), dflt=None):
 
             >>> import numpy as np
             >>> Atom.from_sctype(np.int16, shape=(2, 2))
-            Int16Atom(shape=(2, 2), dflt=0)
+            Int16Atom(shape=(np.int64(2), np.int64(2)), dflt=np.int16(0))
```
graingert, line 309: I'm not sure what to do about these reprs, in the tests I can use `f"{np.int64(2)!r}"` but in the docstrings I have to use either the numpy1 or numpy2 docstrings

I could skip these docstring tests on numpy 2 builds?

graingert: I hacked these in to be 1/2 compatible by using `repr(x) == f" ... "` a bit horrible but it passes on both now

graingert:
@avalentino can I get a review on this please?

tables/atom.py
```diff
@@ -305,21 +305,33 @@ def from_sctype(cls, sctype, shape=(), dflt=None):
         sctype not represented in an Atom is ignored::
 
             >>> import numpy as np
-            >>> Atom.from_sctype(np.int16, shape=(2, 2))
-            Int16Atom(shape=(2, 2), dflt=0)
+            >>> (
+            ...  repr(Atom.from_sctype(np.int16, shape=(2, 2))) ==
+            ...  f"Int16Atom(shape=({np.int64(2)!r}, {np.int64(2)!r}), dflt={np.int16(0)!r})"
+            ... )
+            True
```
avalentino, line 312: As a general comment, I would prefer to keep the examples in the documentation as clean as possible.
If it is not possible to address it easily for both numpy1.x and 2.x then probably it is better to keep a clean docstring working for one of the numpy versions and maybe disable doctests for that function/module.

tables/atom.py
```diff
@@ -305,21 +305,33 @@ def from_sctype(cls, sctype, shape=(), dflt=None):
...
             Traceback (most recent call last):
             ...
             ValueError: unknown NumPy scalar type: 'S5'
-            >>> Atom.from_sctype('float64')
-            Float64Atom(shape=(), dflt=0.0)
+            >>> (
+            ...  repr(Atom.from_sctype('float64')) ==
+            ...  f"Float64Atom(shape=(), dflt={np.float64(0.0)!r})"
+            ... )
+            True
```
avalentino, line 322: Same as above.

tables/atom.py
```diff
@@ -376,10 +394,16 @@ def from_type(cls, type, shape=(), dflt=None):
...
+            >>> (
+            ...  repr(Atom.from_type('bool')) ==
+            ...  f"BoolAtom(shape=(), dflt={np.False_!r})"
+            ... )
+            True
+            >>> (
+            ...  repr(Atom.from_type('int16', shape=(2, 2))) ==
+            ...  f"Int16Atom(shape=({np.int64(2)!r}, {np.int64(2)!r}), dflt={np.int16(0)!r})"
+            ... )
+            True
```
avalentino, line 406: Same as above.

avalentino (changes requested):

tables/atom.py
graingert: Which one should I disable the doctests for?

graingert: I'll disable the doctests on numpy 2

tables/atom.py
graingert: I'm not running the doctests on numpy v2 anymore

graingert:
@avalentino I've made the changes, can I have another review?

tables/atom.py
```diff
@@ -317,9 +317,15 @@ def from_sctype(cls, sctype, shape=(), dflt=None):
...
-            if sctype not in np.sctypeDict:
+            if "," in sctype:
                 raise ValueError(f"unknown NumPy scalar type: {sctype!r}")
-            sctype = np.sctypeDict[sctype]
+            try:
+                dtype = np.dtype(sctype)
+            except TypeError:
+                raise ValueError(f"unknown NumPy scalar type: {sctype!r}") from None
+            if issubclass(dtype.type, np.flexible) and dtype.itemsize > 0:
+                raise ValueError(f"unknown NumPy scalar type: {sctype!r}") from None
```
graingert, line 327:
```suggestion
                raise ValueError(f"unknown NumPy scalar type: {sctype!r}")
```

avalentino (approved):

graingert:
@avalentino sorry about the last minute commit "[should only from None inside the TypeError block](https://github.com/PyTables/PyTables/pull/1068/commits/c6c38721fd67753564a5d21a7225991d95132f63)" just as you approved

avalentino:
Thanks @graingert

PR #1070, opened 2023-10-11
types/doc: Start introducing types

pyssling:
Add a type to _v_pos in descriptions.py, move doc to be under _v_pos. This does not change the sphinx output other than that the type annotation is now present.

### DIFF
--- a/tables/description.py
+++ b/tables/description.py
@@ -800,11 +800,7 @@ class IsDescription(metaclass=MetaIsDescription):
 
     .. rubric:: IsDescription attributes
 
-    .. attribute:: _v_pos
-
-        Sets the position of a possible nested column description among its
-        sibling columns.  This attribute can be specified *when declaring*
-        an IsDescription subclass to complement its *metadata*.
+    .. autoattribute:: _v_pos
 
     .. attribute:: columns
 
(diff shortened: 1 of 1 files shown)


pyssling:
Hi,

This is mostly a proof of concept. I recently started using pytables in a strictly typed setting and realized that I was missing type information in the package, which forced me to create stubs.

I'd be happy to upstream the stubs I have created if they are wanted, and am wondering if this is a nice way to proceed? This would make them "inline". Alternatively they can be kept separate in `.pyi` files, but I think this is less useful when navigating the code.

pyssling:
@FrancescAlted Could I have some feedback regarding this PR? If this is a direction you want taken with the project I can continue modify and extend classes with more type declarations?

pyssling:
@FrancescAlted sorry to ping you again. Would very much appreciate if you could tell me if this is a desirable approach. Alternatively I could start working on a pytables-types package that keeps the types outside of pytables.

FrancescAlted:
Yes, that's is good.  Can you show up that docs are still rendering ok?  Thanks!

graingert:
@FrancescAlted you might want to enable readthedocs PR builder: https://readthedocs.org/dashboard/pytables/advanced/#id_external_builds_enabled then you can see these built directly in the PR view

FrancescAlted:
I don't think I have access to the readthedocs dashboard for PyTables.  @avalentino can you take care of this?  (or add me to the list of managers somehow).

pyssling:
Screenshot 2023-10-21 at 19 58 36

FrancescAlted:
That's good. Please go ahead with other attributes if you want so. Thanks!

PR #1072, opened 2023-10-19
Foreign b2nd array compatibility

ivilata:
This expands the work on b2nd array support with direct chunking (#1056) to better handle such arrays created with other tools, e.g. when the filter values with chunk rank & shape are missing, or by being more relaxed about the format of the b2nd array used for storing the chunk (e.g. by not requiring it to consist of a single inner chunk, or have a specific blocksize). Some missing tests on chunk data shape have been added in the optimized path.

A new example script has been added to test writing and reading partial chunks.

Finally, the workaround for stack smashing on some versions of GCC has been removed (it seems not to be needed anymore).


FrancescAlted (approved):
Looks very good.  Thanks @ivilata !

hdf5-blosc2/src/blosc2_filter.c
```diff
@@ -459,9 +459,18 @@ size_t blosc2_filter_function(unsigned flags, size_t cd_nelmts,
       goto failed;
     }
 
+    /* Although blosc2_decompress_ctx ("else" branch) can decompress b2nd-formatted data,
+     * there may be padding bytes when the chunkshape is not a multiple of the blockshape,
+     * and only b2nd machinery knows how to handle these correctly.
+     */
     if (blosc2_meta_exists(schunk, "b2nd") >= 0
         || blosc2_meta_exists(schunk, "caterva") >= 0) {
```
FrancescAlted, line 467: I think we can remove the "caterva" check here, as it never reached the production status, and its functionality has been included in "b2nd".

PR #1073, opened 2023-11-01
fix library/header detection in system default paths

nega0:
setup.py uses `compiler.has_function()` to test if a dependency exists in the system's "default paths". Unfortunately it immediately discards the result of that test. This PR fixes that and adds a corresponding output message.


setup.py
```diff
@@ -891,7 +893,11 @@ def find_runtime_path(self, locations=None):
 
             continue  # look for the next library
 
-        if libdir in ("", True):
+        if hdrdir is True and libdir is True:
+            print(f"* Found {package.name} in the standard system search dirs.")
+            # set header dir to one of the default include dirs to fudge checks below
+            hdrdir = default_dirs.header[0]
```
avalentino, line 899: Not totally sure that this approach is safe.
E.g. on Debian systems (and derivatives) it is possible to install the library (`.so`) without header files (which are normally provided in a different Debian package).
This kind of situation IMHO would result in an error at build time, and, considering that PyTables takes a while for building, it would be better to raise an error as early as possible.

avalentino:
@nega0 I'm going to close this PR for the time being.
Please feel free to reopen if you think that the comments can be addressed.

PR #1080, opened 2023-11-12
Switch to H5_USE_110_API

avalentino:
Compatibility with HDF5 1.8.x is no longer needed, so we can switch to the newer API

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -777,6 +777,7 @@ if __name__ == "__main__":
     # 1.6.x API by default
     CFLAGS.extend([
         "-DH5_USE_110_API",
+        "-DH5Rdereference_vers=2",
     ])
 
     # H5Oget_info_by_name seems to have performance issues (see gh-402), so we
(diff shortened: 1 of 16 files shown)


FrancescAlted (approved):
Sounds good to me.

setup.py
```diff
@@ -776,30 +776,8 @@ def find_runtime_path(self, locations=None):
     # Force the 1.8.x HDF5 API even if the library as been compiled to use the
```
ivilata, line 776: LGTM, maybe this comment could be updated as well, and I think some ifdefs in `utils.h` may also be removed. Then there's `utilsextension.get_native_type`, but simplifying that may be more involved.

Thanks for the cleanup!

ivilata (approved):

setup.py
avalentino: Thanks @ivilata 
I have performed some cleanup.
I kindly ask you to give another look if you can.

ivilata (approved):
A very thorough cleanup indeed, LGTM!

PR #1085, opened 2023-11-27
Hash pin wheels.yml dependencies

### LINKED ISSUE #1015: Use hash-pin dependencies on workflows
## Description

I would like to suggest another security practice recommended by the [OpenSSF Scorecard][scorecard-repo] which is to hash pin dependencies to prevent dependency-confusion and typosquatting attacks. 

The change would only be applied to GitHub workflows, dockerfiles and shell scripts dependencies. In the PyTables case we will only need to hash pin github workflows.

This means:
- Hash pinning GitHub Workflow actions.
- Using --require-hashes on pip installs made on the workflows.

Along with hash-pinning dependencies, I also recommend adopting dependabot or renovatebot to help keep the dependencies up to date. Both tools can update  hashes and associated semantic version comments and can be configured to only update the github workflows if you wish. 

Let me know if you are open to evaluate those changes and I'll submit the PR(s) ASAP. 

Any questions or concerns just let me know.
Thanks!

## Additional Context

For more informations about dependency confusion / typosquatting attacks:
- [Detection of malicious packages][google-blog]
- [Counter dependency confusion (aka substitution) attacks][azure-paper] 

For more informations about the dependency-update tools:
- [Dependabot][dependabot]
- [Renovatebot][renovatebot]

[scorecard-repo]: https://github.com/ossf/scorecard
[google-blog]: https://security.googleblog.com/2022/04/
[azure-paper]: https://azure.microsoft.com/en-us/resources/3-ways-to-mitigate-risk-using-private-package-feeds/
[dependabot]: https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/
[renovatebot]:https://www.mend.io/renovate/

joycebrum:
Closes #1015 

Hi again, sorry for the delay on sending the PR! 

As mentioned by @diogoteles08, only the wheels.yml workflow is really sensible since it is publishing an artifact, all the other ones are read only workfows used for tests and testing build.

That said, I've only hash pinned the wheels.yml which required different requirements files. Another possibility would be having only one file to the wheels.yml requirements itself, although it would softly increase the installation time on jobs that only needed a part of the requirements. Let me know if you rather a single file and I'll adjust accordingly.

Wheels.yml run with success https://github.com/joycebrum/PyTables/actions/runs/6984015270


avalentino (approved):

avalentino:
Thanks @joycebrum

Issue #1109, opened 2024-02-12
blosc2 filter should set dparams.schunk on decompression


froody:
The blosc2 filter doesn't support compressing data with multiple filters, but IMHO it should support decompressing such data as it is possible to compress data with blosc2 and then write it with `H5Dwrite_chunk`. Loading this data with pytables however results in an error:

```
  File "H5Z.c", line 1391, in H5Z_pipeline
    filter returned failure during read
  File "hdf5-blosc2/src/blosc2_filter.c", line 555, in blosc2_filter
    Cannot decompress chunk into buffer
```

This error can be fixed by setting `dparams.schunk = schunk;` immediately after this line:

https://github.com/PyTables/PyTables/blob/d10c5c32890505c4a6b42c1b69900782683245c7/hdf5-blosc2/src/blosc2_filter.c#L551

To repro, open this file and do `<file>.root.points.read()`:
[blosc2_bytedelta.h5.zip](https://github.com/PyTables/PyTables/files/14235534/blosc2_bytedelta.h5.zip)

PR #1110, opened 2024-02-12
Set schunk when decompressing

froody:
Fixes https://github.com/PyTables/PyTables/issues/1109

### DIFF
--- a/hdf5-blosc2/src/blosc2_filter.c
+++ b/hdf5-blosc2/src/blosc2_filter.c
@@ -549,6 +549,7 @@ size_t blosc2_filter_function(unsigned flags, size_t cd_nelmts,
       }
 
       blosc2_dparams dparams = BLOSC2_DPARAMS_DEFAULTS;
+      dparams.schunk = schunk;
       dctx = blosc2_create_dctx(dparams);
       status = blosc2_decompress_ctx(dctx, chunk, cbytes, outbuf, (int32_t) outbuf_size);
       if (status <= 0) {


avalentino (approved):
LGTM

ivilata:
@avalentino: @FrancescAlted, @martaiborra and me had a look at this issue and Francesc planned to add some comments here that may be useful before merging the PR. Thanks! :slightly_smiling_face:

ivilata (commented):
As I mentioned, Francesc had some comments on this, I just added him as a reviewer.

FrancescAlted (approved):
@froody .  The ultimate reason why you need this workaround is that you are not setting the typesize in the `filters_meta` field when appying the BYTEDELTA filter (see https://github.com/Blosc/c-blosc2/blob/2f8f6edca5a013cc5f9283be251e9cf1c7a00845/include/blosc2/filters-registry.h#L28).  Indeed, this was not your fault, as it was not documented before ;-)

Other than this, it probably does not hurt to apply your patch, and it actually helps in situations where filters would need to access metainfo stored in the `schunk`, so +1 on merging.  Bonus points if you can add a test exercising this.

froody:
Ah, is there any benefit to setting the typesize on the filter meta? Can you save space in the compressed output by omitting the schunk header?

froody:
So is this going to get merged or are you waiting on something?

avalentino:
I was waiting for the test code requested by Francesc
Do you plan to add some unittests?

Issue #1081, opened 2023-11-15
Performance issue at import time on Ubuntu


anttimc:
I noticed a performance issue on Ubuntu with `cpuinfo.get_cpu_info()`. It causes pytables 3.9.1 import to slow down by approximately one second compared to 3.8.0.

The details of the issue are described here: https://github.com/workhorsy/py-cpuinfo/issues/205

avalentino:
This should be fixed via #1091.
Please feel free to re-open if you still have problems.

anttimc:
The issue seems to have been moved from import time to run time. So I think https://github.com/PyTables/PyTables/pull/1091 will not fix the issue if cpu_info is still required (and not optional) dependency.

The real root cause should be fixed on cpu_info side, though.

avalentino:
@anttimc not sure to follow you comment.

The title of the issue refers to "import time" and I think that that issue is indeed solved IMHO.

I agree with you that when you need to compute automatically the optimum chunk shape you still need to import `cpu_info` and hence you experiment the delay cause by `cpu-info` itself.
By the way, the situation is now better then before because we only spend that time if it is really needed and we only do it once (being the result of the operation cached).

It is not totally clear what you are suggesting now.

From my point of view to improve the situation further se have 2 options:

1. do not use `cpu-info` at all. This could lead to worse `chunkshape` values. I'm open to discussion and to review a PR.
2. make `cpu-info` faster. In this case it is not a PyTables issue, we should move the discussion to a dedicated issue on `cpu-info` itself.

So I miss something?

anttimc:
You're correct, the performance issue _at import time_ is solved by the PR. According to the changes in the PR performance issue of `cpuinfo.get_cpu_info()` is now at run time when calling `Leaf._calc_chunkshape` and I think that method is called when opening a file for the first time, right?

1. How essential an optimal `chunkshape` is to the overall performance? I have no idea. In pytables version 3.8.0, `Leaf._calc_chunkshape` does not use `cpuinfo` , and the overall performance is acceptable. https://github.com/PyTables/PyTables/blob/e34d1f78ed905694672c49a33ef3004a2f8899f2/tables/leaf.py#L310
2. I have already opened an issue in `cpuinfo` https://github.com/workhorsy/py-cpuinfo/issues/205 as mentioned in the original post. Let's hope it gets some attention.

avalentino:
[...]

PR #1118, opened 2024-03-08
Store cpuinfo in a cache file for performance

anttimc:
This is yet another fix related to the discussion in issue #1081 

Benefit of this approach:
- the file cache will be available for different processes
- the cpuinfo needs to be processed only once per machine/user

Json format was the first that came to my mind. It could be changed to something else.

### DIFF
--- a/tables/leaf.py
+++ b/tables/leaf.py
@@ -24,7 +24,6 @@ def read_cached_cpu_info():
 
 
 def write_cached_cpu_info(cpu_info_dict):
-    """ Update the cached CPU info in the rc file."""
     with open(Path.home() / '.pytables-cpuinfo.json', 'w') as f:
         return json.dump(cpu_info_dict, f, indent=4)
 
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Thanks @anttimc .  As I read this, I think that the JSON file is read every time is needed.  Would not be better to use a global variable and read from it instead that from disk the second time?  Not a big gain, but it would be almost free with your setup.

Also, may I suggest a more specific name for the file?  Something like `.pytables-cpuinfo.json` should work.

anttimc:
Thanks! I could also put the lru_cache decorator back. Then the get_cpu_info call would be cached within the process and any expensive call within the function would ve avoided

anttimc:
My original idea was to use the "rc" file also for something else. But now after renaming it only store the cpuinfo dictionary.

FrancescAlted (approved):
LGTM

FrancescAlted:
Thanks @anttimc !  If @avalentino is not against this, I'll merge soon.

avalentino (approved):

avalentino:
LGTM
Thanks @anttimc

PR #1119, opened 2024-03-11
Add type hints to path.py

KoStehner:
Hello,

I added type hints to the functions lying in `tables/path.py`.

I did target Python `>=3.9` (the `setup.py` says `>=3.8`, but the `pyproject.toml` defines `>=3.9` -> may be an issue?).
In Python 3.9 we are able to type hint tuples via `(str, str)` instead of `typing.Tuple[str, str]`.

### DIFF
--- a/tables/path.py
+++ b/tables/path.py
@@ -50,7 +50,7 @@ _warnInfo = (
 """Warning printed when a name will not be reachable through natural naming"""
 
 
-def check_attribute_name(name):
+def check_attribute_name(name: str) -> None:
     """Check the validity of the `name` of an attribute in AttributeSet.
 
     If the name is not valid, a ``ValueError`` is raised.  If it is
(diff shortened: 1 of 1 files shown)


KoStehner:
Thank you very much.

In the `setup.py` on lines 533-540 we have
~~~
    # The minimum required versions
    # Keep in sync with ``project.requires-python`` in ``pyproject.toml``.
    min_python_version = (3, 8)
    # Check for Python
    if sys.version_info < min_python_version:
        exit_with_error("You need Python %s or greater to install PyTables!"
                        % '.'.join(str(v) for v in min_python_version))
    print(f"* Using Python {sys.version.splitlines()[0]}")
~~~

There we have `min_python_version = (3, 8)`.
-> I guess it should be updated to 3.9

avalentino (approved):
LGRM

PR #1120, opened 2024-03-11
Add type hints to atom.py

KoStehner:
Hello,

I added further type hints to functions and methods in `tables/atom.py`.

Kind regards
Ko

### DIFF
--- a/tables/atom.py
+++ b/tables/atom.py
@@ -2,7 +2,7 @@
 
 import re
 import inspect
-from typing_extensions import Any, Callable, dataclass_transform, Dict, Union
+from typing_extensions import Any, Callable, dataclass_transform, Dict, NoReturn, Union
 import warnings
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


tables/atom.py
```diff
@@ -101,7 +101,7 @@ def _normalize_shape(shape) -> Shape:
     return tuple(SizeType(s) for s in shape)
 
 
-def _normalize_default(value, dtype):
+def _normalize_default(value: Any, dtype: np.dtype) -> np.ndarray:
```
avalentino, line 104: Here probably it is better to use DTypeLike.
In this case, by the way, we should bump the minimum numpy version to 1.20

avalentino (changes requested):

tables/atom.py
KoStehner: Good idea! 
I upgraded the minimum `numpy` version from `>=1.19` to `>=1.20`, and changed the annotation to `DTypeLike`.

avalentino (approved):

avalentino:
Thanks again @KoStehner

PR #1121, opened 2024-03-12
Add type hints to node.py

KoStehner:
Hello,

I added type hints to `tables/node.py`.

Note: at the beginning I used 
~~~
if TYPE_CHECKING:
    from .group import Group
    from .link import SoftLink
~~~

This is to prevent circular imports as the `tables/group.py` already imports the `tables/node.py`. 

By putting the needed imports behind a [typing.TYPE_CHECKING](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) flag (which always evaluates to `False`), we can use the full type hint support without any need for refactoring.

Best regards
Ko

### DIFF
--- a/tables/node.py
+++ b/tables/node.py
@@ -2,6 +2,7 @@
 
 import warnings
 import functools
+from typing import Any, Callable, Optional, TYPE_CHECKING, Union
 
 from .registry import class_name_dict, class_id_dict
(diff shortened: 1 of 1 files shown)


avalentino (approved):
LGTM

avalentino:
Thanks @KoStehner

PR #1130, opened 2024-03-21
Add type hints to array.py

KoStehner:
Hello,

I added annotations to `table/array.py`.

Kind regards
Ko

### DIFF
--- a/tables/array.py
+++ b/tables/array.py
@@ -2,9 +2,10 @@
 
 import operator
 import sys
-from typing import Any, Optional, Sequence, Union, TYPE_CHECKING
+from typing import Any, Optional, Union, TYPE_CHECKING
 
 import numpy as np
+import numpy.typing as npt
 
 from . import hdf5extension
 from .filters import Filters
@@ -123,7 +124,7 @@ class Array(hdf5extension.Array, Leaf):
     def __init__(self,
                  parentnode: "Group",
                  name: str,
-                 obj: Union[int, float, Sequence, np.ndarray, None]=None,
+                 obj: Optional[npt.ArrayLike]=None,
                  title: str="",
                  byteorder: Optional[str]=None,
                  _log: bool=True,
@@ -166,7 +167,7 @@ class Array(hdf5extension.Array, Leaf):
         """Current row in iterators (sentinel)."""
         self._init = False
         """Whether we are in the middle of an iteration or not (sentinel)."""
-        self.listarr: Union[list, np.ndarray, None] = None
+        self.listarr: Optional[npt.ArrayLike] = None
         """Current buffer in iterators."""
 
         # Documented (*public*) attributes.
--- a/tables/array.py
+++ b/tables/array.py
@@ -26,6 +26,8 @@ if TYPE_CHECKING:
 # obversion = "2.3"    # This adds support for enumerated datatypes.
 obversion = "2.4"    # Numeric and numarray flavors are gone.
 
+SelectionType = Union[int, slice, list[Union[int, slice]], np.ndarray]
+
 
 class Array(hdf5extension.Array, Leaf):
     """This class represents homogeneous datasets in an HDF5 file.
@@ -363,8 +365,8 @@ class Array(hdf5extension.Array, Leaf):
                 return self.listarr    # Scalar case
 
     def _interpret_indexing(
-            self, keys: Union[int, slice, list[Union[int, slice]], np.ndarray],
-        ) -> tuple[np.ndarray, np.ndarray, np.ndarray, list[int]]:
+        self, keys: SelectionType,
+    ) -> tuple[np.ndarray, np.ndarray, np.ndarray, list[int]]:
         """Internal routine used by __getitem__ and __setitem__"""
 
         maxlen = len(self.shape)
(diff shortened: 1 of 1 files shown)


tables/array.py
```diff
@@ -107,17 +113,22 @@ def rowsize(self):
...
-    def __init__(self, parentnode, name,
-                 obj=None, title="",
-                 byteorder=None, _log=True, _atom=None,
-                 track_times=True):
-
-        self._v_version = None
+    def __init__(self,
+                 parentnode: "Group",
+                 name: str,
+                 obj: Union[int, float, Sequence, np.ndarray, None]=None,
```
avalentino, line 124: Probably `Sequence` and `np.ndarray` can be collapsed into  `ArrayLike`

tables/array.py
```diff
@@ -137,33 +148,33 @@ def __init__(self, parentnode, name,
...
-        self._stopb = None
+        self._stopb: Optional[int] = None
         """Stopping row for current buffer. """
-        self._row = None
+        self._row: Optional[int] = None
         """Current row in iterators (sentinel)."""
         self._init = False
         """Whether we are in the middle of an iteration or not (sentinel)."""
-        self.listarr = None
+        self.listarr: Union[list, np.ndarray, None] = None
```
avalentino, line 167: `Optional[ArrayLike]` ?

tables/array.py
```diff
@@ -348,7 +362,9 @@ def __next__(self):
             else:
                 return self.listarr    # Scalar case
 
-    def _interpret_indexing(self, keys):
+    def _interpret_indexing(
+            self, keys: Union[int, slice, list[Union[int, slice]], np.ndarray],
```
avalentino, line 366: Not sure if in this case we could use `ArrayLike`.
Probably your solution is sharper.

tables/array.py
```diff
@@ -590,7 +610,9 @@ def translate_slice(exp, length):
         mshape = tuple(x for x in mshape if x != 0)
         return selection, reorder, mshape
 
-    def __getitem__(self, key):
+    def __getitem__(
+        self, key: Union[int, slice, list[Union[int, slice]], np.ndarray]
```
avalentino, line 614: See comment above.

tables/array.py
```diff
@@ -638,7 +660,9 @@ def __getitem__(self, key):
 
         return internal_to_flavor(arr, self.flavor)
 
-    def __setitem__(self, key, value):
+    def __setitem__(self,
+                    key: Union[int, slice, list[Union[int, slice]], np.ndarray],
```
avalentino, line 664: See comment above

tables/array.py
```diff
@@ -767,21 +798,29 @@ def _read_selection(self, selection, reorder, shape):
...
         """Write `nparr` values in points defined by `coords` coordinates."""
 
         if len(coords) > 0:
             nparr = self._check_shape(nparr, (len(coords),))
             self._g_write_coords(coords, nparr)
 
-    def _write_selection(self, selection, reorder, shape, nparr):
+    def _write_selection(self,
+                         selection: list[tuple[int, int, int, int, str]],
+                         reorder: Optional[tuple[int, np.ndarray]],
```
avalentino, line 822: Probably in this case `ArrayLike` could be used, not sure by the way.
What do you think?

avalentino (changes requested):

tables/array.py
KoStehner: Sounds good! Changed accordingly.

tables/array.py
KoStehner: Unfortunately, a `slice` is not `ArrayLike`. Because of this, I let the annotation as is.
If you want to have the change, I can change it, though.

tables/array.py
KoStehner: I am not exactly sure if this is applicable here as no `Union` type hint is used, thus there is no collapsing.

- For `nparr` the annotation `np.ndarray` is needed as in the following `self._check_shape(nparr, tuple(shape))` call, the `shape` attribute of the numpy array is needed.
- The same applies to `shape`. It needs to be a tuple of integers as we are combining the shape to the `nparr.shape` value in the `self._check_shape` call.
- `reorder` is either `None` or a tuple with two values that do get unpacked. An `ArrayLike` object may just be an array with a single values (or with hundreds of values) and unpacking into two objects will not be possible.
- Similar, each `selection` element consists of four integer and one string (which will be used in `Array.perform_selection` in `tables/hdf5extension.pyx`down the line).

Did you perhaps commented on the wrong line?

avalentino: Sorry, my question was related to the possibility or not to use 
```python
reorder: Optional[tuple[int, ArrayLike]],
```

instead of
```python
reorder: Optional[tuple[int, np.ndarray]],
```

tables/array.py
avalentino: In this case the comment was about the possibility to use 
```python
keys: Union[int, slice, list[Union[int, slice]], ArrayLike],
```

in place of 

```python
keys: Union[int, slice, list[Union[int, slice]], np.ndarray],
```

KoStehner: Ah, got it! Thanks for the input. Changed it to `ArrayLike`.

tables/array.py
KoStehner: Changed to `ArrayLike`

tables/array.py
KoStehner: Changed to `ArrayLike`.

tables/array.py
KoStehner: Makes sense. Yes, changed it to `ArrayLike`. Thanks!

avalentino (approved):
LGTM

avalentino:
Thanks @KoStehner

PR #1136, opened 2024-03-26
Add type hints to index.py

KoStehner:
Hello,

I added annotations to `tables/index.py`.

Further, there is small annotation fix in `tables/indexes.py`.

Best regards
Ko

### DIFF
--- a/tables/index.py
+++ b/tables/index.py
@@ -1147,8 +1147,8 @@ class Index(NotLoggedMixin, Group, indexesextension.Index):
 
     def get_neworder(self,
                      neworder: np.ndarray,
-                     src_disk: Array,
-                     tmp_disk: Array,
+                     src_disk: "Array",
+                     tmp_disk: "Array",
                      lastrow: LastRowArray,
                      nslices: int,
                      offset: int,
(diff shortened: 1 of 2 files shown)


tables/index.py
```diff
@@ -1130,8 +1145,14 @@ def cleanup_temp(self):
         # ...but the memory data cache is dirty now
         self.dirtycache = True
 
-    def get_neworder(self, neworder, src_disk, tmp_disk,
-                     lastrow, nslices, offset, dtype):
+    def get_neworder(self,
+                     neworder: np.ndarray,
+                     src_disk: EArray,
+                     tmp_disk: CArray,
```
avalentino, line 1151: Maybe we could use more generic types for the two parameters above.
in principle anything supporting indexing should work.
I would suggest to use "Array"

avalentino (changes requested):

tables/index.py
KoStehner: Thanks for the feedback. Changed both annotations to `Array`.

avalentino: @KoStehner there is some issue with the CI
I think that you need to use quotes: `Array` --> `"Array"`

KoStehner: You are right, My bad. I fixed it

avalentino (approved):

avalentino:
Thanks @KoStehner

PR #1142, opened 2024-03-31
Add type hints to description.py

KoStehner:
Hello,

I added annotations to `tables/description.py`.

The annotations were a bit tricky sometimes since we handling with class objects instead of instances of classes here.
For this, the `typing.Type` hint was used. See also
- https://docs.python.org/3.9/library/typing.html#typing.Type

Best regards
Ko

### DIFF
--- a/tables/description.py
+++ b/tables/description.py
@@ -2,6 +2,7 @@
 
 import copy
 import warnings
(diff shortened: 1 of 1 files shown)


avalentino (approved):

avalentino:
Thanks @KoStehner

PR #1148, opened 2024-04-02
Add type hints to utils.py

KoStehner:
Hello,

I added annotations to `tables/utils.py`.

Best regards
Ko

### DIFF
--- a/tables/utils.py
+++ b/tables/utils.py
@@ -7,10 +7,15 @@ import warnings
 import weakref
 from pathlib import Path
 from time import perf_counter as clock
+from typing import Any, Callable, Literal, Optional, TextIO, Union, TYPE_CHECKING
 
 import numpy as np
+import numpy.typing as npt
 
 from .flavor import array_of_flavor
+
+if TYPE_CHECKING:
+    from .atom import Atom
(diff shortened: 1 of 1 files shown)


tables/utils.py
```diff
@@ -34,7 +39,7 @@ def correct_byteorder(ptype, byteorder):
         return byteorder
 
 
-def is_idx(index):
+def is_idx(index: Union[int, np.integer, np.ndarray]) -> bool:
```
avalentino, line 42: Not sure about the type hint for the `index` argument.
This function checks whenever the input argument can work as an index or not, hence the arcument could also legitimately be something that cannot be used as an index

avalentino (changes requested):

tables/utils.py
KoStehner: Good point. I guess I was a little bit confused about a missing `else` case (which is just the last `return` statement).
I changed it to `Any` to allow everything.

avalentino (approved):

tables/utils.py
avalentino: Thanks

PR #1151, opened 2024-04-04
Add type hints to expression.py

KoStehner:
Hello,

I added annotations to `tables/expression.py`.

This was another tough one.

Best regards
Ko

### DIFF
--- a/tables/expression.py
+++ b/tables/expression.py
@@ -14,6 +14,7 @@ from .parameters import IO_BUFFER_SIZE, BUFFER_TIMES
 if TYPE_CHECKING:
     from .array import Array
     from .table import Column
+    from .tableextension import Row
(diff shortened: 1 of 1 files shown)


tables/expression.py
```diff
@@ -624,7 +650,7 @@ def eval(self):
 
         return out
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator:
```
avalentino, line 653: Iterator["Row"] maybe?

avalentino (approved):

tables/expression.py
KoStehner: Thanks for the hint! I added `"Row"`.

avalentino (approved):

Issue #1157, opened 2024-04-16
Wheels for macos arm64


maxnoe:
Pytables currently only offers wheels on macos for 64 bit, it would be great if you could also provide wheels for arm64 (apple sillicon)

avalentino:
Dear @maxnoe 
there is currently an incomplete PR #1050 about this topic.
We managed to build wheels but, unfortunately, they seem not to be working due to the bundling of the incorrect version of the blosc2 library.
The origin of the issue is still not clear but any help is really welcome.

cbrnr:
The issue IIRC is that cibuildwheel uses the incorrect platform for installing. Building seems to work. It would be great if you commented/helped in #1050 @maxnoe!

avalentino:
Closing as duplicate of #832 and addressed by PR #1050.

PR #1158, opened 2024-04-16
Enable macos arm wheels

maxnoe:
To address #1157


maxnoe:
How could I trigger the wheel building and testing here?

avalentino:
The failure with numpy2 is due to the fact that pytables still does not support it.

avalentino:
> How could I trigger the wheel building and testing here?

you should use a branch name with the "ci/*" prefix, e.g. "ci/macos_wheels"

PR #1160, opened 2024-04-18
Numpy 2.0 compatibility

maxnoe:


maxnoe:
Python 3.9 builds are failing because the new version of blosc2 that supports numpy 2.0 is requiring python >=3.10

maxnoe:
https://scientific-python.org/specs/spec-0000/ says to drop python 3.9 support in Q4 2023, @avalentino should we simply also drop 3.9 here?

avalentino:
> https://scientific-python.org/specs/spec-0000/ says to drop python 3.9 support in Q4 2023, @avalentino should we simply also drop 3.9 here?

OK for me, thanks

maxnoe:
Ok, we are down to only a handful of failures under 2.0 now

maxnoe:
The change is due to the type promotion changes in numpy 2.0, the tests expect in16, but according to the new rules, the values are promoted to int64.

@avalentino Do we need to keep the old behavior or is it fine to adapt the tests to check for int64 if running under numpy 2.0?

avalentino:
> @avalentino Do we need to keep the old behavior or is it fine to adapt the tests to check for int64 if running under numpy 2.0?

I would need to check on a chase by case basis.
As a preliminary consideration I would say that for all the test involving the expression evaluator it should be fine to follow the numpy2 convention (when run with numpy2).
@FrancescAlted do you agree? What was the approach in numexpr?

maxnoe:
Actually, it's a bit different: the expected type changed, but pytables didn't. So the question would be: do we change the pytest expression engine rules to follow the new type promotions of bumpy 2.0

FrancescAlted:
> > @avalentino Do we need to keep the old behavior or is it fine to adapt the tests to check for int64 if running under numpy 2.0?
> 
> I would need to check on a chase by case basis. As a preliminary consideration I would say that for all the test involving the expression evaluator it should be fine to follow the numpy2 convention (when run with numpy2). @FrancescAlted do you agree? What was the approach in numexpr?

IMO, the NumPy team did a good job in changing promotion rules for 2.0 only when it was common sense, so our approach for numexpr was to follow the new conventions.  I also think that PyTables should adopt the new conventions and note that in the release notes.

avalentino:
Thanks @FrancescAlted

d-chambers:
Numpy 2.0 dropped yesterday. A fresh pip install of pytables is now broken. 

```
import tables
```
produces:
```
  ...
```

maxnoe:
@avalentino I am not familiar enough with the expression code base to make these changes, I would be happy if you take over this PR.

avalentino:
Dear @maxnoe dear @larsoner,
thanks a lot for you contributions to PyTables. I appreciate it a lot and I apologize for not being able to address your PRs more quickly.
I think that this PR and #1176 cover a lot of issues and I would like to merge into the main branch asap.
By the way I would like to do it is steps in order to have more control on what is going.
Probably this will take time but, if you don't mind, I would prefer to proceed like this.

The plan:
* I have already merged some changes that are propaedeutical for supporting numpy 2 in `master` (part of the first commit of this PR)
* then I would like to proceed with dropping Python 3.9 (from this PR)
* refresh dependencies (from this PR) but still keeping `numpy < 2`

Then I would move to the changes in #1176 
In that case I need to carefully review the changes in `setup.py`  to ensure that nothing is broken (not sure that we cover all possible installation scenarios in CI).

Once I have merges all changes that are needed to have a working CI with `numpy < 2` we will have a more clear playground to fix the final issues that still block the support to numpy 2.0

From a practical point of view both this PR and #1176 have conflicts in this moment.
I don't wand to ask you to do additional job e.g. to rebase the PRs or to split them in smaller ones, so, if it is OK for you, I will try to proceed myself by `cherry-pick`ing commits from your PRs as needed.
Of course I will do my best to preserve the attributions.
In the meanwhile I fear that we will have the CI partially broken, but I hope to be able to get back to an acceptable situation asap.

Please let me know if you have any comment.
I think that I will start merging the first commits in `master` starting form tomorrow.

ivilata:
Thanks a lot @avalentino for taking care of the integration of all this! The last CI run on this PR only had test failures in `test_expression` (plus an odd one in `test_indexvalues`), and only with Python 3.12. Really impressive work!

(If you have any ideas on how me or others may help with this, please ask!)

avalentino:
@ivilata I started looking on the residual errors  in `test_expression` and apparently there is an inconsistent behavior, in some cases, from what returned by numexpr/pytables and what returned by the standard `eval`.
Probably it is just a matter of agreeing about what is the expected behavior for us and updating the tests consequently (trying to stay backward compatible).

I think that @FrancescAlted already commented about it but after looking more closely at the errors I still have some doubts.

Depending on what we decide, there could be something to discuss at numexpr level and eventually a submit a small patch for numexpr.

By the way I have not actually checked is something related to this has been already done in #1176 by @larsoner 

From my point of view having a clear agreement on what is the expected behavior for the failing tests would help a lot to speedup the process. If anyone could comment about it I would appreciate.

Of course, as you say, most of the job has been done by @maxnoe and @larsoner . Thanks again.

maxnoe:
Hi, I am happy about any progress in this matter and I really don't mind who commits what in which PR.

Happy to help though!

avalentino:
Thanks a lot @maxnoe 
Closing then.
Manually merged

larsoner:
I'm fine with cherry-picking etc., just give me a minute to squash my commits in #1176

avalentino:
Sure, thanks @larsoner

PR #1176, opened 2024-07-05
Update and fix wheel building

### LINKED ISSUE #1172: Broken with NumPy 2.0.0
PyTables appears to be broken with the recently released NumPy version 2.0.0.

From a fresh python 3.10 virtual environment, after `pip install tables`

`python3 -m tables.tests.test_all` gives:
```python
  File "/usr/lib/python3.10/runpy.py", line 187, in _run_module_as_main
  File "/usr/lib/python3.10/runpy.py", line 110, in _get_module_details
    __import__(pkg_name)
  File "/home/.venv2/lib/python3.10/site-packages/tables/__init__.py", line 44, in <module>
``` 
`pip freeze` gives:
```python
blosc2==2.6.2
msgpack==1.0.8
ndindex==1.8
numexpr==2.10.0
numpy==2.0.0
packaging==24.1
py-cpuinfo==9.0.0
tables==3.9.2
``` 
Downgrading numpy to < 2.0.0 (e.g. 1.26.4) fixes this

larsoner:
Needs #1175, hence in draft mode for now. Working locally on my own branch but needed a place to start keeping track of all the relevant changes.

1. Include changes from #1174 (update artifact upload/download)
2. Include changes from #1160 (limit to Python 3.10+, Numpy 2 fixes)
3. Refactor nightly and normal wheel building -- they were 99% identical, it's just that the nightly one has an upload step at the end, so just make that conditional.
4. Refactor Unix and Windows wheel building to use the same set of steps (with GHA `if` statements for the few that differ)
5. Fix macOS `arm64` wheel building by building our own HDF5 (like h5py does) rather than using the precompiled binaries 
6. Move as much `cibuildwheel` config to `pyproject.toml` as possible to facilitate local running of `python -m cibuildwheel`

Closes #1179
Closes #1178
Closes #1174 
Closes #1173
Closes #1172
Closes #1160


tables/tests/test_indexvalues.py
```diff
@@ -2316,13 +2316,18 @@ def test12a(self):
         self.assertEqual(results1, results2)
 
         # Second selection: bool
+        # TODO: Fails for FINSV2aTestCase
```
larsoner, line 2319: @maxnoe @avalentino this PR is very close to working. This is the one test I couldn't figure out how to fix. There is something wrong with the `results1 = [p["var2"] for p in table1.where('t1var2 == True')]` -- there are four entries that are `True` but it only finds two of them :( Any help here would be greatly appreciated.

It can be replicated locally "just" by building against NumPy 2.0 (e.g., on this branch and with NumPy 2.0 on your local system) and running the test.

maxnoe: Is that maybe an issue of python bool vs. numpy's np.bool_ ?

setup.py
```diff
@@ -456,12 +456,25 @@ def find_directories(self, location, use_pkgconfig=False, hook=None):
...
         directories = [None, None, None]  # headers, libraries, runtime
         for idx, (name, find_path, default_dirs) in enumerate(dirdata):
-            path = find_path(
+            use_locations = (
                 locations
                 or pkgconfig_dirs[idx]
                 or hook_dirs[idx]
                 or default_dirs
             )
+            # pkgconfig does not list bin/ as the runtime dir
```
larsoner, line 465: Bug with upstream packaging (I think)

setup.py
```diff
@@ -623,7 +636,7 @@ def find_runtime_path(self, locations=None):
 
     # Gcc 4.0.1 on Mac OS X 10.4 does not seem to include the default
     # header and library paths.  See ticket #18.
-    if sys.platform.lower().startswith("darwin"):
+    if sys.platform.lower().startswith("darwin") or CONDA_PREFIX:
```
larsoner, line 639: Without this, the packages won't build on Windows (detection fails)

setup.py
```diff
@@ -641,10 +654,11 @@ def find_runtime_path(self, locations=None):
     elif os.name == "nt":
         _Package = WindowsPackage
         _platdep = {  # package tag -> platform-dependent components
+            # Library name, runtime name
             "HDF5": ["hdf5", "hdf5"],
             "LZO2": ["lzo2", "lzo2"],
             "LZO": ["liblzo", "lzo1"],
-            "BZ2": ["bzip2", "bzip2"],
+            "BZ2": ["libbz2", "bzip2"],
```
larsoner, line 661: Pretty sure this was wrong. Wasn't detected before but is now

setup.py
```diff
@@ -826,29 +843,29 @@ def find_runtime_path(self, locations=None):
                 package.target_function,
                 includes=(package.header_name + ".h",),
                 libraries=(package.library_name,),
+                include_dirs=[str(d) for d in inc_dirs],
+                library_dirs=[str(d) for d in lib_dirs],
```
larsoner, line 847: Without these the library test binary wouldn't compile / be detected on windows

ci/github/get_hdf5.sh
```diff
@@ -29,8 +30,6 @@ echo "building HDF5"
 if [[ "$OSTYPE" == "darwin"* ]]; then
     brew install automake cmake pkg-config
 
-    CMAKE_ARCHES="$CIBW_ARCHS"
-    ARCH_ARGS="-arch $CIBW_ARCHS"
```
larsoner, line 33: Not needed because native compilation is used for Intel and ARM64 / M1 now

tables/tests/test_indexvalues.py
larsoner: I thought it could be... tried various tricks here but couldn't get it to work. Feel free to try some stuff locally.

Don't know if this is informative:
```
>>> np.bool(True) == True == np.bool_(True)
np.True_
```
maybe it's because this evaluates to `np.True_` instead of `True`?

.github/workflows/wheels.yml
```diff
@@ -52,185 +61,160 @@ jobs:
...
-          # Keep Python in sync with ``project.classifiers`` in ``pyproject.toml``.
-          CIBW_BUILD: "cp39-${{ matrix.arch }} cp310-${{ matrix.arch }} cp311-${{ matrix.arch }} cp312-${{ matrix.arch }}"
-          # Keep in sync with "Prerequisites" in User's Guide.
-          CIBW_BEFORE_ALL_WINDOWS: >
-            conda create --yes --name=build &&
-            conda activate build &&
-            conda config --env --set subdir ${{ matrix.arch == 'win32' && 'win-32' || 'win-64' }} &&
-            conda install --yes blosc bzip2 cython hdf5 lz4 lzo snappy zstd zlib
-          CIBW_ENVIRONMENT_WINDOWS: 'CONDA_PREFIX="C:\\Miniconda\\envs\\build" PATH="$PATH;C:\\Miniconda\\envs\\build\\Library\\bin"'
-          CIBW_ENVIRONMENT: "PYTABLES_NO_EMBEDDED_LIBS=true DISABLE_AVX2=true"
```
larsoner, line 199: Pretty sure the `CIBW_ENVIRONMENT` vars here weren't being used because the `CIBW_ENVIRONMENT_WINDOWS` was also defined.

If I set `PYTABLES_NO_EMBEDDED_LIBS=true` it fails, presumably because `blosc2` isn't in the build deps? In any case, now `PYTABLES_NO_EMBEDDED_LIBS` isn't used anywhere, and locally `cibuildwheel` on Windows works so hopefully it's okay.

larsoner:
Okay I think I narrowed town the last remaining failure but need a bit of help. This self-contained example fails:

<details>
<summary>MWE</summary>

```
import tables as tb


class Small(tb.IsDescription):
    var2 = tb.BoolCol(dflt=0)


small_blocksizes = (16, 8, 4, 2)  # The smaller set of parameters...
chunkshape = 2
ss = small_blocksizes[2]
nrows = ss * 2-1
with tb.open_file('bad.h5', 'w', node_cache_slots=64) as h5file:
    table1 = h5file.create_table(
        h5file.root, 'table1', Small, "", None, nrows, chunkshape=(chunkshape,),
    )
    for i in range(nrows):
        table1.row['var2'] = i % 2
        table1.row.append()
    table1.flush()
    for col in table1.colinstances.values():
        indexrows = col.create_index(kind="full", _blocksizes=small_blocksizes)
    for i in range(nrows//2):
        table1.row['var2'] = i % 2
        table1.row.append()
    assert not table1.cols.var2.index.dirty

    # TODO: IT IS DIRTY BECAUSE THIS FIXES THINGS:
    # for col in table1.colinstances.values():
    #     col.reindex()

    table1.flush()
    t1var2 = table1.cols.var2
    assert not table1.cols.var2.index.dirty
    results1 = [p["var2"] for p in table1.where('t1var2 == True')]
    results2 = [p["var2"] for p in table1 if p["var2"] is True]
    t1var2_vals = [p["var2"] for p in table1]
    msg = (
        f"Incorrect results for t1var2[n] == True where\n"
        f"t2var2_vals={repr(t1var2_vals)}\n"
        f"\n{results1=}\n{results2=}"
    )
    assert len(results1) == len(results2), msg

```

</details>

It fails the same way as the tests:
```
  File "/home/larsoner/Desktop/rep_tables.py", line 42, in <module>
    assert len(results1) == len(results2), msg
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Incorrect results for t1var2[n] == True where
t2var2_vals=[False, True, False, True, False, True, False, False, True, False]

results1=[True, True]
results2=[True, True, True, True]
```

But if you uncomment these lines:

```
    for col in table1.colinstances.values():
        col.reindex()
```
the example passes. So I'm pushing that to the test for now just to make sure it generalizes **but it's not an actual fix because this shouldn't be necessary.** This seems to be an issue with dirty detection / updating (?) and maybe related to the `copy_if_needed` stuff but I'm not sure. I'm out of my depth on this stuff so if any `PyTables` experts / maintainers can sort this out feel free to push commits!

In the meantime, the wheel builds appear to succeed (at least before my latest commit, but hopefully that will come back green, too :crossed_fingers: )!

larsoner:
Closing for #1183

PR #1183, opened 2024-07-12
ENH: Bump Numpy

### LINKED ISSUE #1172: Broken with NumPy 2.0.0

larsoner:
All (hopefully) of the CI bits and some of the Python >= 3.10 and NumPy 2 compat parts of #1176. @avalentino hopefully this is useful as something you can `cherry-pick`!

Closes #1172


larsoner:
@avalentino this is headed to be green (I think), feel free to have a look and see if it's useful, or if there are changes I can make, etc.!

avalentino:
Thanks @larsoner.
As anticipated in #1160, the idea is to bring in fixes on the CI before moving to numpy2 topics.
I need to cross-check with the other PyTables developers the solutions used for `test_expressions` and `test_indexvalues`.
Will try to work on it later today or during the weekend

larsoner:
@avalentino I can take some time to try and remove the NumPy 2 stuff. I'll open a separate PR for that and if it works close this one.

avalentino:
Thanks a lot @larsoner. It would be fantastic.
By the way, if you do not want to spend time on it I can do it myself.

pyproject.toml
```diff
@@ -3,17 +3,17 @@
 # "Prerequisites" in User's Guide,
 # build jobs in `.github/workflows/*.yml`.
 requires = [
-    "setuptools >= 61.0.0",
+    "setuptools >=61.0.0",
     "wheel",
-    "numpy < 2",
+    "numpy >=2,<3",
```
avalentino, line 8: This is a point that probably we should discuss.
I would like to keep compatibility with numpty 1.x for the time being (it could be v1.26 only if needed).
I think that right now this compatibility is there no we just should not enforce `numpy>=2` in the `pytpoject.toml` file.
Moreover please consider that `ubuntu.yml` (correctly IMHO) still uses `lumpy <2`.

avalentino (changes requested):

pyproject.toml
larsoner: A point of clarification -- this is in the *build-system* requires, not the *project dependencies*. A project built with NumPy 2.0 will have code that is both NumPy 2.0+ and Numpy 1.x compatible. So I think we're okay here?

(As some secondary proof of this beyond the [official docs](https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-specific-advice) stating this compatibility, the Ubuntu test below that installs `numpy<2` does pass without any modification.)

avalentino: The point is that I would like to be able to build PyTables from sources also on platforms where (for any reason) numpy 2 is not available. (At least for one release.)
I'm specifically thinking to Linux packaging (for e.g. Ubuntu/Debian).

In our CI I would expect that numpy 2 is used in any case event if we do not require it explicitly in the `[build-system]` section.

larsoner: ... actually we're already guaranteed that the wheel build will use 2.0 simply because the `build-requirements.in`/`.txt` here pins to NumPy 2.0, so this line can just go back to what it was before #1184

pyproject.toml
```diff
@@ -3,17 +3,17 @@
...
 # build jobs in `.github/workflows/*.yml`.
 requires = [
-    "setuptools >= 61.0.0",
+    "setuptools >=61.0.0",
     "wheel",
-    "numpy < 2",
+    "numpy >=1.20.0",
     "packaging",
     "py-cpuinfo",
     "Cython >=3.0.10",
```
larsoner, line 11: ... and arguably we shouldn't have bumped the `cython` build requirement either if we're trying to help packagers, so I'll revert that, too.

avalentino: Cython is not a big issue, for Ubuntu/Debian at least.
... but going back doesn't hurts for the time being

larsoner:
Looking at https://anaconda.org/conda-forge/bzip2/files they updated 1 day ago, probably not coincidentally the CI worked 2 days ago and is failing today. Looking at https://github.com/conda-forge/bzip2-feedstock/commit/e044094adb40d9ac5a8dbbaa6d0854c608a2d02a I can see that the library should now (only) be found at `libbz2.dll` instead of `bz2.dll` so I think I can adjust. I'll try a fix based on this and if it doesn't work get the VM running again :roll_eyes:

larsoner:
@avalentino all green here, feel free to use this (or not) as you see fit. It at least shows where further work is needed to get proper NumPy 2.0 compatibility.

tables/tests/test_expression.py
```diff
@@ -265,9 +265,12 @@ def test00b_simple_scalars(self):
         if common.verbose:
             print("Computed expression:", repr(r1), r1.dtype)
             print("Should look like:", repr(r2), r2.dtype)
-        self.assertTrue(
-            r1.shape == r2.shape and r1.dtype == r2.dtype and r1 == r2,
-            "Evaluate is returning a wrong value.")
+        msg = f"Evaluate is returning a wrong value: {expr_str}\n{r1=}\n{r2=}"
+        self.assertEqual(r1.shape, r2.shape, msg=msg)
+        # In something like 2 * np.in16(3) + np.int16(2) the result is still a
+        # np.int16 in NumPy 2.0, so we shouldn't actually check this:
```
avalentino, line 271: It is not really clear to me why "we shouldn't actually check this".
Apparently the logic of the test is that the result of the `tb.Expr` evaluation should be totally equivalent to the result of pure Python `eval`.
In principle breaking this logic could could break something on the user side.

On the other hand the current `tb.Expr` behaviour is aligned to the `numexpr` one at the moment.

Honestly I'm not sure how to proceed here.
@ivilata , @FrancescAlted , any suggestion?

tables/tests/test_expression.py
```diff
@@ -305,8 +308,9 @@ def test01b_out_scalars(self):
             if common.verbose:
                 print("Computed expression:", repr(r1), r1.dtype)
                 print("Should look like:", repr(r2), r2.dtype)
-            self.assertTrue(common.areArraysEqual(r1, r2),
-                            "Evaluate is returning a wrong value.")
+            msg = f"Evaluate is returning a wrong value: {expr_str}\n{r1=}\n{r2=}"
+            # On NumPy 2 type promotion is different so don't check type here
+            self.assertTrue(common.areArraysEqual(r1, r2, check_type=False), msg=msg)
```
avalentino, line 313: Same comment of `test00b_simple_scalars` also applies here.

tables/tests/test_indexvalues.py
```diff
@@ -2296,6 +2296,11 @@ def test12a(self):
         self.assertFalse(t1var3.index.dirty)
         self.assertFalse(t1var4.index.dirty)
 
+        # TODO: IT IS DIRTY BECAUSE THIS FIXES THINGS FOR FINSV2aTestCase,
+        # which otherwise fails a test a few lines below!
+        for col in table1.colinstances.values():
+            col.reindex()
```
avalentino, line 2302: This is marked as `TODO`.
I understand that it is a workaround to make the test pass but a fix would be needed somewhere in the PyTables code.
The workaround is for sure interesting because it provides good indications about the possible issue but, in principle, the problem is still there.
I would be not too keen to merge this without at least having a specific github issue for it.
Probably a better approach would be to mark this test as an expected failure for the time being, or something like that.

@ivilata , @FrancescAlted what is you position about this specific point?

tables/tests/test_expression.py
larsoner: Yeah NumPy changed it's up casting behavior (I think,?) so it's possible something needs to be updated.

tables/tests/test_indexvalues.py
larsoner: Yeah I have no idea what the right thing to do is, and this is probably not it! This was about as much as my limited knowledge would allow me to isolate the issue, there is probably a correct fix still remaining to be found / added so that this kludge can be removed.

tables/tests/test_expression.py
ivilata: (Mind you, `f` in the test is `int(3)`, not `np.int16(3)`.)

So PyTables stays with the NumExpr type `int64`, but NumPy changed from `int64` to `int16`. Then relaxing the test seems ok to me, but we should at least check that the dtype kind stays consistent, i.e. `r1.dtype.kind == r2.dtype.kind`.

tables/tests/test_indexvalues.py
ivilata: I agree with @avalentino about not merging without at least having an open issue and adding its number to the comment. However something definitely broke, though I don't know how serious it is. Probably @FrancescAlted has better insights since he knows the indexing code better.

ivilata (changes requested):
The changes look generally good to me, with the exceptions of the overrelaxed type checks for `tb.Expr` and the indexing test issues that I mention in comments.

Thanks a lot @larsoner for the nice work!

tables/tests/test_expression.py
avalentino: Thanks @ivilata, your suggestion sounds a good compromise to me.
@larsoner if it is OK for you we could proceed in that way.

larsoner:
Okay, added a comment about the remaining TODO and added `dtype.kind` checks!

avalentino (approved):

avalentino:
Thanks @larsoner

tables/tests/test_indexvalues.py
FrancescAlted: I am sorry, but I lately have personal duties that prevents me to put quality time on this. I'd appreciate if anybody else can go ahead and try to fix this one.  I think the NumPy team is helping other projects on-boarding with NumPy 2.0, so it is probably worth contacting them about this one.

PR #1187, opened 2024-07-30
Direct chunking API

ivilata:
This adds three new methods to the `Leaf` class that allow working with raw chunks, skipping the HDF5 filter pipeline. The methods are `chunk_info`, `read_chunk` and `write_chunk` (a helper class `ChunkInfo` is also included for the first method).

The API is fully documented in the User's Guide reference chapter, and covered with an introduction and some discussion under optimization tips. Benchmarks have also been carried and discussed in the latter. Example code and unit tests for the implementation are also included.

This work is sponsored by the NumFOCUS Foundation under a Small Development Grant.


tables/leaf.py
```diff
@@ -92,6 +105,46 @@ def calc_chunksize(expected_mb: int) -> int:
...
+    .. attribute:: size
+
+        An integer which indicates the size in bytes of chunk data as it
+        exists in storage.  ``None`` for missing chunks.
+
+    """
+    start: Union[tuple[int, ...], None]
+    filter_mask: Union[int, None]
+    offset: Union[int, None]
+    size: Union[int, None]
```
avalentino, line 145: I think that we dropped support to Python 3.9.
You can use the syntax `int | None` instead of `Union[int, None]` if you want.

tables/leaf.py
```diff
@@ -782,6 +853,117 @@ def flush(self) -> None:
...
+        filter_mask, offset, size = self._g_chunk_info(coords)
+
+        # Align coordinates to chunk boundary.
+        chunkshape = self.chunkshape
+        coords //= chunkshape
+        coords *= chunkshape
+        return ChunkInfo(tuple(coords), filter_mask, offset, size)
+
+    def read_chunk(self, coords: tuple[int, ...],
+                   out: Optional[Union[bytearray, NPByteArray]]=None,
```
avalentino, line 892: In Python v3.10 you can use `Union[bytearray, NPByteArray] | None` if you prefer

avalentino (approved):
LGTM

avalentino:
Thanks @ivilata

avalentino:
The test failure on winwods-latest AMD64 seems to be unrelated.

tables/leaf.py
ivilata: Thanks! I see no other uses of the newer syntax in code, maybe we can leave `Union` for consistency, and upgrade annotations later on.

tables/leaf.py
ivilata: Ditto in the comment above.

FrancescAlted (commented):
LGTM :-)

PR #1188, opened 2024-07-31
MAINT: Fix delvewheel Windows error

larsoner:
Poking around I noticed `master` is failing due to a `delvewheel` problem where it can't find `libblosc2`. I think installing it from conda might be enough to fix it, but we'll see :crossed_fingers:

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -120,7 +120,7 @@ jobs:
         with:
           environment-name: build
           create-args: >
-            python=3.12 blosc c-blosc2 bzip2 hdf5 lz4 lzo snappy zstd zlib pkgconfig
+            python=3.12 blosc bzip2 hdf5 lz4 lzo snappy zstd zlib pkgconfig
           init-shell: bash powershell
         if: runner.os == 'Windows'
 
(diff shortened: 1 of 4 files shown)


larsoner:
@avalentino you committed https://github.com/PyTables/PyTables/commit/03ad4c5a781e7cdda074d63ba8eb3e33b6e26416 which reverted the:
```
            "BZ2": ["libbz2", "libbz2"],
```
that I added previously. This breaks BZ2 support on the wheels. Thoughts?

avalentino:
The change in `_library_prefixes` should make the trick.
Alter the change the issue with bzip2 eas solved and popped up the one with blosc2.
By the way feel free to revert if you want

larsoner:
I pushed a commit to choose the `_platdep["BZ2"]` based on being on Windows *and* having a CONDA_PREFIX, which should be safe enough hopefully. I also improved the wheel test to make sure that all libraries/capabilities are indeed present (this wasn't checked before). Hopefully all comes back green 🤞 

In principle in the future maybe it would be good to have some switch at the build stage in `setup.py` if not all capabilities are being added. That would make it fail earlier. But in the meantime hopefully the post-build check is good enough (and we might as well keep it around even if a build-time check is added later).

larsoner:
Green! 🎉

avalentino (approved):

avalentino:
Thanks @larsoner

PR #1191, opened 2024-08-05
Fix B2ND compression with  Fletcher32

### LINKED ISSUE #1162: carray data written with blosc2 and fletcher32 filters will fail fletcher32 checksum on read
tables is perfectly able to read blosc2:lz4 data from previous versions of tables.  Previous versions were:
tables: 3.8
blosc2: 2.0.0

With newest tables and blosc2 releases, when I write, and then try to read a carray written with blosc2:lz4 and fletcher32 filters, I get a fletcher32 checksum error.  

tables 3.9.2
blosc2 2.6.2
numpy 1.26.4

reproduction code 
```
import tables
import numpy as np
import blosc2
print(tables.__version__)
print(np.__version__)
print(blosc2.__version__)

input_arr = np.array([[0, 1, 1], [1, 0, 1]], dtype='int8')
print(input_arr)

f = tables.open_file('/tmp/tables.h5', mode='w')
print(repr(f))

filters = tables.Filters(complevel=9, complib='blosc2:lz4', shuffle=False, fletcher32=True)
print(repr(filters))

arr = f.create_carray('/', 'data', tables.Int8Atom(), input_arr.shape, filters=filters, chunkshape=None, obj=input_arr)
print(repr(arr))

f.close()

f = tables.open_file('/tmp/tables.h5', mode='r')
print(repr(f))

n = f.get_node('/data')
print(repr(n))

n.read()
```

results in an exception at n.read():

```
HDF5ExtError                              Traceback (most recent call last)
Cell In[31], line 28
     25 n = f.get_node('[/data]
     26 print(repr(n))
---> 28 n.read()

File [/usr/local/rbpython/lib/python3.10/site-packages/tables/array.py:866] in Array.read(self, start, stop, step, out)
    864     raise TypeError(msg)
    865 (start, stop, step) = self._process_range_read(start, stop, step)
--> 866 arr = self._read(start, stop, step, out)
    867 return internal_to_flavor(arr, self.flavor)

File [/usr/local/rbpython/lib/python3.10/site-packages/tables/array.py:823] in Array._read(self, start, stop, step, out)
    820 # Protection against reading empty arrays
    821 if 0 not in shape:
    822     # Arrays that have non-zero dimensionality
--> 823     self._read_array(start, stop, step, arr)
    824 # data is always read in the system byteorder
    825 # if the out array's byteorder is different, do a byteswap
    826 if (out is not None and
    827         byteorders[arr.dtype.byteorder] != sys.byteorder):

File [/usr/local/rbpython/lib/python3.10/site-packages/tables/hdf5extension.pyx:1593] in tables.hdf5extension.Array._read_array()

HDF5ExtError: HDF5 error back trace

  File "H5D.c", line 1061, in H5Dread
    can't synchronously read data
  File "H5D.c", line 1008, in H5D__read_api_common
    can't read data
  File "H5VLcallback.c", line 2092, in H5VL_dataset_read_direct
    dataset read failed
  File "H5VLcallback.c", line 2048, in H5VL__dataset_read
    dataset read failed
  File "H5VLnative_dataset.c", line 363, in H5VL__native_dataset_read
    can't read data
  File "H5Dio.c", line 383, in H5D__read
    can't read data
  File "H5Dchunk.c", line 2856, in H5D__chunk_read
    unable to read raw data chunk
  File "H5Dchunk.c", line 4468, in H5D__chunk_lock
    data pipeline read failed
  File "H5Z.c", line 1391, in H5Z_pipeline
    filter returned failure during read
  File "H5Zfletcher32.c", line 103, in H5Z__filter_fletcher32
    data error detected by Fletcher32 checksum

End of HDF5 error back trace

Problems reading the array data.
```

Works fine if I use zlib instead of blosc2.  With fletcher32 checksum off, I don't see any differences in the read array from the input array that would indicate the data had been corrupted by blosc2...

Let me know if I can provide any other data.

ivilata:
This changes the default filter order to put the Fletcher32 filter at the end of the pipeline, after compression filters. This doesn't make much difference for byte-oriented compressors, but it does for B2ND compression since it understands the structure of data in the chunk. If the Fletcher32 checksum was added before compression to the end of the buffer passed to the Blosc2 filter, it just ignored that data, which caused checksum verification issues on decompression.

By performing and adding the checksum after compression, the compressor doesn't need to worry about the checksum being there (nor does decompression). Of course, this means that enabling Fletcher32 adds 4 bytes to each chunk, but compressing them as part of the data didn't make much sense either.

Some checks have been added to the Blosc2 filter to detect the presence of extra pipeline filters or chunk bytes and disable B2ND optimized read/write in that case.

Fixes #1162.

### DIFF
--- a/hdf5-blosc2/src/blosc2_filter.c
+++ b/hdf5-blosc2/src/blosc2_filter.c
@@ -463,6 +463,15 @@ size_t blosc2_filter_function(unsigned flags, size_t cd_nelmts,
     /* declare dummy variables */
     int32_t cbytes;
 
+    /* It would be cool to have a check to detect whether the buffer contains
+     * an unexpected amout of bytes because of the application of a previous
+     * filter (e.g. a Fletcher32 checksum), thus disabling B2ND optimizations.
+     * However, we cannot know before parsing the super-chunk, and that
+     * operation accepts trailing bytes without reporting their presence.
+     * Thus such a test can only happen outside of the filter, with more
+     * information available about the whole pipeline setup.
+     */
+
     blosc2_schunk* schunk = blosc2_schunk_from_buffer(*buf, (int64_t)nbytes, false);
     if (schunk == NULL) {
       PUSH_ERR("blosc2_filter", H5E_CALLBACK, "Cannot get super-chunk from buffer");
--- a/hdf5-blosc2/src/blosc2_filter.c
+++ b/hdf5-blosc2/src/blosc2_filter.c
@@ -298,6 +298,7 @@ size_t blosc2_filter_function(unsigned flags, size_t cd_nelmts,
   /* Filter params that are only set for B2ND */
   int ndim = -1;
   int32_t chunkshape[BLOSC2_MAX_DIM];
+  size_t chunksize = typesize;
   if (cd_nelmts >= 8) {
     /* Get chunk shape for B2ND */
     ndim = cd_values[7];
@@ -321,6 +322,7 @@ size_t blosc2_filter_function(unsigned flags, size_t cd_nelmts,
     }
     for (int i = 0; i < ndim; i++) {
       chunkshape[i] = cd_values[8 + i];
+      chunksize *= (size_t) cd_values[8 + i];
     }
   }
 
@@ -366,6 +368,14 @@ size_t blosc2_filter_function(unsigned flags, size_t cd_nelmts,
     // cparams.blocksize depends on dimensionality
 
     blosc2_storage storage = {.cparams=&cparams, .contiguous=false};
+
+    if (ndim > 1 && nbytes != chunksize) {
+      BLOSC_TRACE_INFO("Filter input size %lu does not match chunk data size %lu "
+                       "(e.g. Fletcher32 checksum added before compression step), "
+                       "using plain Blosc2 instead of B2ND",
+                       nbytes, chunksize);
+      ndim = -1;
+    }
 
     if (ndim > 1) {
 
--- a/doc/source/usersguide/optimization.rst
+++ b/doc/source/usersguide/optimization.rst
@@ -236,10 +236,11 @@ performance of the HDF5 filter.
 
 Optimized b2nd slicing in PyTables has its limitations, though: it only works
 with contiguous slices (that is, with step 1 on every dimension) and on
-datasets with the same byte ordering as the host machine.  However, the
-performance gains may be worth the extra effort of choosing the right
-compression parameters and chunksizes when your use case is not affected by
-those limitations.
+datasets with the same byte ordering as the host machine and where Blosc2 is
+the only filter in the HDF5 pipeline (e.g. Fletcher32 checksumming is off).
+However, the performance gains may be worth the extra effort of choosing the
+right compression parameters and chunksizes when your use case is not affected
+by those limitations.
 
 
 .. _directChunking:
--- a/hdf5-blosc/src/example.c
+++ b/hdf5-blosc/src/example.c
@@ -72,12 +72,6 @@ int main(){
     r = H5Pset_chunk(plist, 3, chunkshape);
     if(r<0) goto failed;
 
-    /* Using the blosc filter in combination with other ones also works */
-    /*
-    r = H5Pset_fletcher32(plist);
-    if(r<0) goto failed;
-    */
-
     /* This is the easiest way to call Blosc with default values: 5
      for BloscLZ and shuffle active. */
     /* r = H5Pset_filter(plist, FILTER_BLOSC, H5Z_FLAG_OPTIONAL, 0, NULL); */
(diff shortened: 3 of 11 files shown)


ivilata:
I'm not sure yet about what to do with setting `blosc2_support_{read,write}` in `hdf5extension.pyx` and `tableextension.pyx`. Probably they should become false if any other filter besides `blosc2` is detected in the pipeline.

@avalentino, @FrancescAlted any ideas?

avalentino:
> I'm not sure yet about what to do with setting `blosc2_support_{read,write}` in `hdf5extension.pyx` and `tableextension.pyx`. Probably they should become false if any other filter besides `blosc2` is detected in the pipeline.

Do you mean that blosc2 is incompatible with any other filter?
What about e.g. scaleoffset or nbit?

ivilata:
No no, it's just that *optimized B2ND reading/writing* (i.e. feeding chunk data directly into Blosc2 decompression, skipping the HDF5 filter pipeline) would be disabled if other filters besides Blosc2 are detected, so the operations would revert to the normal pipeline, that's it. :wink:

avalentino:
Ah OK
IMHO it is fair.

hdf5-blosc2/src/blosc2_filter.c
```diff
@@ -367,6 +369,14 @@ size_t blosc2_filter_function(unsigned flags, size_t cd_nelmts,
 
     blosc2_storage storage = {.cparams=&cparams, .contiguous=false};
 
+    if (ndim > 1 && nbytes != chunksize) {
+      BLOSC_TRACE_INFO("Filter input size %lu does not match chunk data size %lu "
+                       "(e.g. Fletcher32 checksum added before compression step), "
+                       "using plain Blosc2 instead of B2ND",
+                       nbytes, chunksize);
+      ndim = -1;
```
FrancescAlted, line 377: If I read correctly, this is still allowing to create compressed datasets with Fletcher32 being used prior to Blosc2, right?  I don't think choosing where Fletcher32 is used is supported by PyTables, but my guess is that this trace message is more for pure HDF5/h5py apps that can choose the order of the pipeline. Is that the intent?

FrancescAlted (approved):
I just have a question, but other than that, this looks good to me.

hdf5-blosc2/src/blosc2_filter.c
ivilata: Yes, indeed it is. :slightly_smiling_face: I'd love to add a similar protection for the reading/decompression case, but unfortunately one can't say beforehand in the read buffer contains data that won't be used for decompression, and `blosc2_schunk_from_buffer` is ok with trailing bytes (which is fine IMO) and doesn't report whether it consumed the whole buffer or not. So the situation cannot be detected by the filter alone.

Thus in practice, following the optimized path for reading a B2ND schunk followed by a Fletcher32 checksum would work, but of course the checksum would not be verified at all, which doesn't feel like correct behaviour to me. That more aggressive optimization behaviour may be implemented (showing some warning or something at the PyTables level) but I'm not sure it's worth the effort.

ivilata: Added note on that in commit 9fee83a4.

ivilata:
Ok, commit ddf05f47 disables optimized B2ND in the presence of non-Blosc2 filters. Leaving draft state of PR. :slightly_smiling_face:

ivilata:
@t20100, you may be interested in the fixes for the HDF5 Blosc2 filter here. :slightly_smiling_face:

avalentino (approved):
LGTM
Thanks @ivilata

hdf5-blosc2/src/blosc2_filter.c
FrancescAlted: > Yes, indeed it is. 🙂 I'd love to add a similar protection for the reading/decompression case, but unfortunately one can't say beforehand in the read buffer contains data that won't be used for decompression, and `blosc2_schunk_from_buffer` is ok with trailing bytes (which is fine IMO) and doesn't report whether it consumed the whole buffer or not. So the situation cannot be detected by the filter alone.
> 
> Thus in practice, following the optimized path for reading a B2ND schunk followed by a Fletcher32 checksum would work, but of course the checksum would not be verified at all, which doesn't feel like correct behaviour to me. That more aggressive optimization behaviour may be implemented (showing some warning or something at the PyTables level) but I'm not sure it's worth the effort.

Ok, at this point I don't think checking the Fletcher32 checksum for the optimized path makes sense either; making it not crashing should be enough for now. In the long run, if such a functionality is desired, it would be better to implement a Fletcher32 (or equivalent) into the Blosc2 library, and call it from there.  Perhaps it would be nice to add a note about this in the docs.

ivilata: Yup, added the note in commit 6bbed526.

t20100:
Thanks for pinging! I'm updating blosc2 filter in `hdf5plugin`:  https://github.com/silx-kit/hdf5plugin/pull/320

PR #1194, opened 2024-08-07
Update included HDF5-Blosc to v1.0.1 (build&doc, minor&format fixes)

ivilata:
The only remaining difference is how the Fletcher32 filter is applied last in the example source.


ivilata:
BTW, I created Blosc/hdf5-blosc#36 to also apply the example code change in the original source.

avalentino (approved):

FrancescAlted (approved):
That's great!  Thanks for the effort in doing this.

ivilata:
Since Blosc/hdf5-blosc#36 was merged, both code bases are identical again. It may be worth considering using a submodule for `c-blosc` instead of a source copy.

avalentino:
yes, I agree

PR #1195, opened 2024-08-07
Feature/update requirememt files

avalentino:


ivilata (approved):
Looks great, thanks a lot @avalentino!

PR #1201, opened 2024-08-13
MAINT: Lower verbosity in wheel build

larsoner:
I noticed that the scheduled `nightly` wheel build didn't finish:

https://github.com/PyTables/PyTables/actions/runs/10340087060/job/28620253697

But it's very difficult to parse the log because of the *very* long `pip` output. I want to reduce that and fix the build if necessary.

### DIFF
--- a/tables/tests/test_all.py
+++ b/tables/tests/test_all.py
@@ -1,6 +1,7 @@
 """Run all test cases."""
 
 import faulthandler
+import os
 import sys
 
 import numpy as np
@@ -11,7 +12,9 @@ from tables.tests import common
 from tables.tests.test_suite import suite, test
 
 
-faulthandler.enable()
+# Give people a way to opt out of enabling faulthandler
+if os.getenv("PYTABLES_DISABLE_FAULTHANDLER", "").lower() not in ("1", "true"):
+    faulthandler.enable()
 
 
 def get_tuple_version(hexversion):
--- a/tables/tests/test_all.py
+++ b/tables/tests/test_all.py
@@ -1,5 +1,6 @@
 """Run all test cases."""
 
+import faulthandler
 import sys
 
 import numpy as np
@@ -8,6 +9,9 @@ from packaging.version import Version
 import tables as tb
 from tables.tests import common
 from tables.tests.test_suite import suite, test
+
+
+faulthandler.enable()
 
 
 def get_tuple_version(hexversion):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,7 +139,7 @@ tables = [ "*.dll", "*.so*", "*.dylib" ]
 [tool.cibuildwheel]
 skip = "*-musllinux_* cp313*"
 build = "cp*"
-build-verbosity = 2
+build-verbosity = 1
 test-command = [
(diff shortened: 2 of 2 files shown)


larsoner:
oof from parsing the very long error log we get:
```
024-08-11T13:35:20.6495873Z /tmp/tmp.s96FO8K1KZ/venv/lib/python3.12/site-packages/tables/filters.py:331: FiltersWarning: compression library ``lzo`` is not available; using ``zlib`` instead
2024-08-11T13:35:20.6497689Z   warnings.warn("compression library ``%s`` is not available; "
2024-08-11T13:35:23.6226323Z .....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................s......sh: line 1: 99095 Segmentation fault      (core dumped) python -m tables.tests.test_all
2024-08-11T13:35:24.6634489Z ##[error]Command ['sh', '-c', 'python -c "import tables; keys = \'zlib bzip2 blosc blosc2\'.split(); missing = [key for key in keys if tables.which_lib_version(key) is None]; assert missing == [], missing" && python -m tables.tests.test_all'] failed with code 139. 
2024-08-11T13:35:24.6646514Z 
```
I'll enable `faulthandler` at least for `test_all`. Might be nice someday to switch to `pytest` to get stuff like that for free.

larsoner:
Okay logs are more readable now, and `faulthandler` is enabled in `test_all` for the next time the wheel build segfaults during testing. Ready for review/merge from my end!

tables/tests/test_all.py
```diff
@@ -10,6 +11,9 @@
 from tables.tests.test_suite import suite, test
 
 
+faulthandler.enable()
```
avalentino, line 14: In principle this is OK for me, but I would prefer to use an environment variable if possible

avalentino (changes requested):

tables/tests/test_all.py
larsoner: Ahh I had assumed that it would always be useful to have this enabled -- I've actually never encountered a case where I didn't want it enabled during testing. But I'll add support for something like `PYTABLES_ENABLE_FAULTHANDLER=0` that could be used, that way it's opt-out.

tables/tests/test_all.py
```diff
@@ -11,7 +12,9 @@
 from tables.tests.test_suite import suite, test
 
 
-faulthandler.enable()
+# Give people a way to opt out of enabling faulthandler
+if os.getenv("PYTABLES_DISABLE_FAULTHANDLER", "").lower() not in ("1", "true"):
+    faulthandler.enable()
```
avalentino, line 17: U was more thinking to [PYTHONFAULTHANDLER](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONFAULTHANDLER) actually, but the current implementation is also fine for me.

avalentino (approved):

tables/tests/test_all.py
larsoner: Ahh I didn't even know about that env var. Yes it could be done this way but then all CIs that call tests would have to be updated to have this env var, as would devs when testing locally, etc. So probably less useful than an opt-out mechanism I think?

avalentino: I agree
Thanks

PR #1206, opened 2024-08-15
PyTables v3.10.1 release

avalentino:
Hotfix release fixing the following bugs:

- Fix version constraints for the ``numpy`` runtime requirements (#1204).
  For a mistake it didn't allow to use PyTables with ``numpy`` 2.x.
- Fix compatibility with PyPy (#1205), Thanks to Michał Górny.

Moreover, the following improvements are also implemented:

- Enforce `numpy >= 2` as build constraint (see discussion in #1200).
- Use tuple of plain ints for chunk info coordinates.
  Thanks to Ivan Vilata-i-Balaguer.
- Enable `faulthandler` in `tables.tests.test_all`. Thanks to Eric Larson.

Remaining blocking issues:
- CI test failure with numpy >= 2.1.dev0

PR #1207, opened 2024-08-15
Avoid array resize segfault in direct chunking test with NumPy 2.1rc

ivilata:
See #1206.  There seems to be a bug with NumPy 2.1rc1 where resizing a compound array causes a segfault, so avoid the resizing by using padding or slicing.


avalentino:
Thanks @ivilata 
The workaround looks reasonable
I have reported the issue in the numpy tracker (numpy/numpy#27225)
Before merging I would wait a couple of days for some feedback form numpy guys

avalentino:
The upstream bug https://github.com/numpy/numpy/issues/27225 has been fixed and PyTables v3.10.1 is now out.
I propose to close this PR

ivilata:
Great, thanks @avalentino!

PR #1213, opened 2024-10-31
Type hints: Accept IsDescription, dict, numpy.dtype as table descriptions

jtc42:
Updates the type hints on `File.create_table`'s `description` argument to allow `dict`, `numpy.dtype`, and `Type[IsDescription]` in addition to `Description`. This matches both the actual functionality, and the type hints of `Table.__init__`.

### DIFF
--- a/tables/file.py
+++ b/tables/file.py
@@ -19,6 +19,7 @@ from typing import Any, Callable, Generator, Iterator, Literal, Optional, Type, 
 
 import numexpr as ne
 import numpy as np
+import numpy.typing as npt
 
 from . import hdf5extension
 from . import utilsextension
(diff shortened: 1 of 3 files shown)


tables/file.py
```diff
@@ -914,7 +914,7 @@ def create_group(self,
     def create_table(self,
                      where: Union[Group, str],
                      name: str,
-                     description: Optional[Description]=None,
+                     description: Union[dict, Type[IsDescription], Description, np.dtype, None]=None,
                      title: str="",
```
avalentino, line 918: Shouldn't we use [DTypeLike](https://numpy.org/devdocs/reference/typing.html#numpy.typing.DTypeLike) instead of `np.dtype`?

avalentino (changes requested):

tables/file.py
avalentino: Thanks @jtc42 
The criterion should be to have `DTypeLike` for parameters and `np.dtype` for outputs (unless there is a very good reason to do otherwise).
I would tend for fix `Tables.__init__` and other possible inconsistencies in this case.

avalentino (changes requested):

avalentino (approved):

avalentino:
Thanks @jtc42

PR #1214, opened 2024-10-31
Type hints: Allow multi-dimension chunkshape when creating arrays

jtc42:
Updates the `chunkshape` parameter type hints for `create_carray`, `create_earray`, and `create_vlarray` to allow multi-dimensional values. E.g. `chunkshape=(1, 256, 256)`.

Library already supports this functionality, but this isn't reflected in the type hints.


avalentino:
Thanks @jtc42

Issue #1221, opened 2024-12-11
Error writing cpu_info


alpae:
Hi,

I was running into an error when I tried to write a pytables file in a singularity container that does not have the user's home folder mounted. The problem is actually very minor: 
when trying to cache the cpu info, the `~/.pytables-cpuinfo.json` can not be written:

https://github.com/PyTables/PyTables/blob/8f5ad6042a9024e803053bedccaf5c9a105d5133/tables/leaf.py#L63

I suggest to ignore any IOErrors writing that file, as the functionailty would still remain the same. It would simply need to be computed for every process once, but within a process, because of the LRU cache,  the function would still be fast.

If you agree, I can submit a PR. 

Cheers Adrian

avalentino:
Hi @alpae 
thanks for reporting.
A PR would be very appreciated indeed.
Thanks a lot.

avalentino:
Closed via #1222

PR #1222, opened 2024-12-11
fixes error writing cpu info to cache file

alpae:
fixes issue #1221. 

If ~/.pytables-cpuinfo.json cannot be written (e.g. readonly), the cpu info will now simply 
be cached during the process lifespan (lru-cache) and will need to be recomputed in the next
process.

### DIFF
--- a/tables/leaf.py
+++ b/tables/leaf.py
@@ -59,11 +59,16 @@ def get_cpu_info() -> dict[str, Any]:
 
     try:
         import cpuinfo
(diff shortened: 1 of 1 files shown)


alpae:
I think the broken doc check is unrelated to my changes. seems to be a missing library in the documentation.

avalentino (approved):

avalentino:
Thanks a lot @alpae

Issue #1219, opened 2024-11-30
Cannot find blosc runtime on macOS 15.0 aarch64


outpaddling:
Trying to build under pkgsrc on my Mac Mini produces the following:

```
* Found bzip2 headers at ``/Users/bacon/Pkgsrc/pkg/include``, library at ``/Users/bacon/Pkgsrc/pkg/lib``.
* Found blosc headers at ``/Users/bacon/Pkgsrc/pkg/include``, library at ``/Users/bacon/Pkgsrc/pkg/lib``.
.. WARNING:: Could not find the blosc runtime.
   The blosc shared library was *not* found in the default library
   paths. In case of runtime problems, please remember to install it.
* Run 'blosc2_find_directories_hook'
* Unable to find blosc2 wheel.
* Found blosc2 headers at ``/Users/bacon/Pkgsrc/pkg/include``, library at ``/Users/bacon/Pkgsrc/pkg/lib``.
  * Copying blosc2 runtime library to 'tables' dir because it was not found in standard locations
```

The exact same pkgsrc package works fine on other platforms, e.g. Alma8:

```
* Found bzip2 headers at ``/home/bacon/Pkgsrc/pkg/include``, library at ``/home/bacon/Pkgsrc/pkg/lib``.
* Found blosc headers at ``/home/bacon/Pkgsrc/pkg/include``, library at ``/home/bacon/Pkgsrc/pkg/lib``.
* Run 'blosc2_find_directories_hook'
* Unable to find blosc2 wheel.
* Found blosc2 headers at ``/home/bacon/Pkgsrc/pkg/include``, library at ``/home
```

There are no hacks of any kind in this pkgsrc package: It's using a pristine tables 3.10.1 dist.  I've poked around setup.py a bit, but I'm not seeing anything obvious.  Any hints that help locate the problem would be appreciated.

avalentino:
Dear @outpaddling,
thanks for reporting.
A couple of questions:
1. could you please let us know what command do you use to install PyTables?
Are you using `pip` or `setup.py` directly?
2. apparently `setup.py` finds a `blosc2 ` installation in `/Users/bacon/Pkgsrc/pkg` and `Copying blosc2 runtime library to 'tables' dir`. Could you please confirm that the the library is effectively copied i the `tables` directory?

outpaddling:
> Dear @outpaddling, thanks for reporting. A couple of questions:
> 
>     1. could you please let us know what command do you use to install PyTables?
[...]

PR #1223, opened 2024-12-12
Patch to find runtime dir on macOS

outpaddling:
This fixes failures  of ctypes.CDLL() to find locate dylibs on macOS, per #1219.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -518,7 +518,7 @@ class BasePackage:
                 else:
                     directories[idx] = Path(path).parent
             else:
-                print("Error: path is not set.")
+                print("Warning: path is not set.")
         return tuple(directories)
 
 
--- a/setup.py
+++ b/setup.py
@@ -359,16 +359,37 @@ class BasePackage:
         # (The argument is accepted for compatibility with previous
         # methods.)
 
+        # Try just library name, relying on the OS search path.
+        # This fails on macOS.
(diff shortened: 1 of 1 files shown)


avalentino:
@outpaddling thanks fo resubmitting the PR.
Apparently there is something that breaks many CI tasks, even if, apparently, the changes look quite innocuous.
Do you have an idea of what is happening?

outpaddling:
Does the CI system scan for the word "Error" in the build output?  That's the only thing I can think of.  I added

```
else
    print("Error: path is not set.")
```
It builds fine for me under pkgsrc on Alma, macOS, and NetBSD.

avalentino:
Hi @outpaddling 
tests were failing due to an unrelated issue that should now be fixed in the master branch.
Could I kindly ask you to rebase you changes into master and (force) push?

The PT looks mostly fine, even if I would prefer to not `print` too many debug messages.

avalentino:
@outpaddling unfortunately there seem to be some issues connected to the change that you are proposing.
Any idea about how to proceed?

avalentino (approved):

avalentino:
Thanks @outpaddling

outpaddling:
I have no idea why the checks passed this time.  ;-)  All I did was remove a warning print() and eliminate a variable, using the expression directly instead.

Issue #1115, opened 2024-02-23
Uploading nightly wheels (for testing and better cross-project integration)


jarrodmillman:
Would you be interested/willing to upload nightly wheels to:
https://anaconda.org/scientific-python-nightly-wheels/

It would also be great if you would testing against nightly wheels that you depend on (i.e., numpy).

You can read more about why and how we are building / testing against nightly wheels here:
https://scientific-python.org/specs/spec-0004/

We provide a GitHub action to upload nightly builds to the [scientific-python nightly channel](https://anaconda.org/scientific-python-nightly-wheels):
https://github.com/scientific-python/upload-nightly-action/

We @matthewfeickert, @larsoner, @tupui are happy to help.

avalentino:
Dear @jarrodmillman, deal al,
I think that it would be nice nightly wheels to in https://anaconda.org/scientific-python-nightly-wheels/, even if, at the moment there is no big evolution planned for PyTables, only basic maintenance, 

Currently we have problems with numpy-2.0-dev.
I have a preliminary patch (see also #1083), but, unfortunately, the job is not progressing due to lack of time.

How would you like to proceed?

matthewfeickert:
@avalentino We have the approval (https://github.com/scientific-python/upload-nightly-action/issues/30#issuecomment-1960037619) to add `pytables` to the nightly wheels Anaconda Cloud org https://anaconda.org/scientific-python-nightly-wheels/!

To get started on this please do the following:

## Steps to gain upload access to https://anaconda.org/scientific-python-nightly-wheels/

* [ ] Create an account on [Anaconda Cloud](https://anaconda.org/)
* [ ] Reply here with the link to your account
* [ ] Make a new Anaconda Cloud API token with permissions described in https://scientific-python.org/specs/spec-0004/#process-for-adding-new-projects

![image](https://github.com/scientific-python/upload-nightly-action/assets/5142394/697ce89f-4a51-43d0-a26f-2493f4bdce34)

* [ ] Save the API token as a repo GitHub Actions secret named `ANACONDA_ORG_UPLOAD_TOKEN`
* [ ] Direct me (@matthewfeickert) to a `tables` nightly wheel for me to upload to create the package in the registry
* [ ] Verify that I have added you to the [`pytables` team "Group"](https://anaconda.org/scientific-python-nightly-wheels/groups/pytables) on https://anaconda.org/scientific-python-nightly-wheels
[...]

PR #1234, opened 2025-02-26
MAINT: Fix wheel build and use native aarch64 runner

larsoner:
1. Fix broken URL and associated dir structure
2. Change from emulation to native building using new publicly available Ubuntu arm64 runner
3. Update to latest HDF5 release
4. Add `workflow_dispatch` trigger

Takes care of https://github.com/PyTables/PyTables/issues/1115#issuecomment-2678888676

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -73,7 +73,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       # Keep in sync with "Prerequisites" in User's Guide.
-      HDF5_VERSION: 1.14.3  # H5Dchunk_iter needs at least 1.14.1
+      HDF5_VERSION: 1.14.6  # H5Dchunk_iter needs at least 1.14.1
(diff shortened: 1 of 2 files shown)


larsoner:
Green! :rocket:

larsoner:
... and pushed a commit such that once this is merged, the workflow can be triggered using `workflow_dispatch` via a "run workflow" (or similar) button that will be in the [wheels actions page](https://github.com/PyTables/PyTables/actions/workflows/wheels.yml) by a maintainer, which will cause the wheels to be uploaded at the end so we don't have to wait until Sunday to get them up to `scientific-python-nightly-wheels`!

avalentino (approved):

avalentino:
Thanks @larsoner

larsoner:
@avalentino I think this page should have a "Run workflow" button

https://github.com/PyTables/PyTables/actions/workflows/wheels.yml

Can you click it? Then the nightly wheels should build and upload

PR #1256, opened 2025-09-25
Fix for numexpr 2.13.0 (should be backward compatible)

FrancescAlted:
This fixes an issue with the recent numexpr 2.13.0.  It happens that, when expressions like:

```
ne.evaluate("s | True", {'s': b'pepe'})
```

are evaluated, the new numexpr returns a `TypeError` instead of a `NotImplementedError` as before.

This new behavior is actually more consistent with Python and NumPy, which also throw `TypeError`:

```
In [10]: b"as" | True
TypeError                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 b"as" | True

TypeError: unsupported operand type(s) for |: 'bytes' and 'bool'

In [11]: s = np.array([b"dsd", b"dsd2"])

In [12]: b"as" | True
TypeError                                 Traceback (most recent call last)
Cell In[12], line 1
----> 1 b"as" | True

TypeError: unsupported operand type(s) for |: 'bytes' and 'bool'
```

IMO, fixing this issue in PyTables is the best way to follow the standard behaviour.  What do you think @avalentino ?

### DIFF
--- a/tables/tests/test_queries.py
+++ b/tables/tests/test_queries.py
@@ -485,9 +485,7 @@ def create_test_method(type_, op, extracond, func=None):
                     for _ in range(2)
                 ]
             except TypeError as te:
-                if self.condNotBoolean_re.search(str(te)):
-                    raise SilentlySkipTest("The condition is not boolean.")
-                raise
+                raise SilentlySkipTest("The condition is not boolean.")
             except NotImplementedError:
                 raise SilentlySkipTest(
                     "The PyTables type does not support the operation."


FrancescAlted:
> > > IMO, fixing this issue in PyTables is the best way to follow the standard behaviour.  What do you think @avalentino ?
> > 
> > 
> > I agree I can implement the fix pretty soon, but I'm not sure that I have the time to make a new release right now
> > I can pin numexpr in the conda recipe but people using pip could experiment some issue
> 
> As I see this, only the people that run the tests locally with PyTables binaries and newer numexpr are going to see the failures (15 in total). IMO, people that runs tests locally is because they typically compile manually because they want to provide a fix or a feature, and this normally happens in master, and in this case, the current PR will fix the issue. Do you foresee any other case where this could be problematic?

To be clear: the change in numexpr 2.13.0 only affects the current test suite of PyTables, not code that is already running.  It happens that the test suite was a bit too much picky about the error *message*, and I don't think this would create any major disruption in existing codebase.

Having said that, if you think that there can be any issue that can create havoc somewhere, we would not put obstacles to go back in numexpr and return a `NotImplementedError` again (although I'd prefer not to because of the consistency with Python and NumPy).

avalentino:
I agree with you
the only people potentially impacted should be the ones doing a "new installation via pip" and "using the expression evaluation of that particular type" and "telling on the exception type in some way" 

IMHO the probability that someone has problems is quite low.

In any case I will push a fix soon an if anyone open an issue on this we can eventually try to have a quick release of PyTables

Is it fine for you?

avalentino:
Thanks @FrancescAlted

PR #1263, opened 2025-10-29
Don't be too picky in getting the release version

FrancescAlted:
This is because we have had a glitch in our release process for Python-Blosc2, and we have included a `.dev` version for C-Blosc2.  PyTables requires an integer for the `BLOSC2_VERSION_RELEASE` symbol, but I think this is too picky.  The current PR relax this so that something like "1.dev" can work as a release number too.


avalentino (approved):
LGTM

avalentino:
Thanks a lot @FrancescAlted

PR #1270, opened 2025-12-02
Set Cython "freethreading_compatible" directive.

nascheme:
If the Cython version is new enough, set the "freethreading_compatible" directive.

I need to do more testing yet but building with Python 3.14t and Cython 3.2.2, running "make check" passes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1061,8 +1061,16 @@ if __name__ == "__main__":
                 # developer should have it installed, so it should not be
                 # a hard requisite
                 from Cython.Build import cythonize
+                from Cython import __version__ as cython_version
 
-                cythonize(str(extpfile), language_level="2")
+                compiler_directives = {}
+                if Version(cython_version) >= Version("3.1.0b1"):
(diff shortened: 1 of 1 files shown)


avalentino (changes requested):
The change looks mostly fine, but I would like to have a mechanism to explicitly enable the free-threading feature. The default should be no free -threading compatibility.
Do you think that it is possible?

avalentino:
My point is that a wheel that is free threading compatible has a different ABI tag (ending with "t", e.g. `numpy-2.3.5-cp314-cp314t-macosx_11_0_arm64.whl`).
My idea is that we should be able to produce both free-threading enabled and standard wheels.
At least for the next release the standard wheel should be generated by default and free-threading enabled one should be generated only on explicit request of the user.
In the end it should not make too much difference because people using wheels will go through PyPI (where both options will be available).
It is just a way to stress the experimental character of this change.
At next release + 1, for sure we will re-evaluate the situation.
Does it make sense for you?

avalentino:
Thanks for the clarification

PR #1273, opened 2025-12-10
Add script to run unit tests in parallel threads

nascheme:
You can the parallel tests by running this command:
```
make parallelcheck
```

By default, the total number of tests is limited to 400.  To adjust or remove that limit, run the script directly, e.g.
```
dn=$(echo build/lib.*t); PYTHONPATH=$dn python3  $dn/tables/tests/run_ft.py  --max-tests=1000
```
Remove the `--max-tests` option to run all the tests (will take quite a bit more time).

The `_EXCLUDE_CASES` variable in the script contains a hard-coded list of tests to exclude.  These fail, likely because the test is written in a way that is not thread-safe.  These exclusions should be reviewed however.  Ideally we would mark the cases themselves, similar to what's done for pytest-run-parallel.

This kind of testing is useful for ensuring the extension works with the GIL disabled (e.g. free-threaded Python with the `PYTHON_GIL=0` option).  We don't want the extension to crash when used this way.  That would be the minimum condition for setting the `Py_MOD_GIL_NOT_USED` module flag, IMHO.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ OPT = PYTHONPATH="$(PYBUILDDIR)"
 MD5SUM = md5sum
 
 
-.PHONY: default dist sdist build check heavycheck clean distclean html latex requirements lint
+.PHONY: default dist sdist build check heavycheck parallelcheck clean distclean html latex requirements lint
 
 default: $(GENERATED) build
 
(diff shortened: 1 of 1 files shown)


avalentino:
Thanks @nascheme

PR #1274, opened 2025-12-10
Create CI workflow to test free-threaded (GIL-disabled) build

nascheme:
This adds a new workflow file (ubuntu-ft.yml) that builds the extension using the free-threaded version of Python (currently 3.14t) and then tests using run_ft.py.

There are at least two things that need changing before this gets merged:

- I've limited the number of tests to 400, to speed up the CI run.  That limit should be removed so we run all (non-excluded) tests.
- I'm removing the blosc2 Python package.  I'm not sure why this step is needed, need to investigate further.  With this installed, the utilsextension SO file fails to load with a linker error, libblosc2.so.6 => not found.

### DIFF
--- a/.github/workflows/ubuntu-ft.yml
+++ b/.github/workflows/ubuntu-ft.yml
@@ -1,0 +1,44 @@
+name: Ubuntu free-threaded tests
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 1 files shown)


avalentino:
> I'm removing the blosc2 Python package. I'm not sure why this step is needed, need to investigate further. With this installed, the utilsextension SO file fails to load with a linker error, libblosc2.so.6 => not found.

removing the blosc2 Python package is correct for the Ubuntu build.
We should do it also in the standard Ubuntu build.

PR #1276, opened 2025-12-19
chore: support python 3.14

reneleonhardt:
### Chores
- Support Python 3.14
- Update github-actions and gitsubmodule

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,17 +9,15 @@ updates:
         patterns:
           - "*"
 
-  - package-ecosystem: "pip"
+  - package-ecosystem: "gitsubmodule"
     directory: "/"
     schedule:
       interval: "monthly"
-    groups:
-      pip:
-        patterns:
-          - "*"
 
   - package-ecosystem: "pip"
-    directory: "/.github/workflows/requirements"
+    directories:
+      - "/"
+      - ".github/workflows/requirements"
     schedule:
       interval: "monthly"
     groups:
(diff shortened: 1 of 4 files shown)


avalentino:
> ### Notes
> * Are the numpy exceptions for Python 3.12 still needed in ci.yml?

I think that the spirit was to be sure that both bumpy 1.x and numpy 2 dev are tested.
It was done at the time on the latest python version to limit the number of CI jobs (nothing really related to Python 3.12)
IMHO keep testing dumpy 1.x is no longer needed, we could simply drop the support for it.
I would like to keep lumpy dev, but please feel free to suggest a better way to do it

> * Can Python 3.12 be upgraded in wheels.yml?

Not sure why we have 3.12 there.
It everything keeps working I would not have objections to change

> * Are there plans to build Windows arm64 wheels?

I would her no objection but I do not have time and will to do anything related to windows.
If anything breaks at some point in CI I will just disable the job until a volunteer manages to fix the issue.

> * Are there plans to support the amazing HDF5 2 release?
>   https://www.hdfgroup.org/2025/11/10/release-of-hdf5-2-0-0-newsletter-207/

It would be very nice but it looks a big job.
I have still not investigated the impact to be honest.
In this moment I have no concrete plans but being able to link hdf5 v2 it is something that I have in mind.
This of course, would be just a very basic support for HDF5 v2.
If I understand correctly the library has a lot of very interesting features that should be used only marginally by the current PyTables.

PR #1286, opened 2026-01-07
Add 3.14 wheel building support

### LINKED ISSUE #1261: Python 3.14 support
Is there a plan to provide the wheels for Python 3.14 via PyPi for Windows and Linux and if so, when will they be available?

larsoner:
I noticed there were no scientific-python-nightly-wheels for 3.14. This is a naive attempt to fix it:

1. Remove `cibuildwheel` pin to `<3` (not clear to my why it was added in b02bfa01df04b95088eccfebaed6ec814abefe09, let's find out the hard way!)
2. `pip-compile --allow-unsafe --generate-hashes --strip-extras .github/workflows/requirements/wheels-requirements.in > .github/workflows/requirements/wheels-requirements.txt`
3. Restore all jobs running in matrix using `native` arch (not sure why only Windows was running on first few commits, but this fixed it!)
4. Restore `macos-15-intel` runner and wheels (shouldn't be too onerous hopefully?)

I'm hopeful for (1) that some tweaking of `pyproject.toml` will fix any bugs that come up but we'll see!

Closes #1261

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -185,6 +185,11 @@ jobs:
         with:
           python-version: '3.x'
 
+      - name: Install patchelf
+        if: runner.os != 'Windows'
+        run: |
+          python -m pip install patchelf
+
       - name: Install twine
         run: |
           python -m pip install --require-hashes -r ./.github/workflows/requirements/wheels-requirements.txt


larsoner:
Oof there are four 3.14 test failures on Windows that seem legit:

https://github.com/PyTables/PyTables/actions/runs/20787499734/job/59701020264?pr=1286#step:9:6671

```
  FAIL: None (tables.tests.test_direct_chunk.CArrayDirectChunkingTestCase.None)
    File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-x7gmsz2t\cp314-win_amd64\venv-test\Lib\site-packages\tables\tests\test_direct_chunk.py", line 129, in test_read_chunk
      self.assertEqual(chunk, obj_bytes)
      ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  AssertionError: b'x\x01c`dec\x00\x02\x00\x00K\x00\r' != b'x\x01c`dec```\x00\x00\x00K\x00\r'
```

I'll skip those for now, not sure how critical they are but I think maybe they can be fixed in a follow-up PR?

And not sure why `macos-15-intel` was removed from the matrix, but it [works here](https://github.com/PyTables/PyTables/actions/runs/20787499734/job/59701020297?pr=1286) (other than `cp314t` which fails on all runs). I'll leave it disabled since I'm assuming it was removed for good reason, but can easily add it back if desired.

larsoner:
> Regarding the pinning of cibuildwheels, I did it because it seems to introduce some build failures ion win64 that I do not see with cibuildweels v2.x. ... we get some issue on win64 in the direct_chunking code.

Looks like the failure is [only in 3.14](https://github.com/PyTables/PyTables/actions/runs/20787499734/job/59701020264#step:9:6403). So I suspect the problem isn't that all Windows builds are broken with cibuildwheel 3+, but rather that moving to 3+ enables Python 3.14 builds which are somehow problematic on Windows. I've added a skip for the problematic tests for now, but a safer approach would maybe be to exclude Windows 3.14 from `cibuildwheel` for now?

avalentino:
Thanks @larsoner

abulgher:
> hello, the PR for 3.14 were closed but there is still no package installable on 3.14 https://pypi.org/project/tables/#history
> 
> do you need to publish the wheel manually or something?

I am also surprised by the fact that there were no wheels for 3.14 on pypi. I guess we have to wait until the release of 3.11.0.

(Python 3.15 is expected for October 2026, so we still have a bit of time ahead! :smile: )

PR #1289, opened 2026-01-07
Move to abi3 builds

### LINKED ISSUE #1287: abi3 wheels?
Have you looked into / considered using `abi3` wheels at all? Cython supports it and I've converted other Cython-based projects to use it and it's not so bad, see for example:

https://github.com/nipy/nitime/pull/233

It would cut your wheel build time down to a single job, be future compatible with other versions of Python, etc. The downside if you use heavy `cython` it [can be slower](https://cython.readthedocs.io/en/latest/src/userguide/limited_api.html#performance), but might be worth it to reduce maintenance burden.

Happy to take a stab at a PR to convert to see if it's easily feasible...

larsoner:
Closes #1287

Let's see if it at least can be made to work. Then someone else can see if the produced wheels have sufficient performance (somehow).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ import sys
 import glob
 import ctypes
 import shutil
+import sysconfig
 import fnmatch
 import platform
 import tempfile
@@ -19,6 +20,7 @@ from pathlib import Path
 from setuptools import setup, Extension
 from packaging.version import Version
 from setuptools.command.build_ext import build_ext
+from wheel.bdist_wheel import bdist_wheel
 
 # The name for the pkg-config utility
 PKG_CONFIG = "pkg-config"
@@ -41,6 +43,10 @@ def exit_with_error(head, body=""):
 
 def print_warning(head, body=""):
     _print_admonition("warning", head, body)
+
+
+def is_freethreaded():
+    return bool(sysconfig.get_config_var("Py_GIL_DISABLED"))
 
 
 # Get the HDF5 version provided the 'H5public.h' header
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1049,6 +1061,21 @@ def find_runtime_path(self, locations=None):
...
     ]
 
+    cmdclass = {"build_ext": BuildExtensions}
+    abi3_macros = []
+    abi3_ext_kwargs = {}
+    abi3_cflags = ""
+    if (
+        platform.python_implementation() == "CPython"
+        and "free-threading build" not in sys.version
+        and Version(cython.__version__) >= Version("3.1")
```
larsoner: This was needed to get the Ubuntu CI happy because it uses old Cython (3.0.8) where limited API support was insufficient. I'm silghtly inclined to put `cython >=3.1` in the build-requires instead...

larsoner:
@avalentino green! Feel free to build yourself or [download an artifact from the build](https://github.com/PyTables/PyTables/actions/runs/20799881083?pr=1289) if you want to benchmark etc.

No problem if it's not performant enough, that's already good to know. I can roll the PR back just to some simplifications of the YAML and missing `#include`s (that must have only not been a problem before because they were included by Cython when building or something). But hopefully it's fast enough!

avalentino:
Thanks @larsoner 
It would be nice if you could rebase to have just 2 comments: one for the general improvements and one for abi3.

I did few quick benchmarks and I do not see appreciable slow downs.

In any cases I will ask the review form @FrancescAlted before merging

larsoner:
@avalentino done, both commits green. First one took 69 minutes for `ubuntu-latest` 😱 , the `abi3` version took 5 min, which is a nice CI side benefit! 🚀 

Incidentally, at some point we could add `test_wheels` jobs for example that use the wheels resulting from `build_wheels` jobs, for example one on 3.11 (lowest supported) and another on 3.14 (newest), plus maybe one that uses `scientific-python-nightly-wheels`. I think that can be done in another PR if someone is interested, but I could also do it here if people think it's necessary to trust the `abi3` compat (but I think we can trust `Cython` here).

FrancescAlted (approved):
LGTM

morotti:
hello, can you make a release of pytables?

pytables still doesn't have support for python 3.14, which has been out for 4 months now.
this PR was supposed to make the package compatible, by having single abi3 package compatible with multiple python versions.

Issue #1254, opened 2025-08-19
Busy loop - potentially in search_last_row


alpae:
Hi, 

I have a problem with a process that suddenly enters in an infinite busy loop inside search_last_row of the Index. This busy-loop isn't even interruptible with CTRL-C. 

I tried to identify where it in the library this happens by adding a 
```
import faulthandler
faulthandler.dump_traceback_later(5, repeat=True)  # every 5s
```
before launching the script. `process_family` is executed ~2000 times successfully when it suddenly get stuck.

```
Timeout (0:00:05)!

Thread 0x00007fd666364b80 (most recent call first):
  File "/usr/src/.venv/lib/python3.11/site-packages/tables/index.py", line 2284 in search_last_row
  File "/usr/src/.venv/lib/python3.11/site-packages/tables/index.py", line 2211 in search
  File "/usr/src/.venv/lib/python3.11/site-packages/tables/table.py", line 212 in _table__where_indexed
  File "/usr/src/.venv/lib/python3.11/site-packages/tables/table.py", line 1603 in _where
  File "/usr/src/.venv/lib/python3.11/site-packages/tables/table.py", line 1575 in where
  File "/usr/src/.venv/lib/python3.11/site-packages/pyoma/browser/build/cachebuilder.py", line 127 in load_fam_members
  File "/usr/src/.venv/lib/python3.11/site-packages/pyoma/browser/build/cachebuilder.py", line 112 in process_family
  File "<ipython-input-14-44d9c3965726>", line 5 in <module>
  File "/usr/src/.venv/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3699 in run_code
  File "/usr/src/.venv/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3639 in run_ast_nodes
  File "/usr/src/.venv/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3394 in run_cell_async
  File "/usr/src/.venv/lib/python3.11/site-packages/IPython/core/async_helpers.py", line 128 in _pseudo_sync_runner
  File "/usr/src/.venv/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3171 in _run_cell
  File "/usr/src/.venv/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3116 in run_cell
  File "/usr/src/.venv/lib/python3.11/site-packages/IPython/terminal/interactiveshell.py", line 1005 in interact
[...]

PR #1321, opened 2026-05-08
Fix infinite busy loop in ObjectCache.updateslot_ (issue #1254)

alpae:
When fewer than 10 slots are occupied and a new item would push cachesize over maxcachesize, the eviction while-loop in ObjectCache.updateslot_ would spin forever. Empty slots have atime=0 (initialised with np.zeros), while occupied slots always have atime>=1 (set by incseqn). The sizes.argsort() top-10 selection included empty slots when fewer than 10 slots were in use, and argmin then always picked an empty slot. removeslot_ on an empty slot is a no-op, so cachesize never decreased and the loop never exited.

Fix: filter largidx to only occupied slots (size > 0) before running argmin.

Add tests/test_lrucache.py with regression tests that would hang forever on the unfixed code but complete immediately after the fix.

I realize that this is essentially similiar to PR #890 where @zequihg50 seems to have a same type of problem as reported in #1254

### DIFF
--- a/tables/lrucacheextension.pyx
+++ b/tables/lrucacheextension.pyx
@@ -372,8 +372,15 @@ cdef class ObjectCache(BaseCache):
     self.removeslot_(nslot)
     # Protection against too large data cache size
     while size + self.cachesize > self.maxcachesize:
-      # Remove the LRU node among the 10 largest ones
+      # Remove the LRU node among the 10 largest ones.
+      # Only consider occupied slots (size > 0): empty slots have atime 0,
+      # which is always less than any occupied slot's atime (>= 1), so
+      # argmin would always pick an empty slot, making removeslot_ a no-op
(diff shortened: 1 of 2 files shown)


avalentino:
Dear @alpae 
thanks a lot for looking in this issue, it was indeed quite tricky.
... and sorry if I have not looked at it before.

The PR looks mostly good.
There are many test failures in CI, but they seem not related to you changes in most of the cases.
I would be nice if you could fix the linting error (with pre-commit it should be quite easy).
Moreover I would kindly ask you to rebase on the current master so that we can have a clear picture about possible issues in CI (still on windows some failure is expected at this stage).

avalentino (approved):
Thanks a lot @alpae

alpae:
thanks for accepting this so quickly.

PR #1327, opened 2026-06-29
Non-zero direct-chunk filter mask stored and reported back.  Fixes #1325.

### LINKED ISSUE #1325: `test_write_chunk_filtermask` is fragile when overwriting an existing same-size direct chunk
## Summary

I found a fragile behavior in `tables.tests.test_direct_chunk.EArrayDirectChunkingTestCase.test_write_chunk_filtermask`.

The test assumes that `H5Dwrite_chunk(..., filter_mask=4, ...)` updates the stored `filter_mask` metadata when overwriting an existing chunk. However, in one build configuration, the existing chunk and the replacement direct chunk have the same stored size, and HDF5 appears to overwrite the chunk data in place while keeping the previous `filter_mask=0`.

The actual data roundtrip still succeeds. The failure is only in the metadata assertion:

```text
AssertionError: 0 != 4
```

## Environment

Tested with:

```text
PyTables version: 3.11.1
HDF5 version: 2.1.1
Python version: 3.14.2
NumPy version: 2.4.6
GCC version: 15.2.0
Platform: Linux x86_64, RHEL9
```

Relevant compression libraries in the failing build:

```text
Blosc version: 1.21.6, linked externally as libblosc.so.1
Blosc2 version: 3.1.2
```

## Failure

The failing test is:

```python
tables.tests.test_direct_chunk.EArrayDirectChunkingTestCase.test_write_chunk_filtermask
```

Failure:

```text
FAIL: None (tables.tests.test_direct_chunk.EArrayDirectChunkingTestCase.None)
  File ".../tables/tests/test_direct_chunk.py", line 226, in test_write_chunk_filtermask
    self.assertEqual(chunk_info.filter_mask, no_shuffle_mask)
AssertionError: 0 != 4
```

## Investigation

The test writes a modified chunk using Python `zlib.compress(..., shuffle=False)` and then calls:

```python
self.array.write_chunk(chunk_start, obj_bytes, filter_mask=0x00000004)
```

The data comparison passes, so the direct chunk payload is written and read correctly. Only the final `chunk_info.filter_mask` assertion fails.

I compared two builds:

### Build using external Blosc

```text
obj shape: (5, 5)
obj dtype: uint16
chunkshape: (2, 2)
raw len: 8
compressed len: 16
before: ChunkInfo(start=(0, 0), filter_mask=0, offset=3342, size=16)
after write before reopen: ChunkInfo(start=(0, 0), filter_mask=0, offset=3342, size=16)
after reopen: ChunkInfo(start=(0, 0), filter_mask=0, offset=3342, size=16)
data equal after manual fix: True
```

Here, the existing chunk size and replacement chunk size are both 16 bytes. HDF5 appears to overwrite the existing chunk in place and keep the old `filter_mask=0`.

### Build using bundled c-blosc

```text
obj shape: (5, 5)
obj dtype: uint16
chunkshape: (2, 2)
raw len: 8
compressed len: 16
before: ChunkInfo(start=(0, 0), filter_mask=0, offset=3342, size=14)
after write before reopen: ChunkInfo(start=(0, 0), filter_mask=4, offset=3446, size=16)
after reopen: ChunkInfo(start=(0, 0), filter_mask=4, offset=3446, size=16)
data equal after manual fix: True
```

Here, the original stored chunk size is 14 bytes and the replacement direct chunk is 16 bytes. HDF5 rewrites the chunk record, and the new `filter_mask=4` is reported correctly.

I also checked writing to a missing chunk instead of overwriting an existing chunk. That preserves `filter_mask=4` in both builds.

Example result:

```text
missing chunk before write: ChunkInfo(start=(4, 0), filter_mask=None, offset=None, size=None)
after writing missing chunk with mask=4: ChunkInfo(start=(4, 0), filter_mask=4, offset=3341, size=20)
```

## Interpretation

This looks like a fragile test assumption rather than a general PyTables runtime failure.

The test outcome depends on the incidental compressed size of the existing chunk:

```text
existing size != replacement size -> filter_mask is updated
existing size == replacement size -> filter_mask may remain unchanged
```

So the test can pass or fail depending on the compression/linking environment, even though the direct chunk data roundtrip succeeds.

## Proposed fix

For `EArrayDirectChunkingTestCase.test_write_chunk_filtermask`, avoid testing `filter_mask` by overwriting an existing chunk. Instead, extend the EArray and write a missing chunk. This still verifies that a non-zero direct-chunk `filter_mask` is stored and reported correctly, but avoids relying on HDF5 updating metadata during a same-size in-place overwrite.

A possible patch is:

```python
def test_write_chunk_filtermask(self):
    no_shuffle_mask = 0x00000004  # to turn shuffle off

    chunk_start = (0,) * self.obj.ndim

    if isinstance(self.array, tb.EArray):
        # Avoid relying on HDF5 updating filter_mask metadata when
        # overwriting an existing same-size chunk.  Extend the EArray and
        # write a missing chunk instead, which still verifies that a non-zero
        # filter_mask is stored and reported correctly.
        chunk_start = (
            ((self.array.nrows + self.chunkshape[0] - 1)
             // self.chunkshape[0]) * self.chunkshape[0],
            *((0,) * (self.obj.ndim - 1)),
        )
        self.array.truncate(chunk_start[0] + self.chunkshape[0])
        new_obj = self.array[:]

        source_slice = tuple(slice(0, cs) for cs in self.chunkshape)
        obj_slice = tuple(
            slice(s, s + cs) for (s, cs) in zip(chunk_start, self.chunkshape)
        )
        new_obj[obj_slice] = self.modified(self.obj[source_slice])
    else:
        obj_slice = tuple(
            slice(s, s + cs) for (s, cs) in zip(chunk_start, self.chunkshape)
        )
        new_obj = self.obj.copy()
        new_obj[obj_slice] = self.modified(new_obj[obj_slice])

    obj_bytes = self.compress_chunk(
        new_obj[obj_slice].tobytes(), shuffle=False
    )
    self.array.write_chunk(
        chunk_start, obj_bytes, filter_mask=no_shuffle_mask
    )

    self._reopen()
    arr_obj = self.array[:]
    fixed_bytes = self.shuffled(arr_obj[obj_slice].tobytes())
    fixed_chunk = np.ndarray(
        self.chunkshape, dtype=self.obj.dtype, buffer=fixed_bytes
    )
    arr_obj[obj_slice] = fixed_chunk
    self.assertTrue(common.areArraysEqual(arr_obj, new_obj))

    chunk_info = self.array.chunk_info(chunk_start)
    self.assertEqual(chunk_info.filter_mask, no_shuffle_mask)
```

I am not sure whether HDF5 is expected to update the stored `filter_mask` when overwriting an existing same-size chunk. If that behavior is guaranteed by HDF5, then this may point to a lower-level HDF5 issue. But from the PyTables test-suite perspective, testing a missing chunk seems more robust because it avoids depending on same-size overwrite metadata behavior.

FrancescAlted:


avalentino (approved):

avalentino:
"Wheels / Build source distribution" seems to fail for unrelated reasons

avalentino:
Thanks @FrancescAlted

FrancescAlted:
Sorry @avalentino , but I was still evaluating the issue and wanted to check with my own fork first.  My bad for specifying that it "Fixes #1327"; I was not expecting this to arrive here so fast.  Give some more time before I can get a more in-depth review.

avalentino:
Sorry @FrancescAlted I thought you were finished.
Unfortunately I cannot reopen the PR.
Please feel free to open a new one.

PR #1339, opened 2026-07-23
Use cython >= 3.2

### LINKED ISSUE #1338: No wheels available on scientific-python-nightly-wheels
See `tables` absent from https://anaconda.org/channels/scientific-python-nightly-wheels.

Looks like it's actually due to a failure at the [source distribution](https://github.com/PyTables/PyTables/actions/runs/29689482910) prep step. I can look if it would help!

I'd also be happy to open a PR to add an extra step that opens an issue when this fails to increase visibility if it would help. I know scikit-learn has something like this.

avalentino:
Cython >= 3.2 seems to be needed to support modern compilers.

Fixes #1338.


avalentino:
The change seems to fix the original issue but creates a regression with Python 3.11.
The solution should be to just re-generate the requirements files with Python 3.11.

PR #1343, opened 2026-07-27
Use cython >= 3.2

### LINKED ISSUE #1338: No wheels available on scientific-python-nightly-wheels

larsoner:
@avalentino a small diff... just remove the Cythonized `.c` files so that the in-tree build always creates its own... 

...Then merge `upstream/master` into the branch. Then restore accidental -- I think -- setuptools pin upward drift. Then prevent it from drifting again by tweaking dependabot. Then fix a pyproject.toml <-> requirements.in inconsistency with NumPy floor. Then add a check for it in CIs that would have caught these issues 😄 If this is too much, can revert parts of these! But hopefully they help with maintainability.

Investigated with Claude Opus 5 but I understand and vouch for the changes.

Closes #1339
Closes #1338


avalentino (approved):

avalentino:
Thanks @larsoner

PR #1354, opened 2026-09-17
Add Windows ARM64 build

ndabas:
As the title says -- the one major change I have made here is to swap out `mamba` for `vcpkg`. `mamba`, as it pulls from `conda-forge`, does not have any win-arm64 build of HDF5 (and it cannot build from source.) The `vcpkg` approach is closer to the HDF5 build that other platforms use, IMHO.

I have also updated `blosc2` in `requirements.txt` because they [added win_arm64 wheels](https://github.com/Blosc/python-blosc2/pull/709) in a recent release.

The new `get_deps.ps1` script for Windows has a block to compile `openssl`; that part can be removed once `cryptography` publishes their next release, which should include win_arm64 wheels.


avalentino (approved):

avalentino:
Thanks a lot @ndabas

